PR #8, opened 2016-02-04
Support a bit larger images

lookfwd:
If you don't mind - let's make this a bit larger.


lookfwd:
It works fine with .svg's but it crops (admittedly extremely) large images.

PR #10, opened 2016-04-05
Created README.md in src directory

mikeando:
Primarily so that it shows up in github when viewing the src directory.

IMO a lot of the header of this file could be removed, but I thought I"d keep it true to the original HTML file for now.


bsene:
@arnaudroques is this file relevant from your pointof view ?

PR #15, opened 2016-06-10
Restructured files according to Maven standard layout, removed Ant's build.xml

thetric:
The original project uses Maven as well as Ant for the build process.  It would be easier to stick with Maven and its conventions so that

a) less configuration is necessary, and
b) it is easier for new developers familiar with Maven/Gradle to understand the structure.

In addition, I have
- fixed some Maven build warnings concerning missing Maven plugin versions and file encoding
- set the compiler level to the current JDK version 8 (which is the only currently supported version by Oracle)


thetric:
These further commits move all resources to `src/main/resources`

PR #31, opened 2017-01-19
Information Engineering (IE) data modelling notation support.

mgeorge:
Extended @bminderh's crowsfoot work to include the full IE notation *except* for OR/XOR.

# Includes
- All IE extremities
- A new 'mandatory' visibility modifier '*' produces a black circle
- A syntax alias 'entity' which is aliased to 'class'

# Diagram example
```
@startuml

hide circle
hide empty members

entity Entity {
   * identifying_attribute
   --
   * mandatory_attribute
   optional_attribute
}

' zero or one
A |o--o| B

' exactly one
C ||--|| D

' zero or many
E >o--o< F

' one or many
G >|--|< H

@enduml
```

# To do
Couldn't work out how to do the following myself.
- Make mandatory visibility icon color based on attribute text color.  Currently hard-coded to black which is not ideal.

### DIFF
--- a/src/net/sourceforge/plantuml/ColorParam.java
+++ b/src/net/sourceforge/plantuml/ColorParam.java
@@ -152,7 +152,8 @@ public enum ColorParam {
 	iconProtected(HtmlColorUtils.COL_B38D22),
 	iconProtectedBackground(HtmlColorUtils.COL_FFFF44),
 	iconPublic(HtmlColorUtils.COL_038048),
-	iconPublicBackground(HtmlColorUtils.COL_84BE84);
+	iconPublicBackground(HtmlColorUtils.COL_84BE84),
+	iconIEMandatory(HtmlColorUtils.BLACK);
 	
 	private final boolean isBackground;
 	private final HtmlColor defaultValue;
--- a/src/net/sourceforge/plantuml/classdiagram/command/CommandCreateClass.java
+++ b/src/net/sourceforge/plantuml/classdiagram/command/CommandCreateClass.java
@@ -68,7 +68,7 @@ public class CommandCreateClass extends SingleLineCommand2<ClassDiagram> {
 	private static RegexConcat getRegexConcat() {
 		return new RegexConcat(new RegexLeaf("^"), //
 				new RegexLeaf("TYPE", //
-						"(interface|enum|annotation|abstract[%s]+class|abstract|class)[%s]+"), //
+						"(interface|enum|annotation|abstract[%s]+class|abstract|class|entity)[%s]+"), //
 				new RegexOr(//
 						new RegexConcat(//
 								new RegexLeaf("DISPLAY1", "[%g](.+)[%g]"), //
--- a/src/net/sourceforge/plantuml/classdiagram/command/CommandCreateClassMultilines.java
+++ b/src/net/sourceforge/plantuml/classdiagram/command/CommandCreateClassMultilines.java
@@ -81,7 +81,7 @@ public class CommandCreateClassMultilines extends CommandMultilines2<ClassDiagra
 	private static RegexConcat getRegexConcat() {
 		return new RegexConcat(new RegexLeaf("^"), //
 				new RegexLeaf("VISIBILITY", "(" + VisibilityModifier.regexForVisibilityCharacter() + ")?"), //
-				new RegexLeaf("TYPE", "(interface|enum|abstract[%s]+class|abstract|class)[%s]+"), //
+				new RegexLeaf("TYPE", "(interface|enum|abstract[%s]+class|abstract|class|entity)[%s]+"), //
 				new RegexOr(//
 						new RegexConcat(//
 								new RegexLeaf("DISPLAY1", "[%g](.+)[%g]"), //
--- a/src/net/sourceforge/plantuml/classdiagram/command/CommandLinkClass.java
+++ b/src/net/sourceforge/plantuml/classdiagram/command/CommandLinkClass.java
@@ -119,7 +119,7 @@ final public class CommandLinkClass extends SingleLineCommand2<AbstractClassOrOb
 
 	private static String optionalKeywords(UmlDiagramType type) {
 		if (type == UmlDiagramType.CLASS) {
-			return "(interface|enum|annotation|abstract[%s]+class|abstract|class|object)";
+			return "(interface|enum|annotation|abstract[%s]+class|abstract|class|object|entity)";
 		}
 		if (type == UmlDiagramType.OBJECT) {
 			return "(object)";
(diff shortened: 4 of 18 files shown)


mgeorge:
The above syntax would produce the following diagram.
ie-demo

arnaudroques:
That's great! Congratulation for your work. We will be glad to merge this in the base branch.
And we will have a look at the hard-coded black issue.
We just have a syntax suggestion: wouldn't it be more logical to use { instead of < to be consistent with exising crowsfoot notation ?
That is, using:
E }o--o{ F
G }|--|{ H

What do you think about it ? (You don't have to do the code modification, we could if you want).

mgeorge:
Using '{' is fine by me.  If you could make the change during the merge then that would be great.

Also note that the crowsfoot in this version is smaller and narrower that @bminderh's version.  This is because the crowsfoot in both versions do not work well with non-perpendicular angles since one leg may overlap the entity while the other dangles.  Smaller and narrower makes this effect slightly less obnoxious.  The following image demonstrates what I mean.

test

I thought about trying to dynamically scale the legs to ensure they intersect with the entity, but there are many cases where one leg will never intersect.  If users don't like it they always have the option of using the ortho linetype to avoid weird angles.

mgeorge:
And thank you for accepting this work, and for your great work on PlantUML.  I added the IE notation because we will likely use PlantUML as our primary diagramming tool for teaching several software engineering related course's in the Department of Information Science at the University of Otago (New Zealand).

arnaudroques:
Here is a beta that integrates your code with { modification.
https://dl.dropboxusercontent.com/u/13064071/plantuml.jar

Note that in V8052 we have patched @bminderh version so that non-perpendicular angles are displayed warped. We've just added a projection to Point2D left and right, see https://github.com/plantuml/plantuml/blob/master/src/net/sourceforge/plantuml/svek/extremity/ExtremityCrowfoot.java

The result is not perfect:

Compare for example:

> @startuml
> hide circle
> hide empty members
> 
> foo1 --{ bar1
> foo1 --{ bar2
> foo1 --{ bar3
> 
> aa --o{ bb
> aa --o{ cc
> aa --o{ dd
> @enduml

If you like the result, we could do the same trick with your code.
What do you think about it ?

mgeorge:
Yes that looks good.

mgeorge:
Just noticed a problem with using the '*' character for the new visibility modifier - it conflicts with the creole bold.

I am fine with you choosing another character for the visibility modifier.

arnaudroques:
Well, we like the * notation :-)

So in the last beta https://dl.dropboxusercontent.com/u/13064071/plantuml.jar
we've fixed the following clash:

```
@startuml
hide circle
hide empty members

entity Entity {
   * identifying_attribute
   --
   * mandatory_attribute **in bold**
   **optional_attribute**
}
@enduml
```

Tell us if you find other issues,
Thanks

xorith:
So to my disappointment, the crows feet notation shown above does not appear to be a feature in PlantUML? It's hard to tell what was included and what was not.

xorith:
> Well, we like the * notation :-)
> 
> So in the last beta https://dl.dropboxusercontent.com/u/13064071/plantuml.jar
> we've fixed the following clash:
> 
> ```
> @startuml
> hide circle
> hide empty members
> 
> entity Entity {
>    * identifying_attribute
>    --
>    * mandatory_attribute **in bold**
>    **optional_attribute**
> }
> @enduml
> ```
> Tell us if you find other issues,
> Thanks

Also this doesn't appear to work, with the curly brace after the name of the entity.

mgeorge:
It was merged ages ago.  It is not documented yet, but is basically piggy backing off the class diagrams.

https://www.planttext.com/?text=FSnD2e0W48RX_PnYre9tq6uobe7aJupJmgIxfmHjlitpyPegn7Ab252Fa4zYIbYbHJ-Mm_gHpnAaeSKEQYy5j0P73CylonHqT4zIE3y3BUgE2u3nUHkT5m00

Not sure what you other comment refers to - your exact code seem to do what I would expect it to do:

https://www.planttext.com/?text=HOqx2e0m40NxFSNM2zx1ndF8Ig8in8t4Pn74k_ihKZsO6NZz3YquraGK7ORAmpydCREmXHVFrNmP5IWs7OYlwhe_MDM38fVQD8kOyvHJ4FaYoHiikwRMFEOYwgE7v_i6

xorith:
It would appear to be an error with the Atom plantuml-viewer plugin. My apologies. It is giving me a syntax error.

mlopezgva:
Apart from the documentation, I would like to point there's some kind of "textual representation" for ER that could have been ported to PlantUML. I have found [a comprehensive document about it](https://db.grussell.org/section004.html). In short, it could have been like this:

```erd
@startuml

entity Bus (driver, route)
entity Route (stage)
entity Driver
entity Stage (Town)
entity Town

Bus   m..1 Route : is serviced by
Route 1..m Stage : comprises
Driver m..1  : is allocated to
Stage n..m Town : Passes-though
Route n..m Town : Passes-though
```

Which is more readable than the cryptic `}|-||` (which is related to the drawn symbols, I know).
Also, the attributes can be written in a more classical way, less "java-ish", as show in the entities declarations.

I know I arrived late to the syntax discussion, but this syntax is related to the textual description of ER and, thus, closer to other documentation texts we can find or write (easier to reuse ;-)).

If attributes are listed and not hidden (`hide attributes`, i.e.) they could be drawn following the ERD stadard: UseCase-like oval surrounding their entity.

And, just for completeness and drive the developers crazy, declare the relationships as diamonds if `skinparam erdModel chen`, etc.

But, just a thought. Document the current implementation would be nice ehough. O:)

spawnia:
@mgeorge thank you very much for your implementation work. I would really appreciate some documentation for this feature. You can add it in the wiki: http://wiki.plantuml.net/site/index

mgeorge:
@spawnia Done.  Thank you for actually pointing out how/where to add the documentation.

http://wiki.plantuml.net/site/ie-diagram

spawnia:
@mgeorge thank you very much. I find it more clear to name the page `Entity Relationship diagram`. Information Engineering seems more like the discipline associated with creating ER diagrams.

I actually edited that in just now, did not realize the Wiki allows me to just edit that directly. If you want i can revert that, though.

What do you think?

mgeorge:
I was thinking the same thing.  Keep the edit.  I just added a sentence that clarified which notation is being used (there are a bunch of them and Information Engineering is the actual name of the crows foot notation that most people use).

arnaudroques:
Many thanks for your contributions on documentation.
It's now online at http://plantuml.com/ie-diagram

You can go on with further modifications if needed.
Thanks again !

openbankingproject-ch:
You are my heros!

PR #48, opened 2017-03-27
typos and cleanup

aadrian:


arnaudroques:
Many thanks for your corrections! We will integrate them, maybe except for those in the code we did not write ourself.
Regards,

aadrian:
> We will integrate them, maybe except for those in the code we did not write ourself.

Not sure why would one handle sources like that. The entire code is yours now. If it's not, than it should not be in the ```src``` directory but as a JAR dependency.

Issue #91, opened 2017-12-22
Prefix inlined dependencies


SevereOverfl0w:
As you include your dependencies source code, when using this library with maven there are classpath conflicts. This results in using a version of the code that is unexpected if you wish to use plantuml as a library.

The simple solution here is to sed all references in the project to have the prefix plantuml. Or you could use jarjar as part of the build process https://code.google.com/archive/p/jarjar/#!

arnaudroques:
Thanks for pointing this out.
We have not though about this...
As you have suggested, we are going to modify all external library by adding some prefix (something like ext.plantuml) so that classpath conflicts will be removed.
Regards,

thetric:
In #93 I'm trying to replace the external libs with Maven dependencies which is in my opinion a better approach than merely prefixing as it's easier to update them. 

@SevereOverfl0w what packages are conflicting in your build? Perhaps I can declare them as maven dependencies too

SevereOverfl0w:
Working from memory, it was a Google one.
 Having them as a maven dependency would be great.

I recently learned of the maven shade plugin which can prefix maven dependencies. This is useful for distributing a fat jar on the plantuml website for those not using maven.

On 3 January 2018 22:38:48 GMT+00:00, Dominik Broj <notifications@github.com> wrote:
>In #93 I'm trying to replace the external libs with Maven dependencies
>which is in my opinion a better approach than merely prefixing as it's
>easier to update them. 
>
>@SevereOverfl0w what packages are conflicting in your build? Perhaps I
>can declare them as maven dependencies too
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/plantuml/plantuml/issues/91#issuecomment-355147873

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

thetric:
Ok, in my PR the Google lib zxing is already a maven dependency. The only problem is that a dependency of zxing doesn't work with Java 9 (see my first comment in the PR).

Are you sure that it is still necessary to prefix all external libs in the fat jar? The PR updated zxing to the latest version and should not cause problems

SevereOverfl0w:
I'm guessing a little at how the distributed jars work. Prefixing is only useful if plantuml want to offer a single jar outside of maven.

On 4 January 2018 00:18:22 GMT+00:00, Dominik Broj <notifications@github.com> wrote:
>Ok, in my PR the Google lib zxing is already a maven dependency. The
[...]

Issue #47, opened 2017-03-26
Add Gradle as a build system!


aadrian:
Add Gradle as a build system since it's much easier to manage multi-module projects.

Thank you.

arnaudroques:
Good idea!
However, we know very few about Gradle...
So we are open to contribution to create our build.gradle file. Could you help us on this ?
Thanks!

aadrian:
@arnaudroques of course. I'll propose a module structure (e.g. at least the GPL Ditaa code as a separate module), and submit PRs.

aadrian:
@arnaudroques https://github.com/plantuml/plantuml/wiki/Module-Proposal

thetric:
@aadrian  Some external libraries can be linked as a dependency if a GitHub repo exists so the project size decreases. The website https://jitpack.io/ enables developers to use any GitHub repo as a dependency as it builds and provides the projects on demand.

Example: https://github.com/stathissideris/ditaa provides the `org/stathissideris/ascii2image` package. The Gradle config:

```gradle
repositories {
    maven { url 'https://jitpack.io' }
}

dependencies {
    // latest commit is 66ea
    compile 'com.github.stathissideris:ditaa:66ea2102b5'
}
```

aadrian:
@thetric thank you for the details.

The main problem that I see right now is that GitHub is not the main repo for this project and it doesn't seems like it's used to pull requests from it (see long standing PRs).

thetric:
@arnaudroques ping?

@aadrian I would also suggest to move the Swing UI to a separate module

aadrian:
> I would also suggest to move the Swing UI to a separate module

@thetric good point! I updated the proposal to include an UI module too: https://github.com/plantuml/plantuml/wiki/Module-Proposal

The main issue right now seems to be that Git (and GitHub is not used) as a primary VCS so PRs do not seem to be integrated but (copied).

thetric:
@aadrian Would you still like to migrate the build system? I'd like to help :) I've already added some missing library links in your module proposal.
Some GitHub PRs are accepted by @arnaudroques, so we have good chances to be merged too

aadrian:
@thetric I don't think that the author wants to migrate the project. He is not using the PR feature:
https://github.com/plantuml/plantuml/pulls

GitHub and GIT seem to be just a "backup".

soloturn:
@aadrian lets make easy to digest, and small pull requests.

soloturn:
[...]

PR #93, opened 2018-01-02
Gradle migration

thetric:
Migrates from Ant and Maven to Gradle. This in an initial PR for #47 .

This might also address some issues because of inlined dependencies mentioned in #91. I have replace the source for `org.brotli` and `com.google.zxing` with Maven deps.

Note that due the the zxing update the app has Java 9 problems as it cannot display PNGs. This is due to a bug in zxing's dependency (see this issue jai-imageio/jai-imageio-core#24).


thetric:
Perhaps we can drop the zxing dependency completely? The rare use cases where a QR code is generated (generally when svek/graphviz crashed and we want to print some info) don't justify the dependency. I think a text message and a link should be sufficient. And not everyone has a QR code scanner

thetric:
Closing because of lacking interest by the maintainer :cry:

soloturn:
this is done now, see #47 .

Issue #130, opened 2018-06-15
(timing diagram) allow to change arrow color 


Shadoxtwo:
I would like to change the color for specific arrows in my timing diagrams (just like in sequence diagrams).
Could this be implemented please?

bastiedotorg:
i prepared a patch to enable this functionality:
* default behavior -> takes  ColorParam.arrow
* each arrow can be styled with PartA -[#blue]> PartB
same syntax as with class diagrams.

PR https://github.com/plantuml/plantuml/pull/131

The-Lum:
Hi all,

_[This is an Issue Review]_
This is now fixed.
- Thanks @Shadoxtwo for the request.
- Thanks @bastiedotorg for the PR: #131.
- Thanks PlantUML team for your work.

Regards.

PR #131, opened 2018-06-15
[Timing Diagrams] Enable arrow styling

bastiedotorg:
Hi,
wrt #130 , i prepared a patch to enable this functionality:
* default behavior -> takes  ColorParam.arrow
* each arrow can be styled with PartA -[#blue]> PartB
same syntax as with class diagrams.


arnaudroques:
Thanks for your code !

PR #151, opened 2018-11-22
Allow "hide footbox" in timing diagrams (ignoring it).

sdettmer:
http://forum.plantuml.net/8089/hide-footbox-not-recognized-in-timing-diagrams:

  When placing "hide footbox" at the beginning of a timing diagram, a
  syntax error is reported.

  This "hide footbox" is common to all our diagrams (plantuml.cfg file).

  ------

  We have a similar problem with 'hide footbox' in the global config
  file; the syntax error is not related to the config file though, you can
  place 'hide footbox' directly in the diagram and will get the syntax
  error


arnaudroques:
Thanks your contribution!

PR #161, opened 2018-12-22
changing output from "error" stream to "output" stream as it is

boessu:
basically not an error but an information. It seems as
[qjebbs/vscode-plantuml] gets a problem with that because he gets an error that way.
https://github.com/qjebbs/vscode-plantuml/issues/181


Simulant87 (approved):

arnaudroques:
We are going to simply remove this log.
So I close the PR.

PR #162, opened 2018-12-22
Fixed curly braces and multiple spaces not showing up when compiling to latex

freundTech:
Currently when compiling to latex (-tlatex) curly braces are not escaped and are therefor removed when the latex is compiled. This commit fixes this.
Also fixes mutliple spaces after each other showing up as only one space when compiling to latex


arnaudroques:
It seems fixed in latest releases.
Please tell us if it's not working for you.

PR #173, opened 2019-02-06
Add ability to use any shape from USymbol in nwdiag elements.

qwazer:
Example:
```
@startuml
nwdiag {
  network inet {
	  web01 [shape = cloud]
  }
}
@enduml
```
image


arnaudroques:
Simple and efficient! Thanks for your contribution !

PR #187, opened 2019-04-04
Bugfix/puml 465 support image src with url

mgriffel:
If the image src attribute contains an URL then the Filesystem#getFile() methods throwns an IOExcpetion("the filename, directory name, or volume label syntax is incorrect") because the filename is constructed using the current working directory + image src attribute which is not valid for all file systems.

This is a fix for https://avono-support.atlassian.net/browse/PUML-465 which also contains some samples to reproduce the problem.


arnaudroques:
I'm not sure to fully understand issue described on https://avono-support.atlassian.net/browse/PUML-465

However, if we may help by adding some new options on the core plantuml library side, for example by adding some specific directories for searching files at other places than the current working directory, just tell us.

Thanks!

mgriffel:
Hi Arnaud, 
I still do not know why the problem in PUML-465 only occurs in Confluence and not in the standalone mode. I decided to close this merge request and will do more analysis here and then come back with a fix/solution. 
Just a question: I would like to add some junit tests to the core plantuml library. Do you see any issues with the new dependency to junit (scope test)?

swmatisa:
Hi Arnaud,
I don't know if it is the same (same error message), but I have some problems with proxy.

I do some debug and I found a solution:
In AtomImg line 100:
		try {
			final File f = FileSystem.getInstance().getFile(src);
			if (f.exists() == false) {
				// Check if valid URL
				if (src.startsWith("http:") || src.startsWith("https:")) {
					// final byte image[] = getFile(src);
					return build(src, fc, new URL(src), scale);
				return AtomText.create("(File not found: " + f.getCanonicalPath() + ")", fc);
				}

FileSystem.getInstance().getFile Generate the Exception. If we change the order, it is OK:
		try {
			// Check if valid URL
			if (src.startsWith("http:") || src.startsWith("https:")) {
				// final byte image[] = getFile(src);
				return build(src, fc, new URL(src), scale);
			}
			final File f = FileSystem.getInstance().getFile(src);
			if (f.exists() == false) {
				return AtomText.create("(File not found: " + f.getCanonicalPath() + ")", fc);
			}

arnaudroques:
@swmatisa I'm not sure it's related. But anyway your code proposal is more logical than the current one so will be integrated in next release. You can test it on last beta http://beta.plantuml.net/plantuml.jar
Thanks for your contribution!

arnaudroques:
@mgriffel 

> Just a question: I would like to add some junit tests to the core plantuml library. Do you see any issues with the new dependency to junit (scope test)?

It depends on how the junit tests are added. Maybe you can start with a single test, so that we can see the impacts ?

PR #209, opened 2019-06-01
added ConditionEndStyle skinparam 

jbuhacoff:
Allowed values are "diamond" (default) and "hline" which uses horizontal line instead of diamond merge symbol at endif. See also the [feature request](https://forum.plantuml.net/4015/its-possible-to-draw-if-else-endif-without-merge-symbol).


arnaudroques:
Congratulation and thanks for your code !
The whole conditional management need a deep refactoring, but we keep that for latter...

PR #245, opened 2019-08-19
fix #244

### LINKED ISSUE #244: CreoleHorizontalLine nedd implements splitInTwo method
CreoleHorizontalLine nedd implements splitInTwo() method. otherwize it will invoke AbstractAtom#splitInTwo throw UnsupportedOperationException exception in some case.for example, note with ---- split in usercase with some style.

leechor:
fix #244 and add .ignore


arnaudroques:
Thanks for your patch !

PR #296, opened 2020-02-28
Added %upper/%lower to built-in macro functions

alfredporter:


arnaudroques:
Many thanks for your contribution !

PR #297, opened 2020-03-04
Fix timing diagrams with more than 1000 ticks

Araxor:
Fixes timing diagrams with more than 1000 ticks. Yes, I do have very long diagrams :smile: 

Here is an example diagram that is broken after 1000 ticks: 
http://www.plantuml.com/plantuml/svg/JOt12KCX40NlMyNZLw0_3MWDgM0nIn0sAkucnEvZRlyumy0aSvx-VYbPOHL4iE4MCEfNr8XARwMQuCZuo5pGtfuOoalcyKypKGgKKGrtPvUuEPxXgavAqXvxyGC0

The fix is done in `TimingRuler.getNbTick` method where I have removed the `boolean capped` parameter that was unused and the `Math.min` call that allowed a maximum of only 1000 ticks. It seems both have been added at the same time, but they seem useless. 

I have also changed the return type of the method to `long`, so we can have more than 2147483647 ticks.

By the way, thanks for this amazing tool!

### DIFF
--- a/src/net/sourceforge/plantuml/timingdiagram/TimingRuler.java
+++ b/src/net/sourceforge/plantuml/timingdiagram/TimingRuler.java
@@ -115,16 +115,16 @@ public class TimingRuler {
 		return highestCommonFactorInternal;
 	}
 
-	private int getNbTick(boolean capped) {
+	private long getNbTick() {
 		if (times.size() == 0) {
 			return 1;
 		}
 		final long delta = getMax().getTime().longValue() - getMin().getTime().longValue();
-		return Math.min(1000, (int) (1 + delta / tickUnitary()));
+		return 1 + delta / tickUnitary();
 	}
 
 	public double getWidth() {
-		return getNbTick(false) * tickIntervalInPixels;
+		return getNbTick() * tickIntervalInPixels;
 	}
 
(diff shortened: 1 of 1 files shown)


arnaudroques:
Thanks for your contribution.
However, there is an issue...
There *must* be a limit for the number of ticks.
For example at: https://github.com/plantuml/plantuml/blob/7a5b515bf6cf87e2af2ec4691bf955ce8d031c34/src/net/sourceforge/plantuml/timingdiagram/TimingRuler.java#L161

It's not reasonable to run this `for` loop up to several millions. 
So we suggest to increase the limit maybe to 10,000 instead of 1,000.
Would that be enough for you ?

Araxor:
Here is my use case: I generate timing diagrams from some logs. The tick unit is milliseconds. Then I embed them as SVG in a web page and use the scroll bar to navigate in the diagram.

For this use case, having the possibility to have a very large number of ticks is reasonable. I want to be able to generate diagrams from possibly very long logs.

I am not sure I understand why there _should_ be a limit at all, except maybe Long.MAX_VALUE. If the user wants to spend hours in generating a diagram, it's his problem, no?

Of course, maybe there is a very good reason that I am not aware of. In that case, I am perfectly happy with using a forked version of PlantUML without a limit.

PR #348, opened 2020-07-06
Move license message below getting started

matschaffer:
Having it as the first line made me think I was supposed to accept an EULA or something (for example, add `license` to a diagram so it would save something in my home directory).

I thought moving it to the bottom might help clear up that this is just "empty diagram" behavior.


matschaffer:
Resulting image

thing

Issue #432, opened 2020-12-31
[gantt] thursday are colored in #AAF


liuzhaowei55:
How to do this?

Nordes:
As of now, in the code we can see that we can only mark the day "as closed". 

![image](https://user-images.githubusercontent.com/446572/103963775-118e1d00-5128-11eb-85ae-327ba87f5b95.png)

src/net/sourceforge/plantuml/project/lang/SubjectDayOfWeek.java

In order to specify color like `2021-01-04 to 2021-01-10 are colored in #d2d2d2`, the author (or someone) would have to add the "NLP" part. My guess is that it would be not "so" complicated.

Nordes:
I've done a PR, so maybe it will be merged soon so you can use the feature ;)

![image](https://user-images.githubusercontent.com/446572/103971159-d4318b80-5137-11eb-8edf-c90bb1569f7c.png)

Nordes:
@liuzhaowei55 this will normally be part of the next version of plantuml (merge request was completed)

arnaudroques:
It's [available online](http://www.plantuml.com/plantuml/uml/LP0nQyD034Rt-nKVT0xa3DdZePaB7Os7htTAxFIgCoUPud_Vu0PADMZIUmzqDcVp6zKT-mvBczKjiGWovxA5yDRgNP9ZFpNu98Zcg5VSgIlRXkxOnM2hPcvW4wHQQfECMV7IF-Po2M7eLwzVz3dXer7J1ESDLT6a24r6q31O79-TNNeunaCyWPfHQ6w8-1Rv33w9OdWNNnVyUViY3_0fwiR_I1XUDO-Ok2oYXdWw8dCpyEhIVflXB9htB_m0), and on the last beta http://beta.plantuml.net/plantuml.jar

Thanks to @Nordes for the code !

Nordes:
You're welcome

liuzhaowei55:
@Nordes @arnaudroques Thank you so much for solving this problem so quickly and getting online. I use plantuml in my DoukuWiki.

ctr:
I'm not sure where to leave my comment, but just to say this is a great idea but it isn't working in monthly view for me. Could this be added too?

[...]

PR #439, opened 2021-01-08
432-Implement "[day] are colored in #AAAAFF"

Nordes:
As suggested in #432, I think it would be a nice addition to be able to color the day as we wish. It might be something recurring (such as weekend is a different day for some people or whatever reason).

## Usage
@startgantt gantt
printscale daily

Project starts the 1st of january 2021
sunday are colored in #AAAAFF

[Automatic branch tag on release] as [s1] starts 2021-01-15 and lasts 1 week
then [Setup Automatic Alerts] as [s2] lasts 1 week

[End] happens 150 days after start
@endgantt

## Result
image


Nordes:
I added a wrong commit inside the PR :(. Restarting

PR #448, opened 2021-01-20
Change link decorators to fill with diagram background color instead of white

matthew16550:
Hi

It's my first contribution to PlantUML though I've been using it for many years.  It's such a fun and useful tool, thanks so much for making it.

This PR changes some link decorators to fill using the root background color instead of always using white, as shown below.

Many decorators already use the root background color so I think the hard coded`WHITE` values are just older code that has not been updated?

### Class Diagram

```plantuml
@startuml

skinparam BackgroundColor grey
skinparam ClassBackgroundColor grey
skinparam Shadowing false

foo <-- bar
foo +--o bar
foo #--* bar
foo <|--o{ bar
foo <||--^ bar

@enduml
```

**Before:**
class_diagram

**After:**
class_diagram_after

### Component Diagram

```plantuml
@startuml

skinparam ArrowLollipopColor yellow
skinparam BackgroundColor grey
skinparam ComponentBackgroundColor grey
skinparam Shadowing false

[foo] -- [bar]
foo 0--0 bar
foo 0)--(0 bar
foo #--# bar
foo -(0)- bar
foo -0)- bar
foo -(0- bar

@enduml
```

**Before:**
component_diagram

**After:**
component_diagram_after

### State Diagram

```plantuml
@startuml

skinparam BackgroundColor grey
skinparam Shadowing false
skinparam StateBackgroundColor grey

[*] -->o foo
foo x--> foo

@enduml
```

**Before:**
state_diagram

**After:**
state_diagram_after

### DIFF
--- a/src/net/sourceforge/plantuml/cucadiagram/LinkDecor.java
+++ b/src/net/sourceforge/plantuml/cucadiagram/LinkDecor.java
@@ -111,7 +111,7 @@ public enum LinkDecor {
 	public ExtremityFactory getExtremityFactory(HColor backgroundColor) {
 		switch (this) {
 		case PLUS:
-			return new ExtremityFactoryPlus();
+			return new ExtremityFactoryPlus(backgroundColor);
 		case REDEFINES:
 			return new ExtremityFactoryExtendsLike(backgroundColor, false);
 		case DEFINEDBY:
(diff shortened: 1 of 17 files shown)


arnaudroques:
> Many decorators already use the root background color so I think the hard coded `WHITE` values are just older code that has not been updated?

Exactly !
Many thanks for your contribution: this will be integrated in next official release :-)

PR #450, opened 2021-01-21
Change Activity Diagram "Stop" and "End" symbols to fill using the diagram background color instead of always using white

matthew16550:
Similar to #448, this PR removes Activity Diagram hard coded `WHITE` colors.

```plantuml
@startuml

skinparam ActivityEndColor red
skinparam ActivityStartColor red
skinparam BackgroundColor grey
skinparam Shadowing false

start
end
stop

@enduml
```

**Before**
activity_diagram_before

**After**
activity_diagram_after


matthew16550:
This might be causing an NPE, converted PR to draft while I investigate.

matthew16550:
Never mind about the NPE, it seems that my IntelliJ PlantUML plugin was being strange 😆 .

arnaudroques:
Thanks again !

PR #462, opened 2021-02-12
Set Allow Include by default

hoggmania:


arnaudroques:
For security reason, we did not merge this PR.
Tell us if it's an issue for you

PR #463, opened 2021-02-12
Add "POST /render" endpoint to PicoWeb server.

matthew16550:
It expects a JSON body like:
```json
{
  "source": "PLANTUML_SOURCE_STRING",
  "options": ["-tpdf"]
}
```

Most of the CLI options should work here, including all output formats.

I made this to help speed up the [Nikola PlantUML plugin](https://plugins.getnikola.com/v8/plantuml/) which currently runs PlantUML once per file.  Using this new endpoint reduces its render time for 10 small files on my laptop from roughly 37s to 2s.

Extending Picoweb seems the best approach for the Nikola work because it has no dependencies.

No strong opinion about `/render` as the path, other suggestions welcome.

### DIFF
--- a/src/net/sourceforge/plantuml/FileFormat.java
+++ b/src/net/sourceforge/plantuml/FileFormat.java
@@ -60,8 +60,37 @@ import net.sourceforge.plantuml.ugraphic.UFont;
  * 
  */
 public enum FileFormat {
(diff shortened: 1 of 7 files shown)


arnaudroques:
Thanks for your contribution.
We just hope that this tiny web server will not because a large one :-)

PR #478, opened 2021-03-06
Fix interactive functionality for Usecase diagrams exported to SVG

ioplker:
Hello, again! So I fixed JS code (yes, I know, it's stupid :man_facepalming:).
I guess, the error did not came up cause of some java caching - now every time I will remove `target` directory before rebuilding the project.

Can't wait to bring the joy of interactivity to class diagrams and others :wink: 
So let me know if I can apply same mechanism (with `<g>` elements) further.


arnaudroques:
So we have just released V1.2020.2. 
Don't be surprised : the feature is disabled on the official jar release. You just have to modify [a flag in the source code](https://github.com/plantuml/plantuml/blob/master/src/net/sourceforge/plantuml/svg/SvgGraphics.java#L126).
We did this because we do not want to change SVG generation without telling users. This flag is really temporary.

We have also slightly changed the way to start groups. ([Example here](https://github.com/plantuml/plantuml/blob/master/src/net/sourceforge/plantuml/svek/image/EntityImageDescription.java#L294))

You can now go on with your modifications on this version !

We really need to start a discussion with all users to know how this feature will be used in the future:
- do people want it be default ?
- do they want an option flag in the commande line ? In the diagram itself ?

We think that your contribution opens a  great path for many new features, and that's really nice. Our idea is to allow other users to build their own .css and .js and to integrate it into plantuml.jar
We are going to open a new thread on the forum and on twitter to see what people think about it.

Many thanks for your work !

ioplker:
> So we have just released V1.2020.2.
> Don't be surprised : the feature is disabled on the official jar release. You just have to modify [a flag in the source code](https://github.com/plantuml/plantuml/blob/master/src/net/sourceforge/plantuml/svg/SvgGraphics.java#L126).
> We did this because we do not want to change SVG generation without telling users. This flag is really temporary.
> 
> We have also slightly changed the way to start groups. ([Example here](https://github.com/plantuml/plantuml/blob/master/src/net/sourceforge/plantuml/svek/image/EntityImageDescription.java#L294))
> 
> You can now go on with your modifications on this version !
> 
> We really need to start a discussion with all users to know how this feature will be used in the future:
> 
>     * do people want it be default ?
> 
>     * do they want an option flag in the commande line ? In the diagram itself ?
> 
> 
> We think that your contribution opens a great path for many new features, and that's really nice. Our idea is to allow other users to build their own .css and .js and to integrate it into plantuml.jar
> We are going to open a new thread on the forum and on twitter to see what people think about it.
> 
> Many thanks for your work !

:tada: :sparkles:
You are welcome :)

About the switch. I did think about that and in my opinion it is best to toggle interactive svg via commandline flag. Maybe with `-tsvgi` (the **i** is for *interactive*). *At first sight, though*.

**But**. As contributors now can make *their own modes* via adding `css` and `js` files, I think we should consider the mechanism (interface) for simple creating and choosing this interactive mode - and I suppose we do not want to have many-many flags like `-tsvgi_1, -tsvgi_2` etc. So in that case we can have flag for *choosing interactive* mode, e.g. `-SvgInter=default, -SvgInter=linked` etc.

---
To sum it up we (and PlantUML users) need to answer this questions:
- Is interactive SVG useful at all? (just to be clear)
- Is someone really depend on the old SVG file format? (to find out the default behavior of the SVG exporter)
- What fundamental differences can interactive modes have? (to find out the need to make several of them - if yes, how should we implement switching between them)
- Are there any bugs or feature requests for this first interactive mode? (check-up question for previous one)
- What default behavior do users want? (specific interactive mode or default non-interactive)

---
Now when I wrote all of this I think setting in diagram itself (e.g. *skinparam*) may be the best way to choose interactive mode :stuck_out_tongue: 

---
P.S.
I'll continue with other diagram types for now, thanks for refactor of `startGroup` :)

arnaudroques:
The thread is open on https://forum.plantuml.net/13350/add-interactive-functionality-usecase-diagrams-exported

You can also answer there, if you wish.

DmitryMaletin:
Is the grouping available in the last release?

ioplker:
> Is the grouping available in the last release?

Hi, @DmitryMaletin!
As @alfredporter pointed out, you have to enable [flag in a source code](https://github.com/plantuml/plantuml/blob/master/src/net/sourceforge/plantuml/svg/SvgGraphics.java#L128) to turn this feature on.

I've just tried [last release](https://github.com/plantuml/plantuml/releases/tag/v1.2021.7) and can confirm it works.

P.S.
I suppose you meant this interactive feature the thread is about, correct me if I'm wrong.

PR #479, opened 2021-03-08
Enable CSS control of diagram margin for many diagram types.

matthew16550:
I'm not sure if this kind of PR is helpful, it might just make a big merge conflict with the work done outside GitHub?

This enables CSS control of margins for Board, Gantt, JSON, Mind Map, Network, Salt, Sequence, Timing, WBS & YAML diagrams.

And also tidies some other repeated margin code.
(All the logic for diagram margins is moved to `ImageParameter.calculateDiagramMargin()`)

### DIFF
--- a/src/net/sourceforge/plantuml/ugraphic/ImageParameter.java
+++ b/src/net/sourceforge/plantuml/ugraphic/ImageParameter.java
@@ -40,9 +40,15 @@ import net.sourceforge.plantuml.CornerParam;
 import net.sourceforge.plantuml.ISkinParam;
 import net.sourceforge.plantuml.LineParam;
 import net.sourceforge.plantuml.SvgCharSizeHack;
+import net.sourceforge.plantuml.TitledDiagram;
+import net.sourceforge.plantuml.UseStyle;
 import net.sourceforge.plantuml.anim.Animation;
 import net.sourceforge.plantuml.skin.rose.Rose;
 import net.sourceforge.plantuml.style.ClockwiseTopRightBottomLeft;
+import net.sourceforge.plantuml.style.PName;
+import net.sourceforge.plantuml.style.SName;
+import net.sourceforge.plantuml.style.Style;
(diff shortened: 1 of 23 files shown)


src/net/sourceforge/plantuml/ugraphic/ImageParameter.java
```diff
@@ -84,15 +90,21 @@ public ImageParameter(ColorMapper colorMapper, boolean useHandwritten, Animation
 		this.lengthAdjust = LengthAdjust.defaultValue();
 	}
 
-	public ImageParameter(ISkinParam skinParam, Animation animation, double dpiFactor, String metadata,
-			String warningOrError, ClockwiseTopRightBottomLeft margins, HColor backcolor) {
+	public ImageParameter(TitledDiagram diagram, ColorMapper colorMapper, boolean useHandwritten, Animation animation, double dpiFactor,
```
matthew16550, line 93: This new constructor will hopefully be  short lived.

I think some similar refactorings could end up with a simple constructor like `ImageParameter(TitledDiagram diagram)`.  The logic for backcolor, dpiFactor etc could all be moved into `ImageParameter` so they behave the same for more diagram types.

arnaudroques:
> I'm not sure if this kind of PR is helpful, it might just make a big merge conflict with the work done outside GitHub?

It is definitively helpful. GitHub and our internal SVN server are synchronized for each official release, so right now, there is no conflict.

> And also tidies some other repeated margin code.

Here I feel ashamed: it's something we should have done long time before...
 
> This enables CSS control of margins for Board, Gantt, JSON, Mind Map, Network, Salt, Sequence, Timing, WBS & YAML diagrams.

That's great. Our intention is really that all diagrams use <style>. So finally, in some future, the `UseStyle` class be removed since `useBetaStyle()` will always return `true`. We know it's going to take months/years, but we are slowly but surely moving toward this objective.

At that future moment, we could remove all methods from `ISkinParam` that deals with colors and fonts.

PR #480, opened 2021-03-09
Refactor duplicated logic into ImageParameter class

matthew16550:
Following on from #479.

This will provide consistent support for Diagram Border, Handwritten, Background Color, Metadata (and possibly more) in many diagrams that previously did not support them.

### DIFF
--- a/src/net/sourceforge/plantuml/activitydiagram3/ActivityDiagram3.java
+++ b/src/net/sourceforge/plantuml/activitydiagram3/ActivityDiagram3.java
@@ -226,15 +226,13 @@ public class ActivityDiagram3 extends UmlDiagram {
 		final double dpiFactor = getDpiFactor(fileFormatOption,
 				Dimension2DDouble.delta(dim, margins.getLeft() + margins.getRight(), 0));
 
-		final HColor backcolor = skinParam.getBackgroundColor(false);
-		final String metadata = fileFormatOption.isWithMetadata() ? getMetadata() : null;
-		final ImageParameter imageParameter = new ImageParameter(this, getAnimation(), dpiFactor, metadata,
-				getWarningOrError(), backcolor);
+		final ImageParameter imageParameter = new ImageParameter(this, fileFormatOption, getAnimation(), dpiFactor,
+				getWarningOrError());
 
 		final ImageBuilder imageBuilder = ImageBuilder.build(imageParameter);
 		imageBuilder.setUDrawable(result);
 
-		return imageBuilder.writeImageTOBEMOVED(fileFormatOption, seed(), os);
+		return imageBuilder.writeImageTOBEMOVED(seed(), os);
 
 	}
 
--- a/src/net/sourceforge/plantuml/board/BoardDiagram.java
+++ b/src/net/sourceforge/plantuml/board/BoardDiagram.java
@@ -55,10 +55,6 @@ import net.sourceforge.plantuml.core.ImageData;
 import net.sourceforge.plantuml.graphic.InnerStrategy;
 import net.sourceforge.plantuml.graphic.StringBounder;
 import net.sourceforge.plantuml.graphic.TextBlock;
-import net.sourceforge.plantuml.style.PName;
-import net.sourceforge.plantuml.style.SName;
-import net.sourceforge.plantuml.style.Style;
-import net.sourceforge.plantuml.style.StyleSignature;
 import net.sourceforge.plantuml.svek.TextBlockBackcolored;
 import net.sourceforge.plantuml.ugraphic.ImageBuilder;
 import net.sourceforge.plantuml.ugraphic.ImageParameter;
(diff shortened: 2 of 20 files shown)


matthew16550:
@arnaudroques I would like to refactor a little more in this PR but need some advice first:

### animation
Roughly half the code uses `diagram.getAnimation()` and the other half uses `null`.  I don't understand this bit well enough to decide, are we safe to use `diagram.getAnimation()` everywhere?

### warningOrError
Roughly half the code uses `diagram.getWarningOrError()` and the other half uses `null` or `""`.  I think we can always use `diagram.getWarningOrError()` ?
 
### dpiFactor
This is calculated several ways, are we able to simplify to one algorithm?
* Many places hard code it to `1.0`
* In `BoardDiagram`, `MindMapDiagram`, `PSystemSalt`, `WBSDiagram`, `WireDiagram`:
```java
final Scale scale = getScale();
final double dpiFactor = scale == null ? getScaleCoef(fileFormatOption) : scale.getScale(100, 100);
```
* In `GanttDiagram`, `GitDiagram`, `JsonDiagram`, `NwDiagram`:
```java
final Scale scale = getScale();
final double dpiFactor = scale == null ? 1 : scale.getScale(100, 100);
```
* In `SequenceDiagramFileMakerPuma2`:
```java
scale = getScale(area.getWidth(), area.getHeight());
final double dpiFactor = diagram.getScaleCoef(fileFormatOption);
final double factor = oneOf(scale, dpiFactor);
```
* In `ActivityDiagram3`:
```java
final double dpiFactor = getDpiFactor(fileFormatOption,
		Dimension2DDouble.delta(dim, margins.getLeft() + margins.getRight(), 0));

private final double getDpiFactor(FileFormatOption fileFormatOption, final Dimension2D dim) {
	final double dpiFactor;
	final Scale scale = getScale();
	if (scale == null) {
		dpiFactor = getScaleCoef(fileFormatOption);
	} else {
		dpiFactor = scale.getScale(dim.getWidth(), dim.getHeight());
	}
	return dpiFactor;
}
```
* In `CucaDiagramFileMakerSvek`:
```java
final Dimension2D dim = result.calculateDimension(stringBounder);
final double scale = getScale(fileFormatOption, dim);

private double getScale(FileFormatOption fileFormatOption, final Dimension2D dim) {
	final double scale;
	final Scale diagScale = diagram.getScale();
	if (diagScale == null) {
		scale = diagram.getScaleCoef(fileFormatOption);
	} else {
		scale = diagScale.getScale(dim.getWidth(), dim.getHeight());
	}
	return scale;
}
```

arnaudroques:
> @arnaudroques I would like to refactor a little more in this PR but need some advice first:
> 

Sure.
However, it would be easier for us if you create two new PR, instead of refactoring this one.
- one about animation and warningOrError
- another one about dpiFactor

> ### animation
> Roughly half the code uses `diagram.getAnimation()` and the other half uses `null`. I don't understand this bit well enough to decide, are we safe to use `diagram.getAnimation()` everywhere?
> 

Short answer: yes, we should use `diagram.getAnimation()` everywhere.

This is an unfinished and undocumented feature which would allow to apply affine transformation on diagram and to generated animated images.
For example, [you can have](http://www.plantuml.com/plantuml/uml/SoWkIImgAStDuL8iAKhCAqxBByfDBCdCprCeoYz9B4bLC37bIiv9B2vMIClFDqGmZRWSKlDIWFe0):

```
@startuml
!transformation rotate 45
class foo1
class foo2
@enduml
```

As I said, this is really pending and not really working.
So there is no risk at calling `diagram.getAnimation()` everywhere.


> ### warningOrError
> Roughly half the code uses `diagram.getWarningOrError()` and the other half uses `null` or `""`. I think we can always use `diagram.getWarningOrError()` ?
> 

Agree, we can always use `diagram.getWarningOrError()`.



> ### dpiFactor
> This is calculated several ways, are we able to simplify to one algorithm?
> 

**This part is really more complex, so I think it deserves a separate PR.**
Over time, our strategy has changed on scale and dpi management, that's why there are different versions.
And now, I must say that we are completely lost about the right way to manage scale and dpi :-)
The difficulty is that scaling/zooming/dpi can be modified either in `FileFormatOption` or by `skinparam dpi` or by `scale` [directive in the diagram](https://plantuml.com/en/commons#4252b72e6ebcd4d4).

The right approach is _probably_ the one used in `ActivityDiagram3` / `CucaDiagramFileMakerSvek`
We should also probably move `getScaleCoef()` method from  `TitledDiagram` to `AbstractPSystem` and also give a better name to this method.

So your idea are welcome!

matthew16550:
Ok, I will make new PRs as suggested.
Scale will need some thinking about!

matthew16550:
I'm hoping to go further with the refactoring.  Most of the logic in `exportDiagramInternal()` seems to be repeated in most classes and I think after we extract the scale details then a common pattern will be fairly obvious and we can extract it to a new method.

PR #485, opened 2021-03-11
TestPlayground

matthew16550:
I've been experimenting with this idea while refactoring `ImageParameter`, it's been helpful to batch render various diagrams with different settings then compare in a grid.  Thought I'd suggest it here to see if other devs would make use of it.

As an example, below is the current situation for `scale` & `dpi` support.  The first row uses neither, the second row has `scale 2` and the last row has `skinparam dpi 192`.  We already know these are not consistently implemented, the picture makes the details clearer and re-rendering it will give quick feedback as I adjust that code.

sizing

It was created by:
```java
ALL_ALLOWING_SKINPARAM

  // Make PNGs in the "no_style" dir
  .render(PNG, "no_style")
  
  // Make PNGs in the "scale_2" dir using "scale 2"
  .render(PNG, "scale_2", "scale 2")
  
  // Make PNGs in the "dpi_192" dir using "skinparam dpi 192"
  .render(PNG, "dpi_192", "skinparam dpi 192")
  
  // Make "sizing.png" (the image above) comparing PNGs from "no_style", "scale_2" & "dpi_192" dirs
  .pngHorizontal("sizing", "no_style", "scale_2", "dpi_192");
```

My earlier attempt tried to specify the same sort of thing in a JSON file but I quickly found it was not flexible so changed to this idea which is just some Java methods that can be joined together in various ways.

The target users here are other PlantUML developers so I haven't thought about making it end-user friendly with CLI support etc.


arnaudroques:
Of course, adding test to PlantUML is a nice idea.
However, for many reasons, we would like that tests would be hosted on a separate project.
I understand that this is not completely standard and that it has some drawbacks.

So I suggest that you use https://github.com/plantuml/plantuml-test/ for this PR.

Does it sound good to you ?

matthew16550:
Sure I totally sympathise, deciding these sorts of things can be very difficult.

For this PR a separate repo would work fine as it's a stand-alone idea.

But for more general purpose testing it could add a lot of overhead and discourage people from contributing.  

If source & tests are together in one repo then refactoring is simpler, regression testing can easily be part of the build process and a single PR can discuss code alongside the relevant tests.

I fear that tests in a separate repo would become stale as the main code changes.

What would the dev workflow be for a small fix or for a new feature?

arnaudroques:
> Sure I totally sympathise, deciding these sorts of things can be very difficult.
> 
> For this PR a separate repo would work fine as it's a stand-alone idea.
> 
Ok, so let's start this way for this PR.

> But for more general purpose testing it could add a lot of overhead and discourage people from contributing.
> 
> If source & tests are together in one repo then refactoring is simpler, regression testing can easily be part of the build process and a single PR can discuss code alongside the relevant tests.
> 
> I fear that tests in a separate repo would become stale as the main code changes.
> 
> What would the dev workflow be for a small fix or for a new feature?

I am also working on "regular" project where `test` and `src` are on the same repo, so I fully agree with you and I know the advantage.

**But** there is an aspect that you should know: PlantUML is used in many different configurations.
We are in touch with several users and you will be surprise that most of them are not Java familiar. Some of them even hates Java BTW :-) They _only_ install Java because they want to use PlantUML.

For security reason, some of them want to build PlantUML from the official source.
**So a very important goal is to have the simplest build possible for those guys**, especially when they do not know Java.
That's why we don't use any external lib. You don't even need `ant` to build PlantUML (nor `maven` )

Properly adding tests to the repo would mean adding a dependency to `JUnit` or `TestNG` and we definitively want to avoid this in the core library.

There are many possible options to respect all these constraints. For example, I could think about one first repo that would contains both `test` and `src` (so work for contributor is easy). Any release to this repo would commit to another repo that would only have `src` so that people could simply build from this last one. This could even manage our multi-license source generation.

As conclusion, we know that we must improve all our code management, test and pipeline. We are thinking about it.  
But in the short term, we won't change it right now.

matthew16550:
Ok, I'll move to the other repo.

I've worked in organisations with that sort of security focus so understand.  Guess they just do something like `javac` then `jar`?

Not pushing for it now but I wonder if adding a `test` dir here wouldn't actually change what those orgs are doing.  Maven is already here, it could even provide deps for testing.  Those orgs would just ignore the test stuff.

arnaudroques:
Another contributor has pushed some CI files : I guess that's something we are really going to use in some future. :-)

About your PR, I think that the new method `setSecurityProfile()` is an issue. For security reason, we want it impossible to change SecurityProfile once PlantUML is launched.

An acceptable option would be that `setSecurityProfile()` would work only if the initial profile is `UNSECURE`. Otherwise, it would raise an exception. This way:

- It garanties for users that start PlantUML with profile `INTERNET`/`LEGACY`/`ALLOWLIST`/`SANDBOX ` that there is now way someone can change their security profiles to something unsafe.
- You could launch tests starting with `UNSECURE` profile. This mode allow you to change back and forward to any profiles, which is not an issue for users.

Does it sound good ?

matthew16550:
Yep I've been following #488 - we might want to merge it all into one job once things settle down.

This one does experimental publishing of snapshot builds (currently to my fork - https://github.com/matthew16550/plantuml/packages/669397)
though I haven't yet convinced myself if they are helping the test code or just making things confusing.

`setSecurityProfile()` for my test was only so I didn't need to set `UNSECURE` in the env var 😆 .  Your idea makes sense for future test cases but we don't need it now, I'll just set the env var instead.

matthew16550:
Closing in favour of #492.

The CI pipeline here was an experiment and we don't need to keep it as #488 also has a pipeline.

PR #488, opened 2021-03-14
[Draft] Publish plantuml Dockerimage upon every new release

mambax:
# Flow
1. You create and publish a release, this triggers the GitHub action
2. The action does build the new app, dockerizes it and pushes it to docker-hub

# Notes
I added a draft of both files, there are some brushings necessary from inputs from your side

- How do you properly build your app?
- We can either use the release name (tag) or the maven version as input. I expect them to be the same but extracting it from maven [like so](https://blog.soebes.de/blog/2018/06/09/help-plugin/) could be the better option.
- It would be possible to automate your build process ONLY with GitHub Actions but I fear (as I see the maturity of that project) that is one hell of a bigger change
- You would need to place TWO secrets into this repository or your org: `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` because I will not be able to do that from my fork (the credentials for [this account](https://hub.docker.com/u/plantuml)  - the new image would always be (fully automated) `plantuml/plantuml:<maven or release version>`

- The action uses the local files to rebuild the JAR instead of downloading it from somewhere else. I see this as cleaner than downloading it from SourceForge once more
- The app runs as non-root in my attempt to keep it securer

Anyway, I need a bit of your feedback to make it work smoother (or make it work at all)

### DIFF
--- a/.github/workflows/build_push.yml
+++ b/.github/workflows/build_push.yml
@@ -1,0 +1,44 @@
+name: Java CI with Maven
+
+on:
+  release:
+    types: [ published ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Set up JDK 14
+      uses: actions/setup-java@v1
+      with:
+        java-version: 14
+
+    - name: Build with Maven
+      run: mvn -B package --file pom.xml
+
+    - name: Extract version name
+      shell: bash
+      run: echo "##[set-output name=releaseTag;]$(echo ${GITHUB_REF})"
+      id: extract_tag
+
+    - name: Login to DockerHub
+      uses: docker/login-action@v1 
+      with:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+    - name: Build and push
+      id: docker_build
+      uses: docker/build-push-action@v2
+      with:
+        push: true
+        tags: plantuml/plantuml:${{ steps.extract_tag.outputs.releaseTag }}
+
+    - name: Image digest
+      run: echo ${{ steps.docker_build.outputs.digest }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM openjdk:14-jdk-alpine3.10
+ENV LANG en_US.UTF-8
+RUN \
+  apk add --no-cache graphviz ttf-dejavu fontconfig
+COPY target/plantuml-1.2021.3-SNAPSHOT.jar plantuml.jar
+RUN ["java", "-Djava.awt.headless=true", "-jar", "plantuml.jar", "-version"]
+RUN addgroup -S appgroup && adduser -S appuser -G appgroup
+USER appuser
+ENTRYPOINT ["java", "-Djava.awt.headless=true", "-jar", "plantuml.jar", "-p"]
+CMD ["-tsvg"]


arnaudroques:
> # Flow
> 1. You create and publish a release, this triggers the GitHub action
> 2. The action does build the new app, dockerizes it and pushes it to docker-hub
> 

That's a good start.
However, I think that a docker with `plantuml.jar` file is pretty useless to users. I think that people want the http server (that is `plantuml.war` ) which is released from https://github.com/plantuml/plantuml-server
Actually, this is what is pushed to https://hub.docker.com/u/plantuml
But as long as your contribution does not break the existing Docker images on https://github.com/plantuml/plantuml-server , it's ok for me.

> # Notes
> I added a draft of both files, there are some brushings necessary from inputs from your side
> 
> * How do you properly build your app?

We build it with both `ant` and `maven`.

> * We can either use the release name (tag) or the maven version as input. I expect them to be the same but extracting it from maven [like so](https://blog.soebes.de/blog/2018/06/09/help-plugin/) could be the better option.

Right now, I suggest that we take the simplest option.

> * It would be possible to automate your build process ONLY with GitHub Actions but I fear (as I see the maturity of that project) that is one hell of a bigger change
> * You would need to place TWO secrets into this repository or your org: `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` because I will not be able to do that from my fork (the credentials for [this account](https://hub.docker.com/u/plantuml)  - the new image would always be (fully automated) `plantuml/plantuml:<maven or release version>`

Ok, we'll do that when you think your work is ready.

> * The action uses the local files to rebuild the JAR instead of downloading it from somewhere else. I see this as cleaner than downloading it from SourceForge once more

Agree.


> * The app runs as non-root in my attempt to keep it securer
> 
> Anyway, I need a bit of your feedback to make it work smoother (or make it work at all)

Tell us if you have more questions and thanks for your work !

mambax:
> That's a good start.
However, I think that a docker with plantuml.jar file is pretty useless to users. I think that people want the http server (that is plantuml.war ) which is released from https://github.com/plantuml/plantuml-server
Actually, this is what is pushed to https://hub.docker.com/u/plantuml
But as long as your contribution does not break the existing Docker images on https://github.com/plantuml/plantuml-server , it's ok for me.

This is where I believe you are not correct :) 

I have found dozens of people using plantuml raw to render charts from code in CI:

https://hub.docker.com/r/think/plantuml/
https://hub.docker.com/r/miy4/plantuml

Sad is, instead of trying to contribute too these folks just did their own little sauce. Anyway each his own.

Also, I see more than 1 person trying to encapsulate diagram generation in CI with GHA:

https://github.com/cloudbees/plantuml-github-action
https://github.com/Timmy/plantuml-action
https://github.com/lokkju/github-action-plantuml
https://github.com/frederikheld/render-plantuml-to-wiki-action

Check their Dockerfiles, they all brewed their own sauce because there is no official plantuml docker 😄 

I want to solve that and contribute it to your hands.

arnaudroques:
> > That's a good start.
> > However, I think that a docker with plantuml.jar file is pretty useless to users. I think that people want the http server (that is plantuml.war ) which is released from https://github.com/plantuml/plantuml-server
> > Actually, this is what is pushed to https://hub.docker.com/u/plantuml
> > But as long as your contribution does not break the existing Docker images on https://github.com/plantuml/plantuml-server , it's ok for me.
> 
> This is where I believe you are not correct :)
> 
> I have found a handful of people using plantuml raw to render charts from code in CI:
>  [...]
> Check their Dockerfiles, they all brewed their own sauce because there is no official plantuml docker 😄
> 
> I want to solve that and contribute it to your hands.

Right, I understand now.

As I said, we have very little knowledge about Docker. The good news is that we want to learn.

mambax:
Update: I did not forget, I just have 7 hands full 🥵

MOZGIII:
We now have docker images:
- https://hub.docker.com/r/plantuml/plantuml
- https://github.com/plantuml/plantuml/pkgs/container/plantuml

denisa:
Hasn’t this been implemented in the meantime?

PR #492, opened 2021-03-18
Tweaks to enable "Test Playground" in plantuml-test repo

matthew16550:
This replaces #485 which has mostly moved to plantuml/plantuml-test#2.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,7 +130,12 @@
                 </configuration>
             </plugin>
             <plugin>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <version>3.1.2</version>
+            </plugin>
+            <plugin>
                 <artifactId>maven-source-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


arnaudroques:
Just to let you know that I have started to add a first JUnit test into the main library :-) 
This should *only* impact people using maven. Other people (using `ant` or something else) won't notice.
Note that I am going to be **very** conservative here.

Next official release will only contain this single test. I am really uncomfortable with maven, so I really want to move slowly.

In the mean time, I will also add a new feature that would output diagrams in some "text" description format (instead of PNG/SVG/EPS...) that would be useful for unitary tests and non-regression tests.
We could also probably reuse the `PlaygroundTest` you've built.

Thanks!

matthew16550:
👍 Thats great.

Maven can be very strange and confusing.  I've used it quite a lot, let me know if there's anything I can help with there.

PR #493, opened 2021-03-18
Handle Scale & DPI more consistently for all diagram types.

matthew16550:
The actual calculation can be deferred until `ImageBuilder.createUGraphic()` so we get to delete some code and fix some anomalies:
* GANTT, Network & Timing diagrams did not support DPI.
* Timing did not support Scale.

There are more anomalies that are not fixable here:
* GIT diagrams are a work in progress so do not support Scale or DPI yet.
* JCCKIT does not support Scale or DPI - perhaps it never will?
* Latex & Math do not support DPI - they manage Scale & DPI in their own way so a separate PR would make sense.

### Before
sizing

### After
sizing

### DIFF
--- a/src/net/sourceforge/plantuml/gitlog/GitDiagram.java
+++ b/src/net/sourceforge/plantuml/gitlog/GitDiagram.java
@@ -43,22 +43,18 @@ import java.util.Collection;
 import net.sourceforge.plantuml.AnnotatedWorker;
 import net.sourceforge.plantuml.FileFormatOption;
 import net.sourceforge.plantuml.ISkinParam;
-import net.sourceforge.plantuml.SkinParam;
 import net.sourceforge.plantuml.UmlDiagram;
 import net.sourceforge.plantuml.UmlDiagramType;
-import net.sourceforge.plantuml.UseStyle;
 import net.sourceforge.plantuml.core.DiagramDescription;
 import net.sourceforge.plantuml.core.ImageData;
 import net.sourceforge.plantuml.graphic.InnerStrategy;
 import net.sourceforge.plantuml.graphic.StringBounder;
 import net.sourceforge.plantuml.graphic.TextBlock;
-import net.sourceforge.plantuml.style.ClockwiseTopRightBottomLeft;
 import net.sourceforge.plantuml.svek.TextBlockBackcolored;
 import net.sourceforge.plantuml.ugraphic.ImageBuilder;
 import net.sourceforge.plantuml.ugraphic.ImageParameter;
 import net.sourceforge.plantuml.ugraphic.MinMax;
 import net.sourceforge.plantuml.ugraphic.UGraphic;
-import net.sourceforge.plantuml.ugraphic.color.ColorMapperIdentity;
 import net.sourceforge.plantuml.ugraphic.color.HColor;
 
 public class GitDiagram extends UmlDiagram {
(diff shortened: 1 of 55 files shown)


src/net/sourceforge/plantuml/gitlog/GitDiagram.java
```diff
@@ -93,7 +90,8 @@ protected ImageData exportDiagramInternal(OutputStream os, int index, FileFormat
 			margin1 = 10;
 			margin2 = 10;
 		}
-		final ImageParameter imageParameter = new ImageParameter(new ColorMapperIdentity(), false, null, dpiFactor, "",
+		// TODO change to other ImageParameter constructor?
+		final ImageParameter imageParameter = new ImageParameter(new ColorMapperIdentity(), false, null, "",
```
matthew16550, line 94: GIT diagrams currently do not support Scale / DPI / Handwritten / Background color etc but its easy to enable them by changing to the other ImageParameter constructor here.   Just wanted to check first if thats ok?  Not sure if there was a special reason to use the current constructor here.

src/net/sourceforge/plantuml/ugraphic/ImageBuilder.java
```diff
@@ -281,13 +281,15 @@ private Image getAviImage(AffineTransformation affineTransform) throws IOExcepti
 	private UGraphic2 createUGraphic(FileFormatOption option, long seed, final Dimension2D dim, Animation animationArg,
 			double dx, double dy) {
 		final ColorMapper colorMapper = param.getColorMapper();
+		final double scaleFactor = (param.getScale() == null ? 1 : param.getScale().getScale(dim.getWidth(), dim.getHeight()))
+				* param.getDpi() / 96.0;
```
matthew16550, line 285: Is this calculation correct?

src/net/sourceforge/plantuml/gitlog/GitDiagram.java
arnaudroques: Yes, do whatever you want with this :-)
This is a secret feature, no one is using is yet.
[For example, you can have](http://www.plantuml.com/plantuml/umla/HOxVJl9044VlUOfVyjq0-PIMyaUyCA1YY97eAupTsNPYcJMp2qh2mrkrnCl9d7DcbZ6HfKfIRmZA_6oQNsNe1vDAT5IwE98OzppFiW6UwCZ6xXft3PbuOOSoEBx4ltRTwkE25dvIJF_01_d4XmMjK0iRMLaVKPEwHhJgj1E6M6J5t4sB2VgRzUeU5pVOKKni_z7zyJiESBkNfis-M7XdQuw8KYcsT605MH9FPOhVtV6uTne1lJQT-p8lvwzdTHlscl38-YOQKOSTT_mG4yypvtC-eoldi7dUhi5wm84CAM3tSwiCcaYKhNENhAwIz0K0).
scale/DPI/Handwritten/background won't work because there is no instruction to change it for those diagrams.
But nevertheless, you can do this refactoring.

src/net/sourceforge/plantuml/ugraphic/ImageBuilder.java
arnaudroques: Well, it _sounds_ correct. Now it's difficult to be 100% affirmative without testing.
However, I think you can go ahead for it. We will fix it if needed :-)

arnaudroques:
> There are more anomalies that are not fixable here:
> 
> * JCCKIT does not support Scale or DPI - perhaps it never will?

No, it will never. Don't spend time on this.

> * Latex & Math do not support DPI - they manage Scale & DPI in their own way so a separate PR would make sense.
> 
Agreed !

BTW you are really doing a hard work: thanks for cleaning up all our mistakes !

matthew16550:
Think this is ready now.

arnaudroques:
So merge is done :-)
I've started some test and there seems to be a minor impact.
If you take this simple example:

```
@startuml
state foo
@enduml
```
It looks like left margin is gone with the refactoring.
Do you confirm ?

matthew16550:
Oops yes confirmed some default margin problems, made #494.

PR #495, opened 2021-03-19
Fix default margin regressions

### LINKED ISSUE #494: Regression for default margins
#493 has changed the default margins for Activity & State diagrams:

comparison

And #479 for Sequence Diagrams:

comparison

All other diagrams seem ok.

matthew16550:
Fixes #494

The only differences I see now are with degenerated Class & Description:

comparison

They can probably be adjusted in `EntityImageDegenerated` somehow if you think it's worth doing?

### DIFF
--- a/src/net/sourceforge/plantuml/activitydiagram/ActivityDiagram.java
+++ b/src/net/sourceforge/plantuml/activitydiagram/ActivityDiagram.java
@@ -52,6 +52,7 @@ import net.sourceforge.plantuml.cucadiagram.Ident;
 import net.sourceforge.plantuml.cucadiagram.LeafType;
 import net.sourceforge.plantuml.cucadiagram.NamespaceStrategy;
 import net.sourceforge.plantuml.graphic.USymbol;
+import net.sourceforge.plantuml.style.ClockwiseTopRightBottomLeft;
 import net.sourceforge.plantuml.utils.UniqueSequence;
 
 public class ActivityDiagram extends CucaDiagram {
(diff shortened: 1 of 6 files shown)


arnaudroques:
Thanks, that's really better.

However, there is still a regression on the following example:

```
@startuml
DataAccess - [First Component] 
[First Component] ..> HTTP : use
@enduml
```

There is also a slight change for:
```
@startuml
[-->A : hello
A-->] : hello
@enduml
```
Note that this diff is not a real issue.

>
> They can probably be adjusted in `EntityImageDegenerated` somehow if you think it's worth doing?

No, that's fine. We can keep this one as is.

matthew16550:
Oh.

Calling `SvekResult.calculateDimension()` in `CucaDiagramFileMakerSvek.createFileInternal()` is what ensures the left bit of text is not clipped 🤣 .  Guess we just go back to that instead of most of the changes here?

That sequence diagram doesn't show a difference when I render here.

arnaudroques:
> 
> Calling `SvekResult.calculateDimension()` in `CucaDiagramFileMakerSvek.createFileInternal()` is what ensures the left bit of text is not clipped 🤣 . Guess we just go back to that instead of most of the changes here?
> 
Sorry, I'm not sure to understand your point. :-)

The truth is that, even with this slight regression, the code is far better now.
BTW, I think that this regression is the sign that this clip management was more a hack than a proper solution.

Now fixing this may be more difficult (you will have to dive deeper in the code) , so if you want, I can fix it on my side. But if you want to give a try, I let you go.
**I just don't want that the two of us work on the same time on it.**

Tell me what you prefer. Thanks !

PR #499, opened 2021-03-20
Fix top & left margin problems with CucaDiagrams

### LINKED ISSUE #497: CucaDiagramFileMakerSmetana margin strangeness
The default margins when using `CucaDiagramFileMakerSmetana` have a similar problem to #494 but it's not a recent regression, both `v1.2021.2` and latest are identical for this.

comparison-latest

And when we specify a margin there is extra space added at the top.  e.g. these rows use `margin 0`, `margin 20` & `margin 40`:

margin-comparison-latest

Think the top space issue can be fixed by using `LimitFinder` directly in`CucaDiagramFileMakerSmetana.createFileLocked()` instead of via `ImageBuilder.getFinalDimension()`.

I'm cautious about backwards compatibility here, can the current behaviour be changed?

matthew16550:
Fixes #497

### DIFF
--- a/src/net/sourceforge/plantuml/sdot/CucaDiagramFileMakerSmetana.java
+++ b/src/net/sourceforge/plantuml/sdot/CucaDiagramFileMakerSmetana.java
@@ -102,6 +102,7 @@ import net.sourceforge.plantuml.svek.SvekNode;
 import net.sourceforge.plantuml.svek.TextBlockBackcolored;
 import net.sourceforge.plantuml.ugraphic.ImageBuilder;
 import net.sourceforge.plantuml.ugraphic.ImageParameter;
+import net.sourceforge.plantuml.ugraphic.MinMax;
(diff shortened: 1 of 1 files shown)


arnaudroques:
It's not a big deal, but there are some issues in:

```
@startuml
!pragma graphviz_dot smetana
package "entities" {

A1 -> A2
A3 -> A2 : b2
A3 -> A2
}
@enduml
```
and

```
@startuml
!pragma graphviz_dot smetana
class class1
package pack1 {
  class A1
  class B1
  A1 -- B1
}
@enduml
```

matthew16550:
Hmmm. Think I have the Y adjustments wrong, possibly should be `new YMirror(minMax.getMaxY())` and something is wrong in the `UTranslate` value.  Giving me a headache so I'll need to draw it out with a pencil tomorrow.

PR #510, opened 2021-03-28
Undelete ImageDataComplex constructor that was accidentally removed from net.sourceforge.plantuml.api

matthew16550:
The change was made [here](https://github.com/plantuml/plantuml/pull/507/files#diff-7ce2973dc7559ce7768c5149c7a097555843167fa1a3f30686de499fa3d90fd9)

### DIFF
--- a/src/net/sourceforge/plantuml/api/ImageDataComplex.java
+++ b/src/net/sourceforge/plantuml/api/ImageDataComplex.java
@@ -44,6 +44,13 @@ public class ImageDataComplex extends ImageDataAbstract {
 	private final CMapData cmap;
 	private final String warningOrError;
 
+	@SuppressWarnings("unused")  // available publicly so retained for backwards compatibility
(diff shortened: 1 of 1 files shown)


matthew16550:
This prompts an idea for designing the future public API because the change we made was an implementation detail that probably no-one else should care about.

Perhaps the only part needed in a public API would be the `ImageData` interface?

arnaudroques:
> This prompts an idea for designing the future public API because the change we made was an implementation detail that probably no-one else should care about.
> 
> Perhaps the only part needed in a public API would be the `ImageData` interface?
>

Sure! That's what I mean when I say that the actual public API is badly designed.

My idea is to create a new `com.plantuml.api.wip.v0` package where we could have proposal and test with other users.
Then put `@Deprecated` on all classes and methods of actual `net.sourceforge.plantuml.api`. 

BTW I am not sure that the actual API is widely used. Most plugins run by launching PlantUML as a process. Only few Java PlantUML plugins link with the core library code.

Issue #501, opened 2021-03-21
Future ImageParameter refactorings


matthew16550:
I'd like to talk over my next ideas for simplifying `ImageParameter`.

1. Add these methods in `ImageBuilder` then all the diagrams using `new ImageParameter(TitledDiagram, FileFormatOption)` can use one of them instead.  Longer term they might merge into a single method & adding the annotations is decided automatically somehow.

```java
public static ImageData writeStyledImageWithAnnotations(TitledDiagram diagram, TextBlock textBlock, OutputStream os, int index, FileFormatOption fileFormatOption, long seed)
		throws IOException {
	final ISkinParam skinParam = diagram.getSkinParam();
	final StringBounder stringBounder = fileFormatOption.getDefaultStringBounder(skinParam);
	final AnnotatedWorker annotatedWorker = new AnnotatedWorker(diagram, skinParam, stringBounder);
	return writeStyledImage(diagram, annotatedWorker.addAdd(textBlock), os, index, fileFormatOption, seed);
}

public static ImageData writeStyledImage(TitledDiagram diagram, UDrawable drawable, OutputStream os, int index, FileFormatOption fileFormatOption, long seed)
		throws IOException {
	final ImageParameter imageParameter = new ImageParameter(diagram, fileFormatOption);
	final ImageBuilder imageBuilder = ImageBuilder.build(imageParameter);
	imageBuilder.setUDrawable(drawable);
	return imageBuilder.writeImageTOBEMOVED(seed, os);
}
```

2. Simplify or remove the `index` param.  I haven't seen where the value is actually used?

3. Use `diagram.seed()` in `writeStyledImage()` and remove the `seed` param.  We already pass in the value from `diagram.seed()` in most places except for `GitDiagram`, `JsonDiagram` & `NwDiagram` which use `0` but I assume are safe to use `diagram.seed()`?
Other diagrams that do not use the above new methods are passing in a mixture of `diagram.seed()`, `42` and `0`. I think they can be simplified too but would like to leave them to a later PR when the earlier seed stuff has been removed so it's easier to see what is happening.

4. Looking at the other big `ImageParameter` constructor, it's used by `Help`, `PSystemColors` etc where the user cannot change the skin and the param values are mostly the same for all of them.  So we could extract a 3rd static method in `ImageBuilder` something like `writeImageWithDefaultStyle(UDrawable drawable, OutputStream os, FileFormatOption fileFormatOption)` plus maybe a few params for values that are not constant.

[...]

PR #511, opened 2021-03-28
Move preserveAspectRatio & svgLinkTarget calculation to ImageBuilder class

matthew16550:
As discussed in #501.

### DIFF
--- a/src/net/sourceforge/plantuml/FileFormatOption.java
+++ b/src/net/sourceforge/plantuml/FileFormatOption.java
@@ -65,11 +65,11 @@ public final class FileFormatOption implements Serializable {
 	}
 
 	public FileFormatOption(FileFormat fileFormat) {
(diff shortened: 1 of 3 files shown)


arnaudroques:
It's not a big deal, but I think there is a slight regression on `preserveAspectRatio`.

On SVG, older version outputs `preserveAspectRatio="none"` whereas refactored one outputs now `preserveAspectRatio=""`.

Can you confirm this change ?
Thanks !

matthew16550:
I still get `preserveAspectRatio="none"`.  What is your test setup?

arnaudroques:
> I still get `preserveAspectRatio="none"`. What is your test setup?

So sorry, I should have give an example. I know it's very frustrating when someone says "it does not work for me" without giving a single clue :-)

```
@startuml
listopeniconic
@enduml
```

or

```
@startcreole
this is an example **which is bold**
@endcreole
```

I think other diagrams are working fine.

matthew16550:
Yes I see it now.  Working on a fix.

PR #520, opened 2021-04-04
Simplify default & boolean values in SkinParam

matthew16550:
Also changed `sameClassWidth` param to case insensitive, I'm assuming it was not deliberately case sensitive?

And renamed some existing SkinParam tests for consistency.


arnaudroques:
> Also changed `sameClassWidth` param to case insensitive, I'm assuming it was not deliberately case sensitive?

No, indeed, it was not :-)

PR #551, opened 2021-05-08
Add -filedir option.

matthew16550:
Currently `-pipe` and Picoweb `/render` are allowing `!include` but resolving relative to the process CWD.

"Normal" CLI / GUI rendering is different, includes are resolved relative to the dir containing the PlantUML file.

This PR adds a `-filedir` option to specify where `-pipe` and `/render` should resolve from.

It will be helpful for the Nikola [plantuml_markdown](https://plugins.getnikola.com/v8/plantuml_markdown/) plugin so that includes relative to the markdown file work normally.

Allowing the option with `-pipe` might seem unnecessary because the process could be just run in the relevant directory but I've found it greatly simplifies things when Nikola runs PlantUML inside Docker.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,6 +78,12 @@
             <scope>provided</scope>
         </dependency>
 		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-lang3</artifactId>
+			<version>3.12.0</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
 			<groupId>org.assertj</groupId>
 			<artifactId>assertj-core</artifactId>
 			<version>3.19.0</version>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -77,6 +77,12 @@
             <version>1.10.9</version>
             <scope>provided</scope>
         </dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-lang3</artifactId>
```
matthew16550, line 82: @arnaudroques not sure what you think about adding this as a test dependency?  It's not providing much for this PR.

arnaudroques: Well, we try to limit dependencies. So I think it's better if you can avoid it here.
Sorry about that, but thanks for your contribution !

matthew16550: removed it

PR #560, opened 2021-05-23
Use GitHub workflow to attach JAR to GitHub Releases

### LINKED ISSUE #557: Make jar available from Github?
Could you please make the binary / jar also available as a download from the Github repository? Currently, you can only fetch it from Sourceforce. As Sourceforce was under a lot of heat for [adding Adware to downloads](https://www.reddit.com/r/sysadmin/comments/7c9q1o/is_sourceforge_safe_adware_free/), I would very much prefer to avoid that detour.

It should just be a few clicks to add the .jar to the releases, as [described here](https://docs.github.com/en/github/administering-a-repository/managing-releases-in-a-repository), or [automatically as a Github Action](https://github.com/marketplace/actions/upload-files-to-a-github-release).

Thanks in advance!

matthew16550:
Resolves #557

Here is an [example release](https://github.com/matthew16550/plantuml/releases/tag/example) and associated [workflow run](https://github.com/matthew16550/plantuml/runs/2648982270?check_suite_focus=true).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,12 @@
 name: tests
-on: push
+
+on:
+  pull_request:
+    types: [ opened, synchronize, reopened ]
+  push:
+    branches:
+      - master
+
 jobs:
   run_tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -1,5 +1,12 @@
 name: tests
-on: push
+
```
matthew16550, line 2: This prevents the test workflow running when a release is made but enables it for PRs and everything on the main branch.

.github/workflows/ci.yml
```diff
@@ -12,12 +19,19 @@ jobs:
...
           java-version: 8
 
-      - name: Cache Maven packages
+      - name: Cache local Maven repository
         uses: actions/cache@v2
         with:
-          path: ~/.m2
-          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
-          restore-keys: ${{ runner.os }}-m2
+          path: ~/.m2/repository
```
matthew16550, line 25: Little functional change here, just matching the example in the [docs](https://github.com/actions/cache/blob/main/examples.md#java---maven)

pom.xml
```diff
@@ -227,6 +227,11 @@
                     <mavenExecutorId>forked-path</mavenExecutorId>
                 </configuration>
             </plugin>
+			<plugin>
```
matthew16550, line 230: Pinning dependency versions is always a good idea 😄 
And including it here means that `dependency:go-offline` will download it.

arnaudroques:
Many thanks for your PR !

Issue #599, opened 2021-07-29
plantuml -pipe fails if there's any content after @enduml 


jmones:
`plantuml -pipe` fails if there's any content at all after @enduml.

The root cause seems to be that `Pipe::managePipe()` will attempt to process a 2nd diagram which would fail (because input is insufficient).

## Steps to reproduce
```
$ echo -n "@startuml\na->b\n@enduml\n\n" | java -Djava.awx.headless=true -jar plantuml-1.2021.9.jar -p > /dev/null
ERROR
1
Empty description
Some diagram description contains errors
$ echo $?
200
```
## Expected
```
$ echo -n "@startuml\na->b\n@enduml\n\n" | java -Djava.awx.headless=true -jar plantuml-expected.jar -p > /dev/null
$ echo $?
0
```

## Impact in Emacs plantuml-mode

This is impacting in emacs [plantuml-mode](https://github.com/skuro/plantuml-mode), which will typically send a whole buffer to plantuml via pipe and *fails if the return code of the application is not 0*.

## What works well

Just one `\n` after `@enduml`:
```
$ echo -n "@startuml\na->b\n@enduml\n" | java -Djava.awx.headless=true -jar plantuml-1.2021.9.jar -p > /dev/null
$ echo $?
0
```

No `\n` after `@enduml`:
```
$ echo -n "@startuml\na->b\n@enduml" | java -Djava.awx.headless=true -jar plantuml-1.2021.9.jar -p > /dev/null
$ echo $?
0
```

No tags:
```
$ echo -n "a->b" | java -Djava.awx.headless=true -jar plantuml-1.2021.9.jar -p > /dev/null
$ echo $?
0
```

PR #600, opened 2021-07-29
Bugfix to make plantuml pipe play nice when @startuml & @enduml marks are present

jmones:
Fixes https://github.com/plantuml/plantuml/issues/599

## Behaviours changed:
- When `@start*` & `@end*` are present, any other content outside those will be ignored
- When `@start[whatever]` is found, a specific `@end[whatever]` is sought. Any other `@start...` or `@end...` are ignored.
- When a `@start[whatever]` is found and EOF is found without matching `@end[whatever]`, the latter is added automatically (not that it's not only `@enduml`)
- Changed adhoc parsing of CR LF to more standard use with `BufferedReader` (note that the combination LF CR may behave differently, but that's really not used).

## Other changes:
- `PipeTests` have been coded to provide some scaffolding 
- Added `.java-version` to `.gitignore` for devs using `jenv`
- Changed access to of some methods in `Pipe` from `private` to `package-private` to allow them be tested

## _Suspicious_ behaviours preserved:
- `@@@format ....` is still affects current diagram and all subsequent (unless overriden) and allows garbage after `@@@format ` (as long as there's a `png` or `svg`)
- pipemap ignoring any errors
- Error of first diagram will be masked by success of the subsequent

## Other changes I didn't do...
- Avoided adding libraries, but would have loved lombok to make `TestCase` and `InputExpected` more concise
- Didn't change integration of `Pipe` with other classes (for example, validating `charset` in `Run.java` and passing a `Charset` instead of a `String`)

## Things that didn't work previously, now working
```
$ echo -n "@startuml\na->b\n@enduml\n\n" | java -Djava.awx.headless=true -jar plantuml-fix-pipe-issue.jar -p > /dev/null
$ echo $?
0

$ echo -n "garbage\n@startuml\na->b\n@enduml\ngarbage\n@startuml\na->b\n@enduml\ngarbage\n" | java -Djava.awx.headless=true -jar plantuml-fix-pipe-issue.jar -p > /dev/null
$ echo $?
0
```


arnaudroques:
Many thanks for your Tests, it really helps.
We've just made a slight modification because some tests were not running on Windows: we add to remove all `\r` characters.
Not sure that this is the best way to fix it, but at least it works on Windows this way.

Thanks again,

jmones:
@arnaudroques Wouldn't dare to change the file so much without some test scaffolding... thanks for the feedback!

PR #663, opened 2021-09-10
Refactor some code to use AutoCloseable / try-with-resources

matthew16550:


arnaudroques:
I have to rollback changes made on `AFileZipEntry.java`.

Otherwise, the following example does not work any more and raises some `java.io.IOException: Stream closed`.

```
@startuml
!import demo.zip
alice->bob : before include
!include MyService1/myMethod1.iuml
alice->bob : after include
@enduml
```

With `demo.zip` containing some `MyService1/myMethod1.iuml` file.

I think it's because we should not close the `InputStream` at https://github.com/plantuml/plantuml/blob/master/src/net/sourceforge/plantuml/AFileZipEntry.java#L71

I hope (!) the `InputStream` is closed further in the code and that there is no leak here. This probably means that some refactoring is also needed here :-)

Anyway, thanks for your contribution!

matthew16550:
Oops yes my bad 🤦

arnaudroques:
> Oops yes my bad 🤦

Don't worry, this is not important.

This probably means that we have yet another a [code smell](https://en.wikipedia.org/wiki/Code_smell) here.

We can keep it this way right now, there is enough work in progress :-)

PR #677, opened 2021-09-20
Rename UGraphic2.writeImageTOBEMOVED() -> UGraphic.writeToStream() & remove UGraphic2 interface

matthew16550:
Here is an idea for tidying `writeImageTOBEMOVED()`.  Not sure what the earlier thinking was for this method so feel free to suggest something else or reject if this isn't going in a helpful direction.


src/net/sourceforge/plantuml/ugraphic/FontChecker.java
```diff
@@ -175,8 +175,8 @@ public BufferedImage getBufferedImage(final char c) throws IOException {
 			public void drawU(UGraphic ug) {
 				ug = ug.apply(HColorUtils.BLACK);
 				ug.draw(new URectangle(dim - 1, dim - 1));
-				if (ug instanceof UGraphic2) {
-					ug = (UGraphic2) ug.apply(new UTranslate(dim / 3, 2 * dim / 3));
+				if (!(ug instanceof LimitFinder)) {
```
matthew16550, line 178: Maybe this is strange? I just changed the code to behave the same at runtime.

arnaudroques: Your refactoring makes really sense. It seems that this test is not useful anymore.
BTW, the `getBufferedImage` method can go private now: it's not used outside of `FontChecker` anymore.
Maybe it's related ?

matthew16550: I don't know how `FontChecker` is being used but it looks like all of its methods can be private, the only thing that needs to be public is `main` ?

arnaudroques: Well, you're right. I think that `FontChecker` has been used in some past to make to test on fonts.
It  should not have been committed here. So the best option is simply to delete it :-)

PR #689, opened 2021-10-02
StringBounder injection & some minor related tidyings

matthew16550:
Allows passing a single `StringBounder` instance into lots of places.  It helps the font sprite work (#681) and is probably worthwhile on its own.

I think this is heading this part of the code towards a simpler state but it will need a few more PRs to get there.  Not sure what those would look like just yet!

I've always found this area confusing so there is a lot of small refactoring commits that are hopefully each simple enough to be obviously correct.

### DIFF
--- a/src/net/sourceforge/plantuml/EmptyImageBuilder.java
+++ b/src/net/sourceforge/plantuml/EmptyImageBuilder.java
@@ -46,6 +46,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import net.sourceforge.plantuml.cucadiagram.dot.GraphvizUtils;
+import net.sourceforge.plantuml.graphic.StringBounder;
 import net.sourceforge.plantuml.ugraphic.UAntiAliasing;
 import net.sourceforge.plantuml.ugraphic.color.ColorMapperIdentity;
 import net.sourceforge.plantuml.ugraphic.color.HColor;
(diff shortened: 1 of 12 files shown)


arnaudroques:
Well, before merging this pull request, we probably need a discussion about `StringBounder`.

matthew16550:
Thanks @arnaudroques .  Suspect this caused a headache while reviewing!

arnaudroques:
> Thanks @arnaudroques . Suspect this caused a headache while reviewing!

A philosophical debate :-)

The big question is about the link between `UGraphic` and `StringBounder`.
As you know, the `StringBounder` is a tool that computes the actual size (in pixels) of a String (depending on a Font).

Initially, we though that any `UGraphic` must define its own `StringBounder`. That was the real meaning of the `getStringBounder` method of `UGraphic`.
So putting a `StringBounder` in the constructor of any `UGraphic` should not make sense, since it's the `UGraphic` should create the `StringBounder`, and not the other way around.

Now, it's not that clear :-)

Conceptually speaking, a `UGraphic` does not care about `StringBounder`. `UGraphic` implementations should not use `getStringBounder` method, because they should not care about String size. On the other size, code using `UGraphic` usually needs to know String's size, so usually `UGraphic` clients do need to call `getStringBounder` method.

I hope I did not confuse you here, I guess there will be several refactorings on this part :-)

Thanks again for your contributions.

matthew16550:
A difficult topic indeed.

I just noticed, IntelliJ tells me the `StringBounder` interface is used 1690 times in the source (not including test code / import lines). But its methods are not often called:
* `calculateDimension()` 23 times
* `getDescent()` 2 times

Don't really know what that tells us but seems like a clue there is a simpler abstraction hiding somewhere.

PR #692, opened 2021-10-04
Delete UScale and related code as it is now unused

### LINKED ISSUE #691: Do we need UScale ?
I don't see `UScale` being created anywhere.

Is it something old that we can delete or just something that isn't finished yet?

matthew16550:
Closes #691 

Not sure about deleting `getScaled()` in `UEllipse` / `ULine` / `URectangle`.  I will change the PR if you want to keep them?


arnaudroques:
There have been so many changes around scale management... that's good to clean unused code to make it more clear.

PR #693, opened 2021-10-04
Remove some redundant calls to StringBounder.calculateDimension()

matthew16550:


arnaudroques:
Maybe it's because I'm old fashion, but I am not a big fan of `import static any.method.static` :-)

For example, I think that `TextBlockUtils.createTextLayout(font, shape.getText())` is less ambigous than `createTextLayout(font, shape.getText())` because the last one may reference a local method. Note that it's just a personal taste.

matthew16550:
For me it's usually about less source code.  The less we have, the easier it might be to understand what we are doing.  Provided we don't go into weird code golf land.

And I think import lines don't weigh very heavily vs the main body.

Context is important, if the code is lots of math then `min()` / `max()` is probably obvious.  But `Foo.create()` is usually more obvious than `create()`.

PR #707, opened 2021-10-15
For discussion - UDriverContext

matthew16550:
This PR won't compile, I only changed a few classes to show the idea.

I'm thinking we could make the driver classes stateless and pass everything into the `draw()` method.

`UParam` is already a context, suggest we rename it to `UDriverContext` and add a few methods.

It will help to simplify the philosophical debate from #689 about where a `StringBounder` should be created because drivers will no longer need to have an opinion.

@arnaudroques  ?


evantill:
@arnaudroques @matthew16550 can we close this PR ?

Issue #514, opened 2021-04-02
Java 1.8 source ?


matthew16550:
`build.xml` & `pom.xml` put our source level at 1.7 which is more or less at [end-of-life](https://www.oracle.com/java/technologies/java-se-support-roadmap.html) though some enterprise users might still be running it.

I'm writing a few tests that would be much cleaner with 1.8 Lambdas, do we have a plan for when to move to 1.8 ?

arnaudroques:
> `build.xml` & `pom.xml` put our source level at 1.7 which is more or less at [end-of-life](https://www.oracle.com/java/technologies/java-se-support-roadmap.html) though some enterprise users might still be running it.
> 
> I'm writing a few tests that would be much cleaner with 1.8 Lambdas, do we have a plan for when to move to 1.8 ?

Concerning Java version, we try to be **very** conservative. As you may know, some enterprises might still run older versions of Java (although they won't publicly say it :-)  
So we try to support old Java old versions : we think it's important for some of our users. Until recently we were even supporting 1.6 !
Usually, we follow Oracle support roadmap. Since Java 7 extended support runs until July 2022, let's try to wait until July 2022 to fully switch to 1.8

I know this is annoying and we also would love to use new Java features...

However, maybe we can have an intermediate solution : since you are thinking about using Lambda only in tests, do you think it's doable in `pom.xml` to compile main source code in `1.7` and test source code in `1.8` ?

matthew16550:
Yes I've done something similar once before and it seemed to work ok though I was the only person coding there so could never get feed back if it was causing problems for others people.

I'll try in a future PR, it's only a couple of tests so easy to revert if we don't like it.

PR #740, opened 2021-11-02
Change test sources to Java 1.8

matthew16550:
As discussed in #514 

My #681 work is I think a concrete reason for 1.8 tests.

Not sure how to proceed with this change @arnaudroques, would you like to merge or maybe just try locally to see if it breaks anything for you? 

I have been using it for several weeks on #681.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,6 +57,13 @@
         <version>7</version>
     </parent>
 
+	<properties>
+		<maven.compiler.source>1.7</maven.compiler.source>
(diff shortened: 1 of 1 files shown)


arnaudroques:
@matthew16550 Let's test this directly on the main branch and see the result :-)

BTW, let's really stick to java 1.7 right now. [Some users are complaining that PlantUML does not compile on Java 1.6 anymore](https://forum.plantuml.net/14875/why-from-april-this-year-plantuml-for-java-compatible-anymore).

PR #744, opened 2021-11-05
Change Sequence Diagram so group backgrounds are transparent when the diagram background is transparent

matthew16550:
Currently, if the diagram background is transparent then the "group" background (used in `alt/else`, `group`, `loop`, `opt` etc) is set to white. This PR changes the main rectangle of the group to be transparent in this situation.

This PR does not change the top corner that has the text, it still defaults to white but can be configured to something else and it is a wider reaching area of code so needs more thought.

@arnaudroques this is the last place where `replaceTransparentByWhite` is `true`, everywhere else it is `false` so I think we can remove the parameter?  Suspect we will never need it in future but if we do then we could make a new function with a specific name.

### DIFF
--- a/src/net/sourceforge/plantuml/skin/rose/Rose.java
+++ b/src/net/sourceforge/plantuml/skin/rose/Rose.java
@@ -286,7 +286,7 @@ public class Rose {
 				smallFont = smallFont.changeColor(smallColor);
 			}
 			return new ComponentRoseGroupingHeader(styles == null ? null : styles[0], styles == null ? null : styles[1],
-					param.getBackgroundColor(true), getSymbolContext(stereotype, param, ColorParam.sequenceGroupBorder),
+					param.getBackgroundColor(false), getSymbolContext(stereotype, param, ColorParam.sequenceGroupBorder),
 					bigFont, smallFont, stringsToDisplay, param, roundCorner);
 		}
 		if (type == ComponentType.GROUPING_ELSE) {


arnaudroques:
> @arnaudroques this is the last place where `replaceTransparentByWhite` is `true`, everywhere else it is `false` so I think we can remove the parameter? Suspect we will never need it in future but if we do then we could make a new function with a specific name.

Yes, you can go ahead and remove completely this parameter. Its original usage was due to the fact that transparent support has been added very lately in PlantUML.

Since you had a look on this `Rose` class, I take this opportunity to give you few information about this class. This is really a legacy one. 10 years ago, there was no style, not even skinparam and our original intention was that people would provide other implementations of `Rose` to change colors and styles (`Rose`was implementing an interface at that time). Finally, it was too complex to ask to people to code some Java class to only change colors and we then introduce skinparam features. Latter, we realize that skinparam was no standard and we introduce CSS (through style), work still in progress.

So here is the roadmap for the next 10 years :-)

1. Finish CSS/style migration (so `UseStyle.useBetaStyle()` will always be `true` and we will be to remove a lot of code)
2. Finish Teoz layout engine (that is use `SequenceDiagramFileMakerTeoz` and remove `SequenceDiagramFileMakerPuma2`)
3. Redesign [ASCII art support](https://plantuml.com/ascii-art) to handle other diagrams than sequence diagram. This would probably means that `TextSkin` class will be removed.
4. Remove `Rose`, `ComponentType`, ... to simplify the code.

Don't hold your breath and thanks again for your useful contributions!

matthew16550:
@arnaudroques  Thanks for the info.

PR #755, opened 2021-11-10
Introduction of an internal function %loadJSON to load JSON data

arittner:
This new internal function %loadJSON allows loading a JSON data structure from a local file or external HTTP(s) URL. 
The implementation uses SFile and SURL to manage the security context and configured opt-in folders/domains.

loadJSON supports up to three parameters. A mandatory data-source path (relative file, absolute file-path, http, and https URL), a second default parameter to define a fallback JSON (if the data source has no content), and a third default parameter to choose a charset encoding (default is UTF-8).

A test is included (but only for local files). Short introduction in the JavaDoc comment of LoadJson.java. Also, the test file gives an impression of how to use it.

Example to access a Spring cloud native service (health actuator endpoint):

```
@startuml

' This example works only with the latest plantuml snapshot an implemented
' %loadJSON function

' Default JSON, if we don't get a content from the data source
!$DEF={"status": "not connected"}

' Load from Spring boot service with actuator 'health' endpoint
!$TEST=%loadJSON("http://localhost:7778/management/health", $DEF)

' Load from local file (relative located in plantuml.include.path list)
!$DEMO=%loadJSON("demo.json")

title Load JSON example with $DEMO.status

' Simple usage
[Service] -> ($TEST.status): is

' Embedded rendering of JSON structure
 rectangle Inner #tan as "
    {{json
    $TEST
    }}
    Health of management service at %date()
    "

@enduml

```

JSON load example

Limitations: 
Authentication is currently not supported. Maybe this should be applied to SURL but it's out of scope for this pull request.

### DIFF
--- a/src/net/sourceforge/plantuml/tim/TContext.java
+++ b/src/net/sourceforge/plantuml/tim/TContext.java
@@ -106,6 +106,7 @@ import net.sourceforge.plantuml.tim.stdlib.InvokeProcedure;
 import net.sourceforge.plantuml.tim.stdlib.IsDark;
 import net.sourceforge.plantuml.tim.stdlib.IsLight;
 import net.sourceforge.plantuml.tim.stdlib.Lighten;
+import net.sourceforge.plantuml.tim.stdlib.LoadJson;
 import net.sourceforge.plantuml.tim.stdlib.LogicalNot;
 import net.sourceforge.plantuml.tim.stdlib.Lower;
 import net.sourceforge.plantuml.tim.stdlib.Newline;
@@ -177,6 +178,7 @@ public class TContext {
 		functionsSet.addFunction(new Hex2dec());
 		functionsSet.addFunction(new Dec2hex());
 		functionsSet.addFunction(new HslColor());
+		functionsSet.addFunction(new LoadJson());
 		functionsSet.addFunction(new Chr());
 		functionsSet.addFunction(new Size());
 		// %standard_exists_function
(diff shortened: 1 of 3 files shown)


arittner:
> We will just probably reformat it slightly, but nothing significant.

Sorry for the formatting issues. The new setup of my IntelliJ project was not well aligned (some artifacts from my company formatting stuff).

> I'm sure that users will be very creative with this new feature :-)

Of course. I'm an architect and want to create architecture models on-the-fly. In my opinion this is a certain step in CI/CD. No one likes to draw manually diagrams on any new architecture or deployment change. Now I can fetch OpenAPI outputs, monitor actuator endpoints, read Kubernetes and/or Rancher configs and create model snapshots from dev, test or customer environments. I've integrated PlantUml in gradle and as IntelliJ plugin to raise up a project to create an architecture documentation on-the-fly.

The next step will be to add authentication. Let's see how this will work.

Daniel15:
Shouldn't it be called `load_json`? It's the only camelcase function - all the others use underscores.

arnaudroques:
> Shouldn't it be called `load_json`? It's the only camelcase function - all the others use underscores.

Sure, you're right.
[Last snapshot](https://github.com/plantuml/plantuml/releases/tag/snapshot) handles both syntax (`%loadJSON` and `%load_json`).

We will probably remove `%loadJSON` in some future.
Does it make sense?

PR #757, opened 2021-11-10
Create puml-theme-spacelab-white

KingPlus-lang:

PR #758, opened 2021-11-10
Rename puml-theme-spacelab-white to puml-theme-spacelab-white.puml

KingPlus-lang:


The-Lum:
See also remarks on:
- #757

PR #759, opened 2021-11-11
adds theming for gantt diagrams

bschwarz:
adds theming for gantt diagrams


The-Lum:
Hi @bschwarz, @KingPlus-lang 

After analysis, it appears that `spacelab-white` (from #757) has not been modified.

_I can propose a PR for that._ 🚀 

Regards,
Th.

PR #761, opened 2021-11-11
Improve MAVEN build Performance

hj987654321:
Maven allows you to run the compiler as a separate process by setting `<fork>true</fork>`. This feature can lead to much less garbage collection and make Maven build faster. This project has more than 1000 source files. We can consider enabling this feature.

Maven can recompile only the classes that were affected by a change. This feature is the default. We can activate it by setting `<useIncrementalCompilation>true</useIncrementalCompilation>`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -164,6 +164,10 @@
                 <version>3.8.1</version>
                 <configuration>
                     <debug>true</debug>
+                	<fork>true</fork>
+                	<useIncrementalCompilation>true</useIncrementalCompilation>
+
+
                 </configuration>
             </plugin>
             <plugin>


pzygielo:
Hmmm https://github.com/apache/maven-surefire/pull/394#issuecomment-966250959

mpkorstanje:
@arnaudroques I would recommend reverting. This user has made the same/similar PR to several other projects.

arnaudroques:
@mpkorstanje Ok. I am not sure that this was really useful anyway so I've just reverted it. Thanks for the info.

i-Taozi:
Sorry to disturb u. But I'm are not  spam. We are doing a research work about CI . I want to know whether some configuration is useful for developers and shorten the CI build time.

PR #764, opened 2021-11-12
Add a new artefact containing single jar

The-Lum:
Add another artefact, with only the plantuml.jar 
See:
https://github.com/plantuml/plantuml/discussions/653#discussioncomment-1631002


arnaudroques:
@matthew16550 This is indeed in //
I like the idea of trying all solutions before choosing the right one :-)

PR #774, opened 2021-11-15
mod: add `datetime` of the snapshot

The-Lum:
add `datetime` of the snapshot on the notes


The-Lum:
FYI @matthew16550:
Thanks a lot for the snapshot.. 🎉 
I just create this PR in order to add `date-time` of the snapshot.

arnaudroques:
@The-Lum The `datetime` format is not very human readable. 
Is it possible to change it. Something like:  _Snapshot taken the 2021-11-15 at 21:53 (GMT)_ ?

PR #781, opened 2021-11-18
Maven build profile for building with PDF dependencies to a fat jar

sfuhrm:
After several attempts to collect the required jars documented [here](https://plantuml.com/en/pdf), I have a PR as a suggestion for automating an alternative build with a PDF enabled jar.

This is a change in the pom.xml to output a JAR with the PDF dependencies in case the build profile is "pdf":

```
$ mvn -Ppdf clean package
...
[INFO] --- maven-assembly-plugin:3.1.1:single (make-assembly) @ plantuml ---
[INFO] Building jar: .../plantuml-1.2021.15-SNAPSHOT-jar-with-dependencies.jar
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  41.445 s
[INFO] Finished at: 2021-11-18T14:25:37+01:00
[INFO] ------------------------------------------------------------------------
$ java -jar target/plantuml-1.2021.15-SNAPSHOT-jar-with-dependencies.jar -Tpdf /tmp/ng_deployment_facade.puml 
$ ls -al /tmp/ng_deployment.pdf 
-rw-r--r-- 1 stephan stephan 809721 18. Nov 14:27 /tmp/ng_deployment.pdf
$
```

I hope you find it useful.
It is possible to add this also to the github actions build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,6 +63,7 @@
 		<maven.compiler.target>1.7</maven.compiler.target>
 		<maven.compiler.testSource>1.8</maven.compiler.testSource>
 		<maven.compiler.testTarget>1.8</maven.compiler.testTarget>
+		<main.class>net.sourceforge.plantuml.Run</main.class>
 	</properties>
 
 	<dependencyManagement>
@@ -202,15 +203,22 @@
                 </executions>
             </plugin>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
                 <version>3.2.0</version>
                 <configuration>
                     <archive>
-                       <manifestFile>manifest.txt</manifestFile>
+                        <manifest>
+                            <addClasspath>true</addClasspath>
+                            <useUniqueVersions>false</useUniqueVersions>
+                            <mainClass>${main.class}</mainClass>
+                            <addBuildEnvironmentEntries>true</addBuildEnvironmentEntries>
+                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+                        </manifest>
                     </archive>
                 </configuration>
             </plugin>
-			<plugin>
+            <plugin>
 				<artifactId>maven-surefire-plugin</artifactId>
 				<version>3.0.0-M5</version>
 				<configuration>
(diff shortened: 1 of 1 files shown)


arnaudroques:
Yes, this is interesting!
However, could you resolve the conflicts on `pom.xml` so that we can merge it?

Regarding github actions, I think this should remain optional right now (that is, we won't compile with profile `pdf`).
We'll see if we will change this in some future.

Thanks!

sfuhrm:
Before we're continuing, let's first clarify one question:

In the `manifest.txt` file there are multiple jars in the classpath that can be optionally (?) added/linked in runtime:

```
Class-Path: elk-full.jar avalon-framework-4.2.0.jar batik-all-1.7.jar commons-io-1.3.1.jar commons-logging-1.0.4.jar fop.jar xml-apis-ext-1.3.04.jar xmlgraphics-commons-1.4.jar jlatexmath-minimal-1.0.3.jar jlm_cyrillic.jar jlm_greek.jar vizjs.jar j2v8_win32_x86_64-3.1.6.jar j2v8_linux_x86_64-3.1.6.jar j2v8_macosx_x86_64-3.1.6.jar ditaa0_9.jar
```

I've dropped this temporarily in the pom. 
But this will probably break stuff with some features, right?
Can you give me a hint on what this is all about?

arnaudroques:
> Before we're continuing, let's first clarify one question:

Sure!

> I've dropped this temporarily in the pom.
> But this will probably break stuff with some features, right?

Yes, you should probably restore them

> Can you give me a hint on what this is all about?

I've grouped jar files per features:

**elk-full.jar**
This is need for ELK. See https://plantuml.com/en/elk

**avalon-framework-4.2.0.jar batik-all-1.7.jar commons-io-1.3.1.jar commons-logging-1.0.4.jar fop.jar xml-apis-ext-1.3.04.jar xmlgraphics-commons-1.4.jar jlatexmath-minimal-1.0.3.jar jlm_cyrillic.jar jlm_greek.jar**
Those are needed to generate PDF

**vizjs.jar j2v8_win32_x86_64-3.1.6.jar j2v8_linux_x86_64-3.1.6.jar j2v8_macosx_x86_64-3.1.6.jar** 
This is needed to run Javascript implementation of Graphviz/dot. See https://plantuml.com/en/vizjs

**ditaa0_9.jar**
This is a unfinished try for non GPL version of PlantUML to generate ditaa diagrams. Not sure that is really implemented.

The idea is to put all those jar in the classpath be default so that people can optionally _download_ required jar files and put them at the same place as `plantuml.jar` so that they can use these features.

Hope this helps!

sfuhrm:
Thanks for explanation! 

I've changed the PR to stick with the manifest.txt for the normal build.

The `-Ppdf` profile build will not have this class path, so it won't have these extensions possible. It will have the PDF dependency directly included in the fat jar.

arnaudroques:
I may be wrong, but I think indentation in the new pdf profile are not consistent with the rest of the file.
Could you fix it before we merge ?
Thanks!

sfuhrm:
You're right. I've fixed the indentation and updated the PR.

PR #809, opened 2021-11-30
Improve performance by replacing StringBuffer with StringBuilder

larrydiamond:
Only StringBuffers are being replaced with StringBuilder in this pull request.
No new features in this PR.

StringBuilder is faster than StringBuffer as synchronization does not occur.  
This is safe to do for locally used StringBuffers where there is no need for synchronization, all places changed this PR are locally used.

All tests pass.

### DIFF
--- a/src/ext/plantuml/com/ctreber/acearth/util/Toolkit.java
+++ b/src/ext/plantuml/com/ctreber/acearth/util/Toolkit.java
@@ -96,7 +96,7 @@ public class Toolkit
   {
     boolean lDoCap = false;
     final StringTokenizer lST = new StringTokenizer(pText, ".- ", true);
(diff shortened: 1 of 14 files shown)


larrydiamond:
Build failed on Java 8, I'll take a look and fix it up, sorry about that.

arnaudroques:
Many thanks, tt sounds like a good idea.
Unfortunately, it seems that under Java 8, regex and matcher still need `StringBuffer`.
It's really a pity but we would like to keep compatibility with Java 8.

larrydiamond:
Makes perfect sense to me.   It looks like Java didn't add some methods for StringBuilder until 11.   I think I still have 8 installed somewhere, I'll figure it out.

PR #823, opened 2021-12-09
Add artifact signing to CI workflow & publish signature files in snapshot releases

matthew16550:
As discussed in https://github.com/plantuml/plantuml/discussions/769#discussioncomment-1725978.

To enable signing we will need `ARTIFACT_SIGNING_KEY` & `ARTIFACT_SIGNING_PASSPHRASE` secrets.  It does not matter if they are added after the PR is merged, signing is optional and only happens when `ARTIFACT_SIGNING_KEY` is set.  (This seems simplest so we don't require forked repos set up their own signing keys)

Also renaming the timestamp file to `plantuml-SNAPSHOT.timestamp` as discussed in https://github.com/plantuml/plantuml/discussions/653#discussioncomment-1772902.  (fyi @The-Lum)

### DIFF
--- a/.github/scripts/release.sh
+++ b/.github/scripts/release.sh
@@ -1,13 +1,18 @@
 #!/usr/bin/env bash
 set -ex
 
-mv plantuml.jar         "plantuml-${POM_VERSION}.jar"
-mv plantuml-javadoc.jar "plantuml-${POM_VERSION}-javadoc.jar"
-mv plantuml-sources.jar "plantuml-${POM_VERSION}-sources.jar"
+RELEASE_DIR="target/github_release"
 
-gh release create --target "${GITHUB_SHA}" "${TAG}" \
-  "plantuml-${POM_VERSION}.jar" \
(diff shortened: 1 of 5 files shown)


arnaudroques:
The release job failed, but maybe it's normal, since there is no release at that time :-)

matthew16550:
Bother.  Thought I tested that but obviously not 100%.

It should be ok after adding the GPG secret but it will be broken in forks too so I will find a proper fix.

arnaudroques:
Many thanks @matthew16550 
Next step is on my side now.

I have to add the GPG secret key `ARTIFACT_SIGNING_KEY`, and passphrase `ARTIFACT_SIGNING_PASSPHRASE` as discussed here https://github.com/plantuml/plantuml/discussions/769

The key will have a User ID something like `PlantUML JAR Signing Key <plantuml@gmail.com>`.
The public key will be put somewhere on https://plantuml.com

Here is some [help for myself](https://stackoverflow.com/questions/61096521/how-to-use-gpg-key-in-github-actions) :-)

PR #847, opened 2022-01-05
add basic gradle files to allow conversion to gradle in future, #47

soloturn:
it is just a start to allow adding build files later. hope thats ok. i saw #93, hope its ok to do this in small steps.


arnaudroques:
We have nothing against Gradle but we want to stay under Ant.

Since you can call Ant from Gradle, is this PR really useful ?

I am also unconfortable about adding a binary file (gradle-wrapper.jar), because there is no way to check this file integrity.

soloturn:
added a first draft of build.gradle.kts. the wrapper jar integrity can be checked like this:
https://docs.gradle.org/current/userguide/gradle_wrapper.html#wrapper_checksum_verification

PR #861, opened 2022-01-15
mvn: implementation-version into manifest

soloturn:
the version can be read out of manifest, so no change of version.java would be necessary, so put it in.

@arnaudroques before changing version.java - is there any special logic coupled to the integer version field, which is currently 1200200 ? apart from getting a dotted vesion again?

i saw one usage:
```
  final BitMatrix bit = writer.encode(s, BarcodeFormat.QR_CODE, multiple, hints);
```
but am not clear if it is just information or there is a deeper meaning.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,31 +4,31 @@
 	PlantUML : a free UML diagram generator
 	========================================================================
 	(C) Copyright 2009, Arnaud Roques
-	
+
 	Project Info:  http://plantuml.sourceforge.net
-	 
+
 	This file is part of PlantUML.
-	
+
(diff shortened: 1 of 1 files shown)


soloturn:
added gradle workflow script, triggers switched off, so one can easy fork and enable in case a person prefers gradle releasing.

arnaudroques:
> @arnaudroques before changing version.java - is there any special logic coupled to the integer version field, which is currently 1200200 ? apart from getting a dotted vesion again?

This is not very standard but having the version in a Java file (that is `Version.java`) is quite handy because you have direct access to the version details from java. Is there a real need to change this?

PR #871, opened 2022-01-22
valueOf instead of new

soloturn:
new Integer and similar got a removal warning in java-17, do Integer.valueOf
and similar for other types.

.editorconfig caused some trimming of trailing whitespaces.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,9 +5,13 @@ root = true
 charset = utf-8
 end_of_line = lf
 indent_size = 2
-indent_style = space
+indent_style = tab
 trim_trailing_whitespace = true
 
 [*.md]
 max_line_length = off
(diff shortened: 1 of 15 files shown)


arnaudroques:
> .editorconfig caused some trimming of trailing whitespaces.

That's fine.

However, change from https://github.com/plantuml/plantuml/pull/871/commits/bcceefdea69c43a63dd76c643672dbdf14f44010 is not ok because we do use tabulation right now (instead of 2 chars).

I understand this is a matter of taste here but if we change this convention, we have to do it on the whole project. So could you please revert this one? Thanks!

soloturn:
created a feature request towards intellij, as it would be nice it defaults to a sane "leave the indent style in place":
https://youtrack.jetbrains.com/issue/IDEA-287485

PR #891, opened 2022-02-02
upd: `undone` as `unstarted` (chidren of `task`)

The-Lum:
Make `undone` as `unstarted` (chidren of `task`)

To answer to:
- https://forum.plantuml.net/15299/how-to-set-color-of-the-gantt-unstarted-task?show=15307#c15307

And to make symmetrical with `unstarted`.
See:
https://github.com/plantuml/plantuml/blob/bbeb28002289c1a523fc4cf612959993d2aa5df3/src/net/sourceforge/plantuml/project/draw/AbstractTaskDraw.java#L103-L105

- Is this PR is enough?

Regards.

### DIFF
--- a/src/net/sourceforge/plantuml/project/draw/TaskDrawRegular.java
+++ b/src/net/sourceforge/plantuml/project/draw/TaskDrawRegular.java
@@ -292,7 +292,7 @@ public class TaskDrawRegular extends AbstractTaskDraw {
 			off.add(new Segment(x1, x2));
 		}
 
-		final HColor backUndone = StyleSignature.of(SName.root, SName.element, SName.ganttDiagram, SName.undone)
+		final HColor backUndone = StyleSignature.of(SName.root, SName.element, SName.ganttDiagram, SName.task, SName.undone)
 				.getMergedStyle(getStyleBuilder()).value(PName.BackGroundColor)
 				.asColor(skinParam.getThemeStyle(), getColorSet());
 


arnaudroques:
Thanks for the patch, but I don't think this is a good idea.
I'm not sure, so let's talk about this :-).

Let's start with `unstarted`.
Since `unstarted` is a child of `task`, when you change the `task` color, it changes also `unstarted` color.
(which is consistent for people who don't want to use a different color for unstarted tasks).

alt text

```
@startgantt
<style>
ganttDiagram {
  task {
    BackGroundColor #0F0
  }
}
</style>
[foo] lasts 21 days
[bar] lasts 30 days and is 10% complete
@endgantt
```

This way, people can optionally change the option of `unstarted` task:

alt text

```
@startgantt
<style>
ganttDiagram {
  task {
    BackGroundColor #0F0
    unstarted {
      BackGroundColor #00F
    }
  }
}
</style>
[foo] lasts 21 days
[bar] lasts 30 days and is 10% complete
@endgantt

```
If we do the same for `undone` (that is, changing `undone` so that it's a child of `task`), when people will only change the `task` background,
it would also change the `undone` color, using the very same color).
So my very first example would display as:

alt text

```
@startgantt
<style>
ganttDiagram {
  task {
    BackGroundColor #0F0
  }
}
</style>
[foo] lasts 21 days
[bar] lasts 30 days and is 10% complete
@endgantt
```

which is not good.

So practically, you are forcing people to always specify both setting.
Do you see the issue?

I agree that the current usage is strange and not symmetrical. I think this is the least bad solution.

What do you think about it?
(PS: anyone has better idea for naming other than `unstarted` and `undone`? )

The-Lum:
You are completely right, I agree also:
>I agree that the current usage is strange and not symmetrical. I think this is the least bad solution.

I was blinded by the symmetry of construction, which is not the same for style...

Sorry for inconvenience... _I close this PR._

PR #892, opened 2022-02-03
improve java17, intellij compatibility

soloturn:


arnaudroques:
I have to revert this, it does not compile well on Java 8
https://github.com/plantuml/plantuml/runs/5048435136?check_suite_focus=true

PR #895, opened 2022-02-03
source code java8 compatible, adjust build files to reflect this

soloturn:
update: while this one adresses the java.time issue as well, kindof, please prefer merging #899 instead.

hey @arnaudroques sorry for ovlerlooking that mvn and gradle had a differing source code compatible level:
https://github.com/plantuml/plantuml/runs/5049289268?check_suite_focus=true

am not sure if java7 was deliberate choice or just historic as i could not find why. so 2 options. this pr, making mvn require java8. and other pr, lowering to java7 for gradle.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -18,7 +18,8 @@ dependencies {
 
 group = "net.sourceforge.plantuml"
 description = "PlantUML"
-java.sourceCompatibility = JavaVersion.VERSION_1_7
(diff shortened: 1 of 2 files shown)


arnaudroques:
Usually, we try to support old version of Java as long as possible.
Now Java7 is probably completely outdated, so ok, let's give up Java7 and switch to Java 8.

I prefer to merge this merge first before dealing with https://github.com/plantuml/plantuml/pull/899

Issue #817, opened 2021-12-02
src is using a few things from java 1.8 


matthew16550:
We saw java version compatibility problems in #809 but they were only apparent when building with java 8.

I've been trying the maven-compiler-plugin [release](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release) option to make this sort of problem visible when building with newer java versions.

Added this to `pom.xml`:

```
<profile>
  <id>ensure-runnable-on-java-7</id>  <!-- See https://stackoverflow.com/a/65655542/1848731 -->
  <activation>
    <jdk>[9,)</jdk>
  </activation>
  <properties>
    <maven.compiler.release>7</maven.compiler.release>
    <maven.compiler.testRelease>8</maven.compiler.testRelease>
  </properties>
</profile>
```

And used java 17 to run `mvn compile` which reports some problems of the current code using 1.8 things:

* A few files use `java.time` & `java.time.format` packages
* `net.sourceforge.plantuml.elk.proxy.Reflect` uses `getParameters()`
* `smetana.core.debug.Purify` uses `getDeclaredAnnotation()`

Have not looked closely, some looks to be debugging code so probably will not affect anyone running on java 7 but perhaps some is a problem?

arnaudroques:
> Have not looked closely, some looks to be debugging code so probably will not affect anyone running on java 7 but perhaps some is a problem?

[elk](https://plantuml.com/elk) and [smetana](https://plantuml.com/smetana02) are used with some specific settings, so it's probably not an issue (we could ask to user to avoid Java 7 for this).

About java.time usage, it seems to be used only for Gantt diagram. Here again, we could impose Java 8 here.

Since Java 7 adoption is probably very weak now, I think we can ignore this.

Thanks for pointing out this anyway!

matthew16550:
Shall I add the `release` option to `pom.xml` set to `8` for src & test ?
That should catch the sort of problem from #809 but will not bother us about the 3 points I listed above.

arnaudroques:
> Shall I add the `release` option to `pom.xml` set to `8` for src & test ?

Doing so will not prevent people using Java 7 to run PlantUML ?

matthew16550:
Upon further reading, `--release` seems to prevent use of `--add-exports` which would be annoying for #698 so now I don't want to do it!

Though I'm fairly sure it would not break 7 for us.

arnaudroques:
So let's go!

PR #899, opened 2022-02-04
use javac --release, java8 minimum version

### LINKED ISSUE #898: use --release flag of recent java versions
since java9 java supports a release flag for building to make sure the code is not using newer APIs. this is supported in maven and gradle:

* https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html
* https://docs.gradle.org/current/userguide/building_java_projects.html

using this flag revealed two challenges:
- javac -target -source flags are not good enough guarantee compatiblity with old versions. use --release instead.
- plantuml uses java.time, introduced in java8. it was built always for java7 and the tools did not notice that installing and using real java7 gives compile and runtime errors.

would be cool if this does not happen any more in future.

soloturn:
fixes #898, and 2 issues:

- javac -target -source flags are not good enough guarantee compatiblity with old versions. use --release instead.
- plantuml uses java.time, introduced in java8. it was built always for java7 and the tools did not notice that installing and using real java7 gives compile and runtime errors.

the cost of using the --release flag is that build environments must muse java11+. but we can be sure, it runs for 8 this time :)

i used the master branch on the fork to produce executables in case you want to test beforehand:
https://github.com/soloturn/plantuml/releases

### DIFF
--- a/.github/workflows/ci-gradle.yml
+++ b/.github/workflows/ci-gradle.yml
@@ -71,12 +71,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java_version: [ 8, 11, 17 ]
+        java_version: [ 11, 17 ]
         os: [ macos-10.15,  macos-11, ubuntu-18.04, ubuntu-20.04, windows-2019, windows-2022 ]
         include:
           - release_from_this_build: true
             os: ubuntu-20.04
-            java_version: 8
+            java_version: 11
     runs-on: ${{ matrix.os }}
     env:
       SIGN_ARTIFACTS: ${{ secrets.ARTIFACT_SIGNING_KEY != '' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,12 +71,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java_version: [ 8, 11, 17 ]
+        java_version: [ 11, 17 ]
         os: [ macos-10.15,  macos-11, ubuntu-18.04, ubuntu-20.04, windows-2019, windows-2022 ]
         include:
           - release_from_this_build: true
             os: ubuntu-20.04
-            java_version: 8
+            java_version: 11
     runs-on: ${{ matrix.os }}
     env:
       SIGN_ARTIFACTS: ${{ secrets.ARTIFACT_SIGNING_KEY != '' }}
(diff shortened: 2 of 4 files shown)


arnaudroques:
Here, I don't understand why you suppress java 8 from `java_version: [ 8, 11, 17 ]`.
We really want to build on Java 8 here :-)

soloturn:
> 
the flag does not exist in java8, it would error out. building with java11 or java17 will give a java8 binary as we desire. contrary to the source/target flags it REALLY checks the old api and fails. if not matched, as we saw by setting --release 7 compiling with java17.

the workflow runs for 17 are a little pointless with this pull request, as they compile a java8 binary. i am planning a pull request so the java compiler version can be passed as parameter, so java17 will be able to build a java17 binary again. just to make sure the compatibility stays high as it is now. 

does this make sense?

arnaudroques:
> the flag does not exist in java8, it would error out. building with java11 or java17 will give a java8 binary as we desire. contrary to the source/target flags it REALLY checks the old api and fails. if not matched, as we saw by setting --release 7 compiling with java17.
> 
> the workflow runs for 17 are a little pointless with this pull request, as they compile a java8 binary. i am planning a pull request so the java compiler version can be passed as parameter, so java17 will be able to build a java17 binary again. just to make sure the compatibility stays high as it is now.
> 
> does this make sense?

Yes, it makes sense and sounds fine to me.
However, this is far from my area of expertise here, so, if you don't mind, I would like to know if it sounds also good to @matthew16550 and/or @The-Lum who have also worked on the CI.

In the mean time, you may fix the conflict I've caused by merging the other pull request (sorry about that...)

Thanks for your help!

matthew16550:
We discussed `--release` in #817 and decided against because it prevents use of `--add-exports` which will be a problem for some tests in #681.  Today I thought more about it and am fairly sure we could compile with `--release` then in a later step run the tests with `--add-exports` so it should not be a problem.

`--release` is helpful, it would be a pity not to use it just because of a few tests that have not even been merged, so I have no objection to adding it here.

We support running on 8, so we should test it.  Whether we compile with 8 is a separate question.

The GH workflow tries to support two things:
1. Making a jar that we might release.
2. Checking that users can compile PlantUML for themselves with various combinations of JDK / OS.

`2` is mainly for highly security conscious orgs which might not have much java expertise and just use `javac` / `jar`.

The workflow mixes together `1` & `2`, it does not make `2` obvious at all, it does not even use `javac` directly.  I was always uncomfortable with how I made this bit.

Perhaps we could change the workflow to this:

```yaml
jobs:
  build:
    # Using the newest LTS Java on Ubuntu, create a potentially releasable jar (compile, test, sign)

  matrix:
    # Test the jar from "build" job
    # Also javac, jar, test (to support "2")

  release:
    # Release the JAR from "build" job (maybe as snapshot)
```

PR #904, opened 2022-02-08
github workflow with various release again

soloturn:
it runs on all platforms with java-17, but the javac release flag
is set to 8, 11, and 17 respectively. artifact upload happens
from latest ubuntu LTS, 20.04, java-8.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,12 +71,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java_version: [ 11, 17 ]
+        javac_release: [ 8, 11, 17 ]
         os: [ macos-10.15,  macos-11, ubuntu-18.04, ubuntu-20.04, windows-2019, windows-2022 ]
         include:
           - release_from_this_build: true
             os: ubuntu-20.04
-            java_version: 11
+            javac_release: 8
     runs-on: ${{ matrix.os }}
     env:
       SIGN_ARTIFACTS: ${{ secrets.ARTIFACT_SIGNING_KEY != '' }}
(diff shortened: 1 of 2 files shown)


soloturn:
it restores building against different jdk api's, and runs as well tests on windows, mac and linux, so filesystem issues may pop up. but it does not yet do what @matthew16550 suggested:
```
jobs:
  build:
    # Using the newest LTS Java on Ubuntu, create a potentially releasable jar (compile, test, sign)

  matrix:
    # Test the jar from "build" job
    # Also javac, jar, test (to support "2")

  release:
    # Release the JAR from "build" job (maybe as snapshot)
```
@matthew16550 you have a proposal to get there, even in steps?

.github/workflows/ci.yml
```diff
@@ -104,10 +104,10 @@ jobs:
 
       # Compile / Test / Package are separate steps so the reason for any failure is more obvious in GitHub UI
       - name: Compile
-        run: mvn --batch-mode compile
+        run: mvn --batch-mode compile -Dmaven.compiler.release=${{ matrix.javac_release }}
```
matthew16550, line 107: I'm not sure if we gain anything building with `maven.compiler.release=11` and `maven.compiler.release=17` because it also builds with `maven.compiler.release=8` so the higher numbers must be ok ?

matthew16550 (changes requested):
@soloturn I don't have the head space to code right now but will try to help with reviews.

Have not looked closely at the Gradle build, only at Maven.

Looks like this PR changes to only run the tests on JDK 17, I think we should keep testing with the older versions because we support people using them.  Also the Java 2D rendering internals are sometimes different in older versions.

PR #912, opened 2022-02-11
switch default build to gradle, fixes #47

### LINKED ISSUE #47: Add Gradle as a build system!

soloturn:
in case it does not work pls revert commit 7d256adde6e761aacd2070bf1a67b72ae7c0e121  to switch the default back to mvn.

### DIFF
--- a/.github/workflows/ci-gradle.yml
+++ b/.github/workflows/ci-gradle.yml
@@ -1,18 +1,18 @@
 name: CI gradle
 
 on:
-#  create:
-#  pull_request:
-#    types: [ opened, synchronize, reopened ]
-#    paths-ignore:
-#      - '*.md'
-#      - 'docs/**'
-#  push:
-#    branches:
-#      - master
-#    paths-ignore:
-#      - '*.md'
-#      - 'docs/**'
+  create:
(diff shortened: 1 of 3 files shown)


soloturn:
the signing failed here, am not sure if it is just because its my user though. in the fork the following settings are there, and it signs the artefacts:

image

arnaudroques:
But why is it failing in on build(8, ubuntu-20.04) only?
And the "create artifacts" step seems to be skipped in other jobs.
Any though?

soloturn:
> But why is it failing in on build(8, ubuntu-20.04) only? And the "create artifacts" step seems to be skipped in other jobs. Any though?

found it. maven has a condition to sign only when the environment variable ```env.SIGN_ARTIFACTS``` is true. on a pull request this is not true, otherwise i could create a pull request and trigger signing with your key. which makes sense. for gradle, this condition was not there.  it is now in the github workflow, in 2 steps. because it does not make a lot of sense in my opinion to call "gradle sign" AND set an environment variable to make it sign.

PR #915, opened 2022-02-12
Add support for passing graphattributes layout parameter through to graphviz

blipper:


blipper:
The dependency is test one so it doesn't impact anything at run time?  Only at build time?  So that should be ok no?

arnaudroques:
> The dependency is test one so it doesn't impact anything at run time? Only at build time? So that should be ok no?

In theory, you are 100% right.
Our main issue is that right now, we do not use gradle on our internal development server.
So if we merge the pull request right now, it will break our development tool.

Maybe (probably?) this will change someday, but right now, we are busy on too many thing to open a new front :-)

soloturn:
> > The dependency is test one so it doesn't impact anything at run time? Only at build time? So that should be ok no?
> 
> In theory, you are 100% right. Our main issue is that right now, we do not use gradle on our internal development server. So if we merge the pull request right now, it will break our development tool.

what are you using, resp what commands are you running?

arnaudroques:
Great!
This has been manually integrated in last commit https://github.com/plantuml/plantuml/commit/e7785ff1448a9aa534c46d001f457c7b322c1ed3.

Thanks again,

PR #917, opened 2022-02-13
create plantuml-pdf.jar and upload it to github

soloturn:
allows to create a jar with dependencies like maven before. contrary to before, this one uploads the jar as well.


The-Lum:
Hello @soloturn, and all,

Just a question:
Are there some reasons that the output file name is named:
- `plantuml-pdf-<version>.jar` or `plantuml-pdf-SNAPSHOT.jar`
- and not `plantuml-<version>-pdf.jar` or `plantuml-SNAPSHOT-pdf.jar` as the other files?

_Just to confirm..._
Regards.

PR #925, opened 2022-02-15
Preserve element group structure in svg output to allow more intelligent post-processing

blipper:


arnaudroques:
Thanks for your contribution, but I have a few questions:

* Could you change:
`Map<UGroupType, String> typeIDent = new HashMap<>();` 
to
`final Map<UGroupType, String> typeIDent = new EnumMap<>(UGroupType.class);`?

* You have added few Logs, could you remove them?

* Why did you comment https://github.com/plantuml/plantuml/blob/master/src/net/sourceforge/plantuml/svg/SvgGraphics.java#L998 ?

Thanks!

arnaudroques:
> It looks like that code was added as a safety check if the closeLink was not called. Since we are now allowing recursive nesting if we kept the check it would close the group before including the url group.

I think this code is useful when there is a link inside an element that already have a link.

For example:

```
@startuml
class Observation  {
name [[[http://link1.com]]]
}
url of Observation is [[http://link2.com]]
@enduml
```

I'm not 100% sure.
Can you check if this example is still working? Thanks!

PR #930, opened 2022-02-16
Add a new !pragma svginterface <true|false> to control INTERACTIVE mode on SVG output and Add support for interactive mouseover/mouseout focus on SVG to focusing on a node and its edges in a complex diagram.  This is enabled with !pragma svginteractive.

blipper:
Can set to either '!pragma svginterface true' or !pragma svginterface false'.  Will default to false.

### DIFF
--- a/src/net/sourceforge/plantuml/sudoku/GraphicsSudoku.java
+++ b/src/net/sourceforge/plantuml/sudoku/GraphicsSudoku.java
@@ -43,7 +43,10 @@ import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.List;
 
-import net.sourceforge.plantuml.*;
+import net.sourceforge.plantuml.Dimension2DDouble;
+import net.sourceforge.plantuml.EmptyImageBuilder;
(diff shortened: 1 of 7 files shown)


blipper:
Are the CI tests flakey?

arnaudroques:
That's a good idea, but before merging this:

- I don't like `import net.sourceforge.plantuml.*;` so please restore imports
- Instead of using a pragma in the constructor of SvgGraphics, use a boolean.

So instead of:

`long seed, String preserveAspectRatio, LengthAdjust lengthAdjust, DarkStrategy darkStrategy, Pragma pragma) {`

use

`long seed, String preserveAspectRatio, LengthAdjust lengthAdjust, DarkStrategy darkStrategy, boolean interactive) {`

Thansk!

blipper:
Done

koppor:
@blipper Could you maybe change the PR title and the text. It is `!pragma svginteractive true` to enable the feature, not `svginterFACE` (capital letters by me).

PR #933, opened 2022-02-24
Add Support for Font Selection and Transparent Background in Ditaa

kolrami:
This adds support for selecting the font in ditaa diagrams by specifying `font-family`, `--font-size` and/or `--font-variant` in the ditaa() line.

The change is actually taken over from @pepijnve who implemented in https://github.com/pepijnve/ditaa/issues/1. I extended it for supporting the necessary flags from plantuml.

Furthermore, I took over the possibility for a transparent background already supported in stathissideris/ditaa.


arnaudroques:

PR #961, opened 2022-03-13
Remove redundant Gantt code

matthew16550:
Hi @arnaudroques, I started looking into details of Gantt code and IntelliJ tells me these pieces are no longer used.

I have not tested this as I do not know enough details to make meaningful test cases, do you have some test files that I can use?

(AFAIK this does not conflict with #952)

fyi @Venthe


matthew16550:
@arnaudroques Oops my bad 🤦, saw your comment in #952.  `core3` is new and unused not old and unused 🤣 .

If we add a `pragma` or something to enable it then IntelliJ will stop distracting me showing the code is unreachable.

PR #955, opened 2022-03-11
Remove maven

soloturn:
you want to remove the maven build or leave it ?

in case remove, this pull request renames the gradle workflow stuff into the original names. and removes the maven pom.

PR #962, opened 2022-03-13
Gradle publish

soloturn:
call it with 
```
gradle publish
```
which will put it into the filesystem, build/repo, so one can see what would go remote. this could be removed later on. if an env variable MAVEN_USERNAME and MAVEN _PASSWORD is exported beforehand, upload it to sonatype oss.

see comments in pull request #955.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -115,6 +115,14 @@ publishing {
 			name = "fstest"
 			url = uri(layout.buildDirectory.dir("repo"))
 		}
+		maven {
+			name = "OSSRH"
+			url = uri("https://oss.sonatype.org/service/local/staging/deploy/maven2/")
+			credentials {
+				username = System.getenv("MAVEN_USERNAME")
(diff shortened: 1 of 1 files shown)


soloturn:
currently i am wondering how to prevent javadoc beeing published, as it is not published currently. if publish from local works, we could extend it to work via github action along these lines:
https://docs.github.com/en/actions/publishing-packages/publishing-java-packages-with-gradle

```
        env:
          MAVEN_USERNAME: ${{ secrets.OSSRH_USERNAME }}
          MAVEN_PASSWORD: ${{ secrets.OSSRH_TOKEN }}
```

arnaudroques:
Many thanks for your help.
You should really add your name somewhere in those files (and your email if you are not afraid of spam :-)
I am really uncomfortable about taking credit for something I did not do.

Note also that there are already two repository secrets `OSSRH_USERNAME` and `OSSRH_PASSWORD` that you could use and see if it works.

PR #973, opened 2022-03-28
Venthe

soloturn:
3 first commits from @Venthe, to split up #952.


soloturn:
took out all but readme change, @arnaudroques . better so?

Issue #635, opened 2021-08-27
Link in `ref` element not displayed


redcatbear:
## Situation

When adding a link `[[...]]` in a `ref` element of a sequence chart, that link does not appear unless it is prefixed by regular text.

Since links are especially useful in `ref` sections, I think this should work.

## Steps to reproduce

```
@startuml
Participant A
ref over A : [[link does not work]]
ref over A : [[link neither does this]] here
ref over A : see [[link this works]]
@enduml
```

### Reproducibility

100%

### Environment

* Plantuml 8059
* dot - graphviz version 2.43.0 (0)
* Linux 5.11.0-31-generic `#33-Ubuntu SMP Wed Aug 11 13:19:04 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

gwilherm:
Are ref supposed to manage links in a special way ?
Because I found a fix for this issue but a removed some code that I don't understand what is it supposed to do.
I believe that I may have break something :D

![image](https://user-images.githubusercontent.com/1017720/160493034-7a00288a-b2c3-4a2d-9f2b-e9b6a0dfe9b7.png)

https://github.com/gwilherm/plantuml/commit/429f4b1598bc7f51f1f35ce963a422fd04750b81

The-Lum:
@gwilherm: _Good, but..._

I liked the idea of setting the link at the beginning for the whole shape of the ref., as:

```puml
@startuml
title test of link on ref
ref over A : [[link]] (Link on the whole shape)\n with words on ref
ref over A : [[https://plantuml.com/link]] (Link on the whole shape)\n with words on ref
@enduml
```

>[![](http://www.plantuml.com/plantuml/svg/dSsn3i8m40FGlL_X4HQwToexVr0wHFIgHAHtKSwGtwULM5aOFDbw7fsXyhNbZebPG772LkIaJvYYojhjWRsbuee1qtHKyupJxJjX5BHeMU0n53dV5IqnebbT_8SHoU93tvSSz7Y_F6ph_u97qMK7FW00)](http://www.plantuml.com/plantuml/uml/dSsn3i8m40FGlL_X4HQwToexVr0wHFIgHAHtKSwGtwULM5aOFDbw7fsXyhNbZebPG772LkIaJvYYojhjWRsbuee1qtHKyupJxJjX5BHeMU0n53dV5IqnebbT_8SHoU93tvSSz7Y_F6ph_u97qMK7FW00)

[...]

PR #979, opened 2022-04-01
Evol/link in ref

gwilherm:
As @The-Lum suggested in https://github.com/plantuml/plantuml/issues/635#issuecomment-1082354399

This change puts the main link of a reference box before the `:` in a sequence diagram

Feel free to discuss any point

### Before:
```
@startuml
ref over A,B : [[mainlink{tooltip}]] inline ref with [[link{tooltip} links]]

ref over A,B
  [[mainlink{tooltip}]]
  with multi line text and [[link{tooltip} links]]
endref
@enduml
```

### After
```
@startuml
ref over A,B [[mainlink{tooltip}]]: [[firstlink]] inline ref with [[link{tooltip} links]]

ref over A,B [[mainlink{tooltip}]]
  [[firstlink]]
  with multi line text and [[link{tooltip} links]]
endref
@enduml
```

### Gives:
image
Whole area link leading to "mainlink"

### DIFF
--- a/src/net/sourceforge/plantuml/sequencediagram/command/CommandReferenceMultilinesOverSeveral.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/command/CommandReferenceMultilinesOverSeveral.java
@@ -55,7 +55,7 @@ import net.sourceforge.plantuml.ugraphic.color.NoSuchColorException;
 public class CommandReferenceMultilinesOverSeveral extends CommandMultilines<SequenceDiagram> {
 
 	public CommandReferenceMultilinesOverSeveral() {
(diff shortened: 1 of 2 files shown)


arnaudroques:
Many thanks for cleaning the code and the syntax.
I think that it somehow break ascending compatibility, but since the impact is very limited, I think it's still ok, since the result is more consistent.

Thanks also to @The-Lum for syntax suggestion.

The-Lum:
Hello @gwilherm, @arnaudroques, and all,

For a to-do list....

- It is plan to do the same thing (adding topurl, ...) for group, alt, loop,...?

_Sorry I am not currently a good java developer._
But I can help to populate the to-do list...

Regards.

arnaudroques:
> * It is plan to do the same thing (adding topurl, ...) for group, alt, loop,...?
> 
> _Sorry I am not currently a good java developer._ But I can help to populate the to-do list...

Well, coding is not the only way to contribute.
And your contributions in documentation, forum and to-do list are really helpful!

So populate the todo list is really a good idea. Giving short examples is also a very good point.

PR #980, opened 2022-04-02
Restore ability to change background color of class/entity header

alfredporter:


alfredporter:
@arnaudroques Test A0005 is now failing because the `header.BackgroundColor` style is changing both the page header and the header of the two classes.  Are my changes correct and this would become the new expected result or should the background color of the classes not be affected by the global `header` style?

arnaudroques:
I guess that you are trying to fix something like the following example (which is not completely working):



```
@startuml
<style>
document {
  header {
    FontColor red
	BackgroundColor lightGray
  }
}
  
  class {
	  FontColor blue
	  BackgroundColor yellow
	  
	  header {
	  	FontSize 20
	  	FontColor green
	  	BackgroundColor red
	  }
}
</style>

header some page header

class Foo {
  dummy
}
@enduml
```

The "Foo" background should be red, and it is not.
We're going to have a look on this.

arnaudroques:
You are right about `Test A0005`. We [have change it](https://github.com/plantuml/plantuml/commit/7cca06b6677e68650ad191ca7b53d67ebca636ef) because the style part was not correct there..

Your pull request was very helpful, but we [have implemented is slightly differently](https://github.com/plantuml/plantuml/commit/7146f93b7950d8957bf7e6a613a5c4ce5ee6c063). 

Tell us if does not solve your issue.

alfredporter:
It does solve my issue, and I see the short-comings in my original commit.  Thank you for the corrections, and for the great work on PlantUML!

The-Lum:
Hello all,

- Could you do the same thing for `object`?

```puml
@startuml
<style>
document {
  header {
    FontColor red
    BackgroundColor lightGray
  }
}
  
object {
  FontColor blue
  BackgroundColor yellow
	  
  header {
    FontSize 20
    FontColor green
    BackgroundColor red
  }
}
</style>

header some page header

Object Foo {
  dummy
}
@enduml
```

>[](http://www.plantuml.com/plantuml/uml/ROz1oi8m48NN_vnYRl2Bwo95eIvTU8Aq6TBg92FJ18dIkojDGT3kXWVlVT_KOpGQasUenfYPTc2bIvv2n2SWzcGiwNmYDXBYGLWKbUoSx4rtTIefs9BpuFfuL9C1SO89447Q2tLbxDDlEH7yrZCnont-4DVGv-51kDrygJWb2YioHR9eLF_BRx2CZk89RyRHGW4u5Sr6P2RQv7s62MeAxvDV)

Regards.

arnaudroques:
Good idea!
[It's done for object](https://www.plantuml.com/plantuml/uml/RO_12i8m44JFxbVi7oYUYnI5UlJW5wJDabOtMKaJ94h_tT0K1Eqb31Dcvcqz1kL3j0pL61BJ0RHqqP8B-0R4dfGcFql4Hbmu2Oj7JtfsZggx6o_HwUBpOFfmzYhbtmcc_8AqD-fAsJVVSYJuZoTYbYTi4DUchyEBSB_xGJ6UoAt05CY2KMsNss0f7SKIFfIXPGKkXR8HcGTrj3RbO4qkI_u0).

Thanks for the suggestion

The-Lum:
Good 👍... _(But always more...)_
 _Then, next steps:_

----
1. Could you do the same thing for `map` and `header of map` ?
See request here:
- https://forum.plantuml.net/14714/can-we-apply-style-to-map?show=14809#c14809

----
2. Could you do the same thing for `fields of class` or `methods of class`?
See request here:
- https://forum.plantuml.net/14563/style-object-allow-style-on-object-diagram?show=14575#c14575

----
_[Have a good night]_

jongracecox:
This issue appears to have regressed in [v1.2022.6](https://github.com/plantuml/plantuml/releases/tag/v1.2022.6). The header background color no longer works. Testing using the following code (taken from above) I get:

```plantuml
@startuml
<style>
document {
  header {
    FontColor red
    BackgroundColor lightGray
  }
}
  
object {
  FontColor blue
  BackgroundColor yellow
	  
  header {
    FontSize 20
    FontColor green
    BackgroundColor red
  }
}
</style>

header some page header

Object Foo {
  dummy
}
@enduml
```

image

Is this still an issue or have I done something wrong?

arnaudroques:
This is strange. It seems to work with [online version](https://www.plantuml.com/plantuml/uml/ROz1QiCm44NNdLFC3LgwDiMqu2wpo0baQv0TZpHXB66KuBidM0w1nBlXm__lJJr5ep5vXcgAcUa7h7J9KuXu1SIUZ2LTJyH6GlmJ5aKbkoQ_fXkTIWgsv3ouFlwhoO2um0A88Es9kZBsxBUS2DxhcPXbXW_4FVHnk11-VxseE2KAEp95icXKdzjli8rEuWdFnj63CekEGt00XwBRYAnacxpFi41DuNxo3G00).

Could you double check the version you are using?
Thanks!

jongracecox:
This is also strange as I think I _was_ using the online version. If it's working online then this probably isn't broken. I will re-test later.

PR #986, opened 2022-04-09
interfaces need no public members

soloturn:


soloturn:
@arnaudroques would be cool to have either a comment or a merge to better know why you like certain pull requests, and others not so much :)

arnaudroques:
> @arnaudroques would be cool to have either a comment or a merge to better know why you like certain pull requests, and others not so much :)

Yes, sorry for the lag...

About this PR, it's really a question of taste.
Even if `public` if not needed, I'd like to emphasize the fact that the method is public.

But once again, it's really a question of taste and habit.

soloturn:
ah, clear. deleted. i like to have no warnings in whatever tool is used so real warnings stick out more. but these ones do not matter at all for that purpose.

Issue #997, opened 2022-04-16
The message is located in the wrong position with its arrow when use most themes


quinnInChina:
this is the result when i use lastest version of plantuml server
```
@startuml
!theme united

note "must on_time < off_time" as n1
state waiting {
	state first : **// create_time == update_time //**
	state second : **// create_time != update_time //**
}
waiting : **// now < on_time && is_delete == 0 //**

[*] --> first : <color #3820e7>export</color>

waiting -> working : <color #3820e7>auto</color>
working : **// now >= on_time //**
working : **// && now <= off_time //**
working : **// && is_delete == 0 //**

working -> second : <color #3820e7>modify</color>
note bottom of working: create rule

working -> offline : <color #3820e7>manual</color>
offline : **// now >= off_time //**
offline : **// && is_delete == 0 //**

offline -> second : <color #3820e7>modify</color>

waiting --> deleted : <color #3820e7>delete</color>
deleted : **// is_delete == 1 //**

deleted -> [*]

@enduml
```

will see 
1、the message "auto" far away from its arrow
2、the title "waiting" has wrong color

The-Lum:
Hello @quinnInChina and all,

>1、the message "auto" far away from its arrow

It seems due to the normal location, see without any theme:

[...]

PR #1002, opened 2022-04-21
upd: fix some `BackGroundColor` when `shadowing` is used

The-Lum:
Attempts to fix [QA-15871](https://forum.plantuml.net/15871/skinparam-shadowing-fills-partitions-similar-spaces-dark)
I observed just one issue on State _(cf. end comparison of https://github.com/plantuml/plantuml/issues/997#issuecomment-1100606593)_

### DIFF
--- a/skin/plantuml.skin
+++ b/skin/plantuml.skin
@@ -82,7 +82,7 @@ element {
 }
 
 group {
-  BackGroundColor transparent
+  BackGroundColor white
   LineThickness 1.0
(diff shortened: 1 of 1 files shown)


The-Lum:
Hello all,

_That is null and void:_
- Not consistent with the fact that it should rather put everything transparent by default, and that above the modifications would change according to the needs the transparency by a color or not.

_Cf. conform to #738_

PR #1028, opened 2022-06-02
Add Swift support for Protocol and Struct in class diagram

bobjacobsen:
This extends the class diagram support to include Protocol and Struct types with circle-P and circle-S letters respectively.  This allows use of the diagrams for the Swift language.

Classes

This PR doesn't include updates for the web pages (see the ["Declaring element"](https://plantuml.com/class-diagram) section, as couldn't find where that was maintained.  All that's needed is to add "protocol" and "struct" examples to that section.

PR #1056, opened 2022-07-13
Add support for `Exception` in class diagram

The-Lum:
Requested by:
- https://forum.plantuml.net/16258/adding-exception-keyword-for-class-diagram

And with the examples or templates (Thanks @bobjacobsen) of:
- #1028

Here is a proposal to support  `Exception` in class diagram.

### DIFF
--- a/skin/plantuml.skin
+++ b/skin/plantuml.skin
@@ -205,6 +205,9 @@ spot {
   spotEntity {
     BackgroundColor #ADD1B2
   }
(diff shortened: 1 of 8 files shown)


The-Lum:
Thanks @arnaudroques,
It was a good usecase to initiate me, _this night_, to [GitHub Codespaces](https://docs.github.com/en/codespaces).

PR #1057, opened 2022-07-14
Include clause overlooked in PR1028

bobjacobsen:
@The-Lum pointed out that I missed a clause while adding Struct and Protocol in PR #1028. See discussion in the PR.  This fixes that oversight.

### DIFF
--- a/src/net/sourceforge/plantuml/classdiagram/command/CommandHideShowByGender.java
+++ b/src/net/sourceforge/plantuml/classdiagram/command/CommandHideShowByGender.java
@@ -131,6 +131,10 @@ public class CommandHideShowByGender extends SingleLineCommand2<UmlDiagram> {
 			gender = EntityGenderUtils.byEntityType(LeafType.ABSTRACT_CLASS);
 		} else if (arg1.equalsIgnoreCase("annotation")) {
 			gender = EntityGenderUtils.byEntityType(LeafType.ANNOTATION);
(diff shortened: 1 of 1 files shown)


arnaudroques:
Thanks for your contribution!
There is a slight conflict to resolve now, could you fix it so that we merge your pull request?
Thanks again.

The-Lum:
Sorry, that is due to my PR:
- #1056

Then _[I am not a Java expert]_, perhaps refactor those lines, in order to avoid duplicate code...

https://github.com/plantuml/plantuml/blob/c5e44fbbb95b1759dd581a0f1bc2b89e620db0bb/src/net/sourceforge/plantuml/classdiagram/command/CommandHideShowByGender.java#L153-L175

Regards.
_[Thanks @bobjacobsen]_

arnaudroques:
> Then _[I am not a Java expert]_, perhaps refactor those lines, in order to avoid duplicate code...

It's really not your fault!
Those duplication were present in the initial code.
We are aware that this is not a good situation.
This is a clear sign that we need some refactoring here and that we need some abstraction (We probably need a `Gender` class to encapsulate all these).
However, as long as a good solution does not emerge, we prefer to keep the code in that non-perfect state (at least, it's easy to understand).

Thanks all for your contributions and BTW there are no such thing as Java expert, there are only people increasing their knowledge in Java :-)

bobjacobsen:
I've resolved the conflict.  Thanks for your patience.

PR #1088, opened 2022-08-16
resolves #1086 replace printStackTrace by java.util.logging.Logger

### LINKED ISSUE #1086: Use java.util.logging.Logger instead of printStackTrace
It makes it possible to bridge/route log messages to another logger. Would you be OK with this change? (I can submit a pull request)

Mogztter:
resolves #1086


arnaudroques:
Thanks for your code!
We have slightly refactored it is last commit https://github.com/plantuml/plantuml/tree/master/src/net/sourceforge/plantuml/log
It's a matter of taste but we rename `Logger` to `Logme` to avoid name clash.

We are also curious to get more info about your changes in https://github.com/plantuml/plantuml/blob/master/src/net/sourceforge/plantuml/ProgressBar.java
 Are you also using this logger ?

Thanks!

Mogztter:
> We have slightly refactored it is last commit https://github.com/plantuml/plantuml/tree/master/src/net/sourceforge/plantuml/log
It's a matter of taste but we rename Logger to Logme to avoid name clash.

We could also name it `LoggerService` or `LogProducer`, naming is hard 😬

Mogztter:
> We are also curious to get more info about your changes in https://github.com/plantuml/plantuml/blob/master/src/net/sourceforge/plantuml/ProgressBar.java
Are you also using this logger ?

I'm using a java.util.Logging logger named `com.plantuml.ProgressBar`. It makes it possible to bridge either `com.plantuml` or `com.plantuml.ProgressBar` or both.

This logger is using a basic handler that prints to stderr the message.

Mogztter:
As a second step, I intend to replace `System.err.println` by a log.info/warn/error depending on the case. Would that be OK?

arnaudroques:
> As a second step, I intend to replace `System.err.println` by a log.info/warn/error depending on the case. Would that be OK?

Sure, that's a good point. The code definitively need some improvements here :-)

Thanks for proposing it!

PR #1183, opened 2022-11-04
Use Unix line endings in the output files

travkin79:
For some reason UmlSource uses Carriage Return and Linefeed as line endings, even if the written file used only Linefeed line endings. In order not to use different line endings in the same file, I suggest to use Linefeed line endings only, for any written file on any OS.

I patched that. Unfortunately, I didn't understand (yet), how exactly the unit tests are working and could not fix all of the unit tests. It would be great if you would take a look at my changes and adapt the tests.

(I assume, the seed values in the test files are some kind of hashes that changed due to other line endings.)


evantill:
tests are failing due to some random seed added to the metadata

maybe releated to this commit
https://github.com/plantuml/plantuml/commit/68be6ae4a0dbfdc6c4f4103f90b7238e0fa4b17a
```diff
> diff  ../actual.txt  ../expected.txt
4c4
< seed: 7067927655347766828
---
> seed: 5605069588648637213
```

arnaudroques:
I will have a look at this one.
However, I am wondering that is the best option between:
  * always use `System.lineSeparator()`
  * always use `\n`

in `UmlSource.getPlainString()`.
Any idea?

evantill:
I would go to allow write using system separator and converting to \n on read

matthew16550:
I agree, `System.lineSeparator()` is usually the best approach.

Only looked briefly, it seems that seed for `UmlSource` does change when the line endings change:

```java
	public long seed() {
		return StringUtils.seed(getPlainString());
	}
```

So that test would have different output on different OS 😢.

Maybe it would be easier to test seed separately from those basic non-regression tests?

travkin79:
Wouldn't it be an even better approach to stay with the line ending of the given source file? E.g. on a Windows system, given a file with Linux line endings only I would expect PlantUML to create output files with Linux line endings, although the current system uses other line endings. Don't you agree?

matthew16550:
@travkin79 that's a good point.  Line endings are such a silly complication, so simple but always wrong for some group of users.

Being OS specific and auto-detecting from each input file are both valid options.

OS specific is probably a quick change.  Adding auto detection would need some effort.

arnaudroques:
We have fixed this in last snapshot, so I close the issue.
Please reopen it if it does not work for you.
Thanks

PR #1215, opened 2022-12-16
Add scale parameter to includegraphics for tikz

barskern:
I've noticed that images that are added to a tikz diagram are not scaled despite the `scale` parameter being passed as mentionned in docs [here](https://crashedmind.github.io/PlantUMLHitchhikersGuide/PlantUMLSpriteLibraries/plantuml_sprites.html#add-images-to-a-diagram). With `includegraphics` it's as easy as setting the `scale` parameter. However I am not to familiar with the plantuml package so I don't know how to access the `scale` parameter from the user. Perhaps I could get some help and we could get this merged? 😄


arnaudroques:
Thanks for your contribution.
We'll finish the job in the incoming days :-)

PR #1135, opened 2022-09-27
Publish to OSSRH

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

@arnaudroques I don't have a way to test

PR #1306, opened 2023-03-04
Proposal : Split the project in smaller modules

evantill:
I would like to split the project in smaller modules to make it easier to work with.

Example of modules coud be : a module for each diagram type, a module for the cli interface etc..

This is the first step : move everything in a module legacy.

What do you think of it ?


evantill:
The jar compile fine on my computer but I have not tested the release process (I don't know how to test it)

arnaudroques:
Well, you should probably [open a new discussion](https://github.com/plantuml/plantuml/discussions) about what you'd like to achieve.

I'm not sure that we could use modules right now because we want to keep Java 8 compatibility.

BTW, we are also refactoring the project into a better package organization to make it easier to work with, but it's still an ongoing task.

evantill:
ok I will open a new discussion.
My first goal is to split the gradle project into a multi-project build.
By the meantime I will update this PR to show what does it look like.

arnaudroques:
Note also that we need to fix https://github.com/plantuml/plantuml/pull/1135 before changing anything on the build.

For some unknown reason, .asc signature files are not published to https://oss.sonatype.org so we still use maven to build PlantUML

evantill:
I will look at #1135.

PR #1307, opened 2023-03-05
Allow only one underscore when escaping -->

rohitsud:
This improves https://github.com/plantuml/plantuml/pull/1303 by preventing expressions like `*--____>` from being allowed. Now only expressions like `*--_>` and `*-->` will be allowed in maps,

### DIFF
--- a/src/net/sourceforge/plantuml/cucadiagram/BodierMap.java
+++ b/src/net/sourceforge/plantuml/cucadiagram/BodierMap.java
@@ -74,7 +74,7 @@ public class BodierMap implements Bodier {
 	}
 
 	public static String getLinkedEntry(String s) {
-		final Pattern p = Pattern.compile("(\\*[-_]+\\>)");
+		final Pattern p = Pattern.compile("(\\*-+_?\\>)");
 		final Matcher m = p.matcher(s);
 		if (m.find()) {
 			return m.group(1);


arnaudroques:
> This improves #1303 by preventing expressions like `*--____>` from being allowed. Now only expressions like `*--_>` and `*-->` will be allowed in maps,

Well, why should we prevent expressions like `*--____>` ? Somehow, we could consider `_` and `-` as synonymous, couldn't we?

rohitsud:
> > This improves #1303 by preventing expressions like `*--____>` from being allowed. Now only expressions like `*--_>` and `*-->` will be allowed in maps,
> 
> Well, why should we prevent expressions like `*--____>` ? Somehow, we could consider `_` and `-` as synonymous, couldn't we?

We could... but the intention of this was to maintain parity with escaping in sequence diagrams where multiple `_` aren't allowed.

The following is invalid:

```
@startuml
Alice -> Bob : hello
Bob --__> Alice : hi
@enduml
```

I think it would be a new concept itself if we allowed using `-` and `_` interchangeably and I was trying to avoid new semantics.

arnaudroques:
> We could... but the intention of this was to maintain parity with escaping in sequence diagrams where multiple `_` aren't allowed.
>  [...]
> 
> I think it would be a new concept itself if we allowed using `-` and `_` interchangeably and I was trying to avoid new semantics.
>

Yes, you are right. 

Sometime, PlantUML language is not very consistent (sorry about that).

So it's probably time to clearly define the use of `_` and to change behavior in *all* diagrams.

So here are the possible options:

1. `_` used as escape character to avoid HTML comment clash. In that case, `_` will be only possible just before the `>`. This `_` will be ignored, so `-->` and `--_>` will be synonymous
2. `_` used as possible character to avoid HTML comment clash. Only one `_` will be only possible just before the `>`. This `_` will be have the same meaning as `-` , so `--->` and `--_>` will be synonymous
3. Treat `_` and `-` as synonymous in any arrows expression, regardless of character position. 

Frankly, we have initially implemented option 3 because it was easier to code.

Even if this will slightly change compatibility, I think we should implement the same behavior everywhere.

And now I find that option 1 is the more logical and elegant solution.
But let's see what other people say, because we will not able to change this rule yet another time :-)

What is your opinion?

rohitsud:
I like option 1. especially since it won't lead with backward compatibility issues for all the folks who are using that solution in sequence diagrams already. 

Option 2. is confusing as a person may inadvertently use a `_` thinking of using it as an escape letter but instead use it for a longer arrow.
3. introduces new semantics which seems unnecessary.

Issue #1313, opened 2023-03-08
missing signature in OSSRH publication


evantill:
see PR #1135  for the the description of the problem.

see https://github.com/plantuml/plantuml/pull/1135#issuecomment-1460868078 about the reason of the missing signature

The solution is to keep the signing properties in the environment variable when calling the publish tasks.

PR #1314, opened 2023-03-08
fix: missing signature in OSSRH publication (#1313)

evantill:
Adding `signingKey` and `signingPassword`properties when calling gradle publication tasks from the CI workflow fix the bug.

Fixes: #1313

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,6 +155,9 @@ jobs:
           GITHUB_TOKEN: ${{ github.token }}
           OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
           OSSRH_PASSWORD: ${{ secrets.OSSRH_PASSWORD }}
+          #do not remove signing key and password or signatures will not be published
+          ORG_GRADLE_PROJECT_signingKey: ${{ secrets.ARTIFACT_SIGNING_KEY }}
+          ORG_GRADLE_PROJECT_signingPassword: ${{ secrets.ARTIFACT_SIGNING_PASSPHRASE }}
         run: |
           .github/scripts/release-snapshot.sh
           gradle publish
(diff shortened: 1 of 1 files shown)


arnaudroques:
That's really great!
We are going to release V1.2023.3 this weekend.
I'm sure it's going to work.

Really, a big thank you!

.github/workflows/ci.yml
```diff
@@ -166,6 +169,9 @@ jobs:
           TAG: ${{ github.event.ref }}
           OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
           OSSRH_PASSWORD: ${{ secrets.OSSRH_PASSWORD }}
+          #do not remove signing key and password or signatures will not be published
+          ORG_GRADLE_PROJECT_signingKey: ${{ secrets.ARTIFACT_SIGNING_KEY }}
```
gliptak, line 173: so having that prefix makes it to build step?

https://github.com/plantuml/plantuml/blob/3c53361064052a3174cf8090c464cb6de60aafd5/build.gradle.kts#L163-L166

evantill: Could you please elaborate on that a little further ? I did not understand your question.

But yes this is related to L163 conditions

gliptak: I was asking if ORG_GRADLE_PROJECT_* prefix allows value to be visible in above `signing` section

evantill:
can you give me the name and description text ?

arnaudroques:
> can you give me the name and description text ?

Sure, we were not very creative... https://github.com/plantuml/plantuml/commit/957dfea0b5063b4f9990452afa0605bb0ac2922d
Name: PlantUML
Description: PlantUML is a component that allows to quickly write diagrams from text.

This may be improved :-)

.github/workflows/ci.yml
evantill: Gradle project properties is realy powerfull.
One way to pass a project property is using enviroment variable.
`ORG_GRADLE_PROJECT_`is a standard prefix. 

Example: if you have a project property named `myProp`, instead of adding `myProp=some_value`in `gradle.properties`you can use an environment variable `ORG_GRADLE_PROJECT_myProp=some_value`

See this part of the documentation : 
https://docs.gradle.org/current/userguide/build_environment.html#sec:project_properties

arnaudroques:
Many thanks also to @gliptak because it's you that put us on the way initially.

Issue #1335, opened 2023-03-17
ClassAttributeFont(Style|Size|…) overrides ClassFontStyle(Style|Size|…)


nstanger:
## Describe the bug

From version 1.2022.2, `ClassAttributeFont(Style|Size|…)` skinparams change font attributes for *all* text in a class, not just the attributes, also overriding the effect of any explicit `ClassFont(Style|Size|…).` This didn’t happen up to version 1.2022.1.

## To Reproduce

Minimum working example:

```plantuml
@startuml
skinparam {
    ClassFontStyle bold
    ClassFontSize 14
    ClassAttributeFontStyle plain
    ClassAttributeFontSize 10
}
class Class {
    -attr1
    --
    +method1()
}
@enduml
```

## Expected behavior

Class title should be size 14 and **bold**, attributes and methods should be size 10 and *not* bold.

## Screenshots

Output of MWE using 1.2022.1 (old default rose skin):
![test_1_2022_1](https://user-images.githubusercontent.com/4316101/225803297-39273da2-79ae-4632-921a-a61b483a27bf.svg)

Class title is larger and **bold**, as expected.

Output of MWE using 1.2023.4 (new default grey skin):
![test_1_2023_4](https://user-images.githubusercontent.com/4316101/225803350-6029f2ef-63e6-4020-aaa5-fe78170e8813.svg)

Class title is the same size as the attributes and *not* bold. I’ve also confirmed the same behaviour in 1.2022.2, 1.2022.14, and 1.2023.0.

## Desktop (please complete the following information)

- OS: macOS Monterey 12.6.3 (behaviour also confirmed under a couple of different Linuxes)
- Browser Vivaldi
- Version 5.7.2921.63 (stable)

## Smartphone (please complete the following information)

N/A

## Additional context

This behaviour first appeared in version 1.2022.2 and is presumably present in all subsequent versions (definitely present in 1.2022.2, 1.2022.14, 1.2023.0, and 1.2023.3). I note that version 1.2022.2 introduced the new default skin, so I wonder whether it is somehow related to that?

It is possible to work around this using Creole and or HTML formatting in the class title, e.g.:

```plantuml
class "<size:14><b>Class2</b></size>" as c2 {
   ...
}

class "<size:14>**Class3**</size>" as c3 {
   ...
}
```

But this is a bit of a hack and regardless it still seems incorrect for `ClassAttributeFont...` to override `ClassFont...`.

[...]

PR #1345, opened 2023-03-19
Build a native image (Linux)

ggrossetie:
The native image is attached to the build as an artifact. For instance: https://github.com/ggrossetie/plantuml/actions/runs/4461027164

### DIFF
--- a/.github/workflows/native-image.yml
+++ b/.github/workflows/native-image.yml
@@ -1,0 +1,50 @@
+name: Native Image
+
+on:
+  push:
+    branches:
+      - native-image
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  workflow_config:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v3
+
+      - name: Get jar
+        run: |
+          wget "https://github.com/plantuml/plantuml/releases/download/${PLANTUML_VERSION}/plantuml-${PLANTUML_VERSION#?}.jar" -O plantuml.jar
+        env:
+          PLANTUML_VERSION: "v1.2023.4"
+
+      - name: Install system requirements
+        run: sudo apt install build-essential libz-dev zlib1g-dev
+
+      - name: Get GraalVM
+        run: |
+          wget "https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-${GRAALVM_VERSION}/graalvm-ce-java17-linux-amd64-${GRAALVM_VERSION}.tar.gz" -O graalvm-ce.tar.gz
+          mkdir graalvm-ce
+          tar xvzf graalvm-ce.tar.gz -C graalvm-ce --strip-components 1
+          ./graalvm-ce/bin/gu install native-image
+        env:
+          GRAALVM_VERSION: "22.3.1"
+
+      - name: Generate GraalVM configuration
+        run: |
+          mkdir native-image-config-dir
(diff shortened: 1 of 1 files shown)


arnaudroques:
Great, thanks for your contribution.

However, could we run [this CI more often](https://github.com/plantuml/plantuml/issues/1335) ? I mean, even on the `main` branch ?
@evantill Any though?

ggrossetie:
> Is it possible to build a native exe for macos and windows ?

It should, I can give it a try 😉 

> However, could we run https://github.com/plantuml/plantuml/issues/1335 ? I mean, even on the main branch ?
@evantill Any though?

I've created a separated workflow to demonstrate how to build a native image but ideally it should run as part of the release process and create a native image using the jar produced for that release.

.github/workflows/native-image.yml
```diff
@@ -0,0 +1,50 @@
...
+          mkdir graalvm-ce
+          tar xvzf graalvm-ce.tar.gz -C graalvm-ce --strip-components 1
+          ./graalvm-ce/bin/gu install native-image
+        env:
+          GRAALVM_VERSION: "22.3.1"
+
+      - name: Generate GraalVM configuration
+        run: |
+          mkdir native-image-config-dir
+          echo 'Bob->Alice: Hello' | ./graalvm-ce/bin/java -agentlib:native-image-agent=config-output-dir=./native-image-config-dir -jar plantuml.jar -tpng -pipe > out.png
```
The-Lum, line 40: Hi @ggrossetie, and all,

_Sorry for the newbie question..._

What are the meaning of those line?
Is an example as `Bob->Alice: Hello` is mandatory in order to generate the `GraalVM configuration`?

Regards.

.github/workflows/native-image.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: sudo apt install build-essential libz-dev zlib1g-dev
+
+      - name: Get GraalVM
+        run: |
+          wget "https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-${GRAALVM_VERSION}/graalvm-ce-java17-linux-amd64-${GRAALVM_VERSION}.tar.gz" -O graalvm-ce.tar.gz
+          mkdir graalvm-ce
+          tar xvzf graalvm-ce.tar.gz -C graalvm-ce --strip-components 1
+          ./graalvm-ce/bin/gu install native-image
+        env:
+          GRAALVM_VERSION: "22.3.1"
```
The-Lum, line 35: Another minor question:
- Should we take only version `22.3.1` of `GraalVM`
- or should we take the latest?

_For `plantuml.jar` that is OK, we will plan to use the current version, the `wip` of the current CI himself._


Regards.

.github/workflows/native-image.yml
ggrossetie: This process is called "Assisted Configuration with Tracing Agent": https://www.graalvm.org/22.0/reference-manual/native-image/Agent/

> Is an example as Bob->Alice: Hello is mandatory in order to generate the GraalVM configuration?

It's just an example and we might need to run PlantUML more than once to trigger all execution paths: "It can be necessary to run the target application more than once with different inputs to trigger separate execution paths for a better coverage of dynamic accesses. The agent supports this with the config-merge-dir option which adds the intercepted accesses to an existing set of configuration files"

For now it seems sufficient to use a simple example such as "Bob->Alice: Hello".

.github/workflows/native-image.yml
ggrossetie: > Should we take only version 22.3.1 of GraalVM

I think we should use a fixed version and manually update it (with careful consideration).

> For plantuml.jar that is OK, we will plan to use the current version, the wip of the current CI himself.

👍🏻

PR #1361, opened 2023-03-26
Append source as comment to generated SVG

walkingice:
# Description

in commit 117102bb65b007f6086838dfc058b278ca69cb38, the comment (source of plantuml that generates SVG file) was dropped.

This commit append it back.

# Reproduce steps

create any valid `foo.puml` file

```bash
$ plantuml foo.puml -tsvg
$ cat foo.svg  # the content of foo.puml was supposed there, but it is not
```

### DIFF
--- a/src/net/sourceforge/plantuml/klimt/drawing/svg/SvgGraphics.java
+++ b/src/net/sourceforge/plantuml/klimt/drawing/svg/SvgGraphics.java
@@ -1025,7 +1025,11 @@ public class SvgGraphics {
 	public void addCommentMetadata(String metadata) {
 		// ::comment when __CORE__
 		final String signature = getMetadataHex(metadata);
-		final String comment = "SRC=[" + signature + "]";
+		final StringBuilder sb = new StringBuilder();
+		sb.append(String.format("SRC=[%s]", signature));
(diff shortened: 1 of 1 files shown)


walkingice:
I was thinking that should I use `$ plantuml -metadata foo.svg` extract source from metadata. But that command only works for generated PNG files. For SVG files, it returns null due to `SImageIO.getImageReaders` cannot find usable reader for SVG format. Since SVG is kind-of readable XML file, maybe appending source as comment is most convenient way?

By the way, by checking the appended `metadata`, it seems the source of plantuml file repeated once. I think this is caused by another part. If it was not an intended behavior, I could try to investigate it.

arnaudroques:
> I was thinking that should I use `$ plantuml -metadata foo.svg` extract source from metadata

Yes, I think this has beed fixed in last release.
Could you double-check it?
Thanks!

walkingice:
Oh! Yes, I just rebased to latest master branch, and I could use `-metadata` to extract source from SVG. I guess this is intended behavior, rather than appending source as comment.

Let me close this PR, thanks for your patient!

walkingice:
This has been fixed. To get the source from a generated SVG, use `plantuml -metadata TheGenerated.svg`

PR #1362, opened 2023-03-26
Build macOS and Windows native images

ggrossetie:
#### What I Did

- Enable cross OS cache (to share cache between ubuntu-latest, windows-latest and macos-latest)
- Replace manual install steps of GraalVM by `graalvm/setup-graalvm@v1` (official GitHub Actions to install GraalVM)
- Enable matrix job to build macOS, Windows and Linux native images using the same steps
- Add `plantuml-darwin-amd64-*` and `plantuml-win-amd64-*.exe` in the release directory

### DIFF
--- a/.github/scripts/release-snapshot.sh
+++ b/.github/scripts/release-snapshot.sh
@@ -20,6 +20,8 @@ ln -s "../libs/plantuml-${RELEASE_VERSION}-javadoc.jar"          "${RELEASE_DIR}
 ln -s "../libs/plantuml-${RELEASE_VERSION}-sources.jar"          "${RELEASE_DIR}/plantuml-SNAPSHOT-sources.jar"
 ln -s "../libs/plantuml-pdf-${RELEASE_VERSION}.jar"              "${RELEASE_DIR}/plantuml-pdf-SNAPSHOT.jar"
 ln -s "../libs/plantuml-linux-amd64-${RELEASE_VERSION}"          "${RELEASE_DIR}/plantuml-linux-amd64-SNAPSHOT"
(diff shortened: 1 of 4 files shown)


arnaudroques:
Thanks for your contribution.
However, we have a minor regression: in version title, we now have a strange `~v` whereas we use to have the beta version of the snapshot (something like `v1.2023.6beta1` )
Any idea why?
Could you restore it?
Thanks again!

ggrossetie:
I changed the `POM_VERSION` to `RELEASE_VERSION`, it's probably related. I will take a closer look.

The-Lum:
Hi all,
I can fixed that tomorrow.
Regards.

arnaudroques:
> Hi all, I can fixed that tomorrow. Regards.

Thanks!
Actually, Guillaume has _partially_ fixed it.

@ggrossetie You can leave it like this, we'll take the point. Thanks again for your work, you did a great job!

The-Lum:
Hi all,

That is now fully fixed (with #1363)
Thanks @ggrossetie.

But the mean question is now:
- Is it necessary or not to change all `POM_VERSION` on `RELEASE_VERSION`?
_Especially on [CI.yml file](https://github.com/plantuml/plantuml/blob/master/.github/workflows/ci.yml)._

_To debate..._
Regards.

ggrossetie:
> Is it necessary or not to change all POM_VERSION on RELEASE_VERSION?

The goal was to use this value in other commands/scripts that's why I moved it outside `release.sh` and `release-snapshot.sh`:

```
echo "release_version=$(gradle properties -q | grep "version:" | awk '{print $2}')" >> $GITHUB_OUTPUT
```

Since it relies on Gradle we cannot move this step in `workflow_config` unless we install Gradle or find another way to get this value.

PR #1365, opened 2023-03-27
feat: support alternate environment variable names

fgreinacher:
At the moment it's not possible to set properties like `plantuml.allowlist.url` via environment variables.

This MR suggests automatic normalization from Java system properties to environment variables by replacing dots by spaces and upper-casing. For the case above the expected environment variable name would be `PLANTUML_ALLOWLIST_URL`.

### DIFF
--- a/src/net/sourceforge/plantuml/security/SecurityUtils.java
+++ b/src/net/sourceforge/plantuml/security/SecurityUtils.java
@@ -210,11 +210,16 @@ public class SecurityUtils {
 	}
 
 	public static String getenv(String name) {
-		final String env = System.getProperty(name);
+		String env = System.getProperty(name);
 		if (StringUtils.isNotEmpty(env))
 			return env;
 
-		return System.getenv(name);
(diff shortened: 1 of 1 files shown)


arnaudroques:
> I don't think it's possible to set an environment variable with a dot:

I think it's possible under Windows.

What I mean is that you can have:
```
String result = System.getenv(name);
if (result==null) {
		final String envName = name.replace(".", "_").toUpperCase();
		result = System.getenv(envName);
}
return result;
```

Does it make sense?

fgreinacher:
Ah wow, crazy Windows :)

Sure, I'll adapt accordingly!

fgreinacher:
@arnaudroques It's adapted now

arnaudroques:
> @arnaudroques It's adapted now

Yes, thanks for your contribution!
And sorry about Windows! :-)

fgreinacher:
@arnaudroques Would be awesome if you could release this together with the fixes for `plantuml-server` 🙇

PR #1395, opened 2023-04-19
Build native images using plantuml-pdf

ggrossetie:
Also, run plantuml-pdf.jar with additional inputs to make sure that all dynamic calls are registered in the native image config.

In addition, includes static resources (.txt and .png) in the native image.


ggrossetie:
@philCryoport This pull request is about building a native images using GraalVM. [Kroki.io](https://kroki.io/) is now using https://github.com/yuzutech/plantuml/releases and it seems to be working fine.

PR #1424, opened 2023-05-11
Added monochrome monospaced theme for better shapes control

bondden:
This is a plain-styled theme with some features:

1. Monospaced fonts bring additional control on diagram layout and shapes widths
2. Some additional styling for Gantt

### DIFF
--- a/themes/puml-theme-mono.puml
+++ b/themes/puml-theme-mono.puml
@@ -1,0 +1,230 @@
+''
+'' A mostly monochrome theme with monospaced font, that adds more structure and tidiness with better shapes control, brought by monospaced font
+''
+'' Original Author: Denis Bondarenko feat Matthew Leather
+''
+
+!$THEME = "mono"
+
+' Theme colors
+
+!if %not(%variable_exists("$CLR_BG"))
+  !$CLR_BG = "white"
+!endif
+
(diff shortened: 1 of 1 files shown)


themes/puml-theme-mono.puml
```diff
@@ -0,0 +1,230 @@
...
+'' Original Author: Denis Bondarenko feat Matthew Leather
+''
+
+!$THEME = "mono"
+
+' Theme colors
+
+!if %not(%variable_exists("$CLR_BG"))
+  !$CLR_BG = "white"
+!endif
```
The-Lum, line 13: Hi @bondden, and all,
 
_Here is a general remark._

The short names are good (`$CLR_BG` VS `$BGCOLOR` VS `$PUML_BGCOLOR`), ... 
But it is not share with all theme!

Also @bondden:
- Perhaps don't use short name for external variable,
_only in a pinch for internal variable..._

@all:
-> But which convention use between: `$BGCOLOR` VS `$PUML_BGCOLOR`...

@PlantUML team, or any contributor:
- Make an inventory of the variable named management on all the current themes...
- Maybe a best practices guide should be created to help the developer to create new themes...

Regards.
Th.

PR #1456, opened 2023-06-14
Java20

soloturn:
java-20 deprecates new URL(), use new URI()
    
cite from the ticket towards openjdk:

The URL class does not itself encode or decode any URL components
according to the escaping mechanism defined in RFC2396. It is the
responsibility of the caller to encode any fields, ...
    
In Java SE 1.4 a new class, java.net.URI, has been added to mitigate
some of the shortcoming of java.net.URL. Conversion methods to create
an URL from an URI were also added.
    
references:
* https://inside.java/2023/02/15/quality-heads-up/
* https://bugs.openjdk.org/browse/JDK-8294241

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@
 *.iml
 out
 
+#vscode files
+/bin
+
 # Ant result file
 plantuml.jar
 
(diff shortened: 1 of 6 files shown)


soloturn:
@arnaudroques oculd you have a look please? rebased now to current master.

Issue #1485, opened 2023-07-21
%ord() function in PlantUML preprocessing?


Potherca:
**Is your feature request related to a problem? Please describe.**
The preprocessor has `%hex2dec()` and `%dec2hex()` functions, which is great.

But I noticed there is no `%ord()` counterpart to the `%chr()` function.

**Describe the solution you'd like**
I would like to have a `%ord()` function that returns a Unicode value for a given character.

**Describe alternatives you've considered**
There isn't really a conceivable alternative to this, within the current limits of preprocessor functionality.

**Additional context**

| Name | Description | Example | Return
-- | -- | -- | --
| `%chr` | Return a character from a given Unicode value | `%chr(65)` |  `A` |
| `%ord` | Return a Unicode value from a given character | `%ord("A")` or `%ord(A)`|  `65` |

PR #1487, opened 2023-07-21
feat: add `ord` builtin function

The-Lum:
To answer to the request:
- #1485

Here is an attempt to add `ord` builtin function.

_Please perhaps to review before to merge..._

[Another question is perhaps also adding a full new complete preprocessor...]

### DIFF
--- a/src/net/sourceforge/plantuml/tim/TContext.java
+++ b/src/net/sourceforge/plantuml/tim/TContext.java
@@ -121,6 +121,7 @@ import net.sourceforge.plantuml.tim.stdlib.LogicalXor;
 import net.sourceforge.plantuml.tim.stdlib.Lower;
 import net.sourceforge.plantuml.tim.stdlib.Newline;
 import net.sourceforge.plantuml.tim.stdlib.Now;
+import net.sourceforge.plantuml.tim.stdlib.Ord;
 import net.sourceforge.plantuml.tim.stdlib.RetrieveProcedure;
 import net.sourceforge.plantuml.tim.stdlib.ReverseColor;
 import net.sourceforge.plantuml.tim.stdlib.ReverseHsluvColor;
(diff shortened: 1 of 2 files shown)


The-Lum:
_Oupppsssss..._

I thought that the `cast` can pass... 
But no, not expert Java for that ...

```.cy
/home/runner/work/plantuml/plantuml/src/net/sourceforge/plantuml/tim/stdlib/Ord.java:62: error: incompatible types: java.lang.String cannot be converted to int
			final int value = (int) values.get(0).toString();
			                                              ^
```

_Sorry for that._
I will go to the code...

arnaudroques:
Thanks for the try!

You should probably split the line in two and could use:

```
final char firstChar = values.get(0).toString().charAt(0);
final int value = (int) firstChar;
```

Now you can even check if `values.get(0).toString().length()` is greater than zero.

Good luck with the code :-)

The-Lum:
Hi @arnaudroques, @Potherca, and all,

_Thanks a lot @arnaudroques, for the hint..._ 🚀 


FYI: this is only a first try and with the restriction of **strict ASCII character**.
_(doesn't work for `Extended ASCII` or `Unicode`... [an code addenda will have to be added (but I'm no expert on this...)])_

Here are the first tests _(for the current implementation)_:

| Input | Output |
| --- | --- |
|  OK        |  ✔️   |
| `%ord(A)`   | 65 |
| `%ord(ABC)` | 65 |
| `%ord("A")` | 65 |
| `%ord('A')` | 65 |
| `%ord(' ')` | 32 |
| `%ord('!')` | 33 |
| `%ord('"')` | 34 |
|             |    |
|  KO        | :x: |
| `%ord(é)`  | :x: |
| `%ord(à)`  | :x: |
| `%ord()`   | :x: |
| `%ord(')`  | :x: |

<details>
<summary>Input test code</summary>

```plantuml
' OK
%ord(A)
%ord(ABC)
%ord("A")
%ord('A')
%ord(' ')
%ord('!')
%ord('"')

' KO
'%ord(é)
'%ord(à)
'%ord()
'%ord(')
```
</details>

Regards,
Th.

Issue #1467, opened 2023-06-23
Qualifier labels and associations' lines not correctly drawn for qualified associations


travkin79:
**Problem description**
When using qualified associations the qualifier labels overlap with classes, are placed on the wrong side of an association, or the association's line between the qualifier and the target class is not visible. In addition, the length of edges cannot be modified if a direction is added to the association definition.

I have a few examples to illustrate that.

This issue is a follow up of a [previous discussion in the user forum](https://forum.plantuml.net/16397/add-qualified-associations-to-class-diagrams). That discussion contains hints to additional issues with qualified associations.

**Steps to reproduce - Example 1**

1. Create a diagram as follows

```
@startuml
    interface Map<K,V>
    class HashMap<Long,Customer>

    Map <|.. HashMap
    Shop [customerId: long] ---> "customer\n1" Customer
    HashMap [id: Long] -r-> "value" Customer
@enduml
```

2. Observe the rendered [result](https://www.plantuml.com/plantuml/uml/JOwz2i8m64NtFiMbSrDmbH8ABefr4bogmqSQRI5DIdwSV7ZJs83huPpBRPqdwyEi2W2Oj9VsIKBYGajphcuyOQ78EHp9ZIlkZ7vLX-2ycQNbHJ8YH_EfwombU1tDWbvivcdOGyNq0SOOHvdvNUzAvBdKRHlefvXqlyIkoPjKa7zoA_KGhty1). Here, the qualifier label "id: Long" overlaps with the class `Customer` and hides the association line between the classes `HashMap` and `Customer` or it doesn't exist at all. There has to be a bigger distance between the classes `HashMap` and `Customer` for the association to be drawn correctly.
Please notice that the association length between the classes `Shop` and `Customer` could successfully be drawn longer than usually by writing `--->` instead of `-->`.

![image](https://github.com/plantuml/plantuml/assets/10887297/5ed57277-6c68-40e5-bb4e-962292d29ae1)

**Expected behavior**
I expected the diagram to look something like
![image](https://github.com/plantuml/plantuml/assets/10887297/ead8bd6c-8dc8-49fd-8d84-aecb18a03d90)

**Steps to reproduce - Example 2**

1. Create a diagram as follows

```
@startuml
    class HashMap<K,V>

    HashMap [key: K] -u-> V1
    HashMap [key: K] -r-> V2
    HashMap [key: K] -d-> V3
    HashMap [key: K] -l-> V4
@enduml
```

[...]

PR #1489, opened 2023-07-24
Patch for #1467

travkin79:
I started improving drawing qualified associations. A few issues still remain.


arnaudroques:
Many thanks for the fix: this part is really tricky.
Note that we have some old users that are still under Java 7(!), so we might rewrite the "stream" part so that it compiles there.
Sorry about that!

travkin79:
No problem, @arnaudroques.
I've update the code in my fork.

Is there a way to add the new commits to this PR (not creating a new PR)?

The-Lum:
_[Just for the record]_
There are some new overlapping defects about "qualified association":
- https://forum.plantuml.net/19066/issues-related-to-qualified-association

Regards,
Th.

PR #1520, opened 2023-08-26
Add some documentation

Benjamin-Davies:
Document some of the classes and packages that I think need it the most. The new docs all either explain the big-picture control flow or document classes whose purpose is difficult to discern without analyzing their usage.

I have made some assumptions about the purpose and meaning of "plasma" and associated classes. I think I am mostly correct.


arnaudroques:
Documentation this code is a huge task :-)
So many thanks for your contribution.

PR #1588, opened 2023-10-27
#1475 Fix show statement

### LINKED ISSUE #1475: Class diagrams: show statement cannot be parsed
**Problem description**
Some class diagrams with `hide` and `show` keywords are not correctly parsed. A diagram that could be rendered successfully with PlantUML 1.2022.7, cannot be rendered anymore with PlantUML 1.2023.9. The `show` statement seems to produce rendering issues.

**Steps to reproduce**
1. Create a diagram with the following content and [try rendering it](https://www.plantuml.com/plantuml/uml/POx1IWCn48Rl-nG_YBH5t0VOKrqzAeZ51yYQqGII6SZCjWVHPpSDgrqmXp3y_pSVirTpnUQSkWr2z0IN4fnPYTDif73iaSc2UEtxlWFGi9hbYOgse7sRMgFkG8ESO85-rn0PRycfOcnWwyUblGWMvWwV7PPt2tQP1XoiHFxuYsyG_O0ag-WANWvoLbPmknjmbEXR-xLoZoltTvdv34uYYHp_O-zNyyFbo7SHeu8NOa-5F4xH0fwJOtjzUiIHYaPXN2_3TjVjApddz0C0)

```
@startuml
    ' hide all attributes and methods...
    hide members
    
    '... but show the methods in class B
    show B methods
    
    class A {
        - name: String
        + id: long
        # doSomething(): void
    }
    class B {
        ~run(): boolean
    }
    class C
    class D
    
    footer Rendered with PlantUML version %version()
@enduml
```

2. You'll see a syntax error in line 8, although the syntax is correct according to the PlantUML [documentation](https://plantuml.com/en/class-diagram#6a8ec84e53ede3ae).

3. Try rendering the same diagram after deleting the line `show B methods` and the diagram will be rendered correctly.

**Expected behavior**
The diagram is rendered as follows, but with the current style. Here is the diagram rendered with an earlier PlantUML version (I think, it was 1.2022.7).
image

**Execution environment**

- OS: macOS Ventura 13.4
- JRE: OpenJDK Runtime Environment Temurin-11.0.15+10 (build 11.0.15+10)
- GraphViz: dot - graphviz version 8.0.5 (20230430.1635)

travkin79:
This patch should fix #1475. I've checked that using a few diagram examples, but automated tests would be more useful, of course.

### DIFF
--- a/src/net/sourceforge/plantuml/abel/EntityGender.java
+++ b/src/net/sourceforge/plantuml/abel/EntityGender.java
@@ -37,4 +37,6 @@ package net.sourceforge.plantuml.abel;
 
 public interface EntityGender {
 	public boolean contains(Entity test);
+
+	public String getGender();
 }
--- a/src/net/sourceforge/plantuml/abel/EntityGenderUtils.java
+++ b/src/net/sourceforge/plantuml/abel/EntityGenderUtils.java
@@ -44,6 +44,11 @@ public class EntityGenderUtils {
 			public boolean contains(Entity test) {
 				return test.getLeafType() == type;
 			}
+
+			@Override
+			public String getGender() {
+				return type.name();
+			}
 		};
 	}
 
@@ -51,6 +56,11 @@ public class EntityGenderUtils {
 		return new EntityGender() {
 			public boolean contains(Entity test) {
 				return test.getUid().equals(entity.getUid());
+			}
+
+			@Override
+			public String getGender() {
+				return entity.getUid();
 			}
 		};
 	}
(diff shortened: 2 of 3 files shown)


arnaudroques:
I am not 100% sure :-), but I think that `quarkInContext()` cannot return `null`.

So is the following test useful?
```
			final Quark<Entity> quark = diagram.quarkInContext(true, diagram.cleanId(arg1));
			if (quark == null) {
				return CommandExecutionResult.error("No such quark " + arg1);
```

Unless you have found such an example (which is possible).

Any though?

travkin79:
Hi @arnaudroques,
I not that familiar with the PlantUML source code. I just wanted to be sure and it seems that `CucaDiagram.quarkInContext(...)` calls `EntityFactory.firstWithName(...)`, which calls `Plasma.firstWithName(...)` and that one seems to sometimes (maybe very rare cases) return `null`. I didn't check all the other cases.

```
        public Quark<DATA> firstWithName(String name) {
		final PEntry<DATA> ent = stats.get(name);
		if (ent == null)
			return null;
		return ent.first;
	}
```

PR #1655, opened 2024-01-27
fix: Restore missing file `HelpTest.java` on an accurate folder.

The-Lum:
Here is a PR, in order to:
- Restore missing file (`HelpTest.java`) on a more accurate folder.

_Ref._:
- https://github.com/plantuml/plantuml/commit/9c62daa35dbaef9c6c6c326243ec66b9c2192d95#diff-8339f059250da046b265ec42196aeabd490484fd12d009ec690def2a7488efe8
- https://github.com/plantuml/plantuml/commit/9c62daa35dbaef9c6c6c326243ec66b9c2192d95
- https://github.com/plantuml/plantuml/commit/e739cd8f5dccfdac2cec085ca37e2f1081b16b5d

---

Creation of this new folder:
- `test\testhelper`

@arnaudroques: Do you agree with the name?

Regards,
Th.

### DIFF
--- a/test/testhelper/TestHelper.java
+++ b/test/testhelper/TestHelper.java
@@ -1,0 +1,23 @@
+package testhelper;
+
+import static net.sourceforge.plantuml.test.PlantUmlTestUtils.exportDiagram;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+
+class TestHelper {
+
+	@Test
+	public void test_help_themes() throws Exception {
+
+		final String output = exportDiagram(
+				"@startuml",
+				"help themes",
+				"@enduml"
+		).asString();
+
+		assertThat(output)
+				.startsWith("                                \nHelp on themes")
+				.contains("bluegray", "hacker");
+	}
+}


arnaudroques:
> @arnaudroques: Do you agree with the name?

Yes, that's a good idea.
Since we are on names and packages, I think you could move all files from `test/net/sourceforge/plantuml/test` to this new `testhelper` package (or to yet another subpackage of `testhelper` ).

Why?

Because it makes sense to have (for example) `test/net/sourceforge/plantuml/security/SFileTest.java` at this place because we are testing a class which is in the actual package `net.sourceforge.plantuml.security` in the PlantUML library.
But there no package `net.sourceforge.plantuml.test` in the PlantUML library, so there should not be any  `test/net/sourceforge/plantuml/test`.

Does it make sense ?

The-Lum:
That makes sense, but...`testhelper` or `testutils` ??..

For naming... there is a difference between `helper` or `utils`...

- https://stackoverflow.com/questions/12192050/what-are-the-differences-between-helper-and-utility-classes#12192155
- https://stackoverflow.com/questions/2676704/naming-convention-for-utility-classes-in-java
- https://www.baeldung.com/java-helper-vs-utility-classes
- ...

I must read more on that, and I don't hesitate to propose you some new PR...

---

And I will also to create a `render` package.
But I don't know where to put them?
Perhaps on the same place... on `test\render`?

_See you on a next PR..._
Regards,
Th.

Issue #1667, opened 2024-02-01
feature: random() function -- with a way to pick a number between X & Y


philCryoport:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
My intended use is for MQTT Packet Identifiers in a sequence diagram.  

Brief overview of MQTT Packet Identifiers:
* MQTT Packet Identifiers range from decimal 1 to decimal 65535 and are used for traceability between:
  * Each MQTT Client -> MQTT Broker request, and 
  * The MQTT Broker's acknowledgment of a particular request in a message back to the transmitting MQTT Client
* MQTT Packet Identifiers have to be unique for each MQTT Client Request -- and with the upper limit being 65535, the Client is expected to sequentially use these Packet Identifier numbers, rolling back to 1 after 65535.

Now in a sequence diagram where I'm highlighting particular requests made by a MQTT Client to the MQTT Broker in the diagram, I _could_ illustrate the use of these MQTT Packet Identifiers by just incrementing the packet identifer by 1, 10, or 100 -- using higher increment quantities to illustrate that there were requests sent in-between the actual particularly highlighted requests.

I thought it would be more realistic to instead increment the MQTT Packet Identifier number by a random value -- say between 101 and 499.

Hence this request for a `random()` function of some sort where I could have PlantUML (semi) randomly pick a number between X & Y (in this case, between 101 and 499)

**Describe alternatives you've considered**
In theory I could hit one of the random number generator websites on the web each time I want to increment the MQTT Packet Identifier.  However, my concern is that the generator website then becomes a dependency for which I have no control.

**Additional context**
[Originally discussed in PlantUML Discord](https://discord.com/channels/1083727021328306236/1202347142161891399/1202347142161891399)

arnaudroques:
Hi!
Implementing a random function in some software that generates documentation seems completely crazy, so I like the idea. :-)

This should be easy to do.

I propose that `%random(5)` would return an integer between 0 and 4. Note that this function will **not** be safe for cryptographic usage...
Does it match your need ?

arnaudroques:
So you can [try this for example](https://www.plantuml.com/plantuml/uml/SoWkIImgAStDuKhCoKnELT2rKqZAJrAmKb0jIinBoS_LC34mqAI9d4DgNWhGNW00).

PR #1668, opened 2024-02-01
feat: Improve new `Random` function

The-Lum:
Hello PlantUML team,

From:
- https://github.com/plantuml/plantuml/issues/1667
- And this first commit: https://github.com/plantuml/plantuml/commit/dbaf8ac2cd50f04ac4ec4309e7dadae234ea8375

Here is a PR in order to improve new builtin `Random` function.

With management of `0`, `1` or `2` param.:
- `random()` -> `0` or `1`
- `random(n)` -> a random integer between `0` and `n-1`
- `random(min, max)` -> a random integer between `min`  and `max-1`

For @arnaudroques:
I use `java.util.Random` like on some other package...
Then I use: `EaterException.located("Error on Random: Too many argument")`...
I don't know if that good or not... Perhaps return 0 in this case...

_(Have a good night...)_
Regards,
Th.

_[FYI: @philCryoport]_

### DIFF
--- a/src/net/sourceforge/plantuml/tim/TContext.java
+++ b/src/net/sourceforge/plantuml/tim/TContext.java
@@ -122,7 +122,7 @@ import net.sourceforge.plantuml.tim.stdlib.Lower;
 import net.sourceforge.plantuml.tim.stdlib.Newline;
 import net.sourceforge.plantuml.tim.stdlib.Now;
 import net.sourceforge.plantuml.tim.stdlib.Ord;
-import net.sourceforge.plantuml.tim.stdlib.Random;
+import net.sourceforge.plantuml.tim.stdlib.RandomFunction;
 import net.sourceforge.plantuml.tim.stdlib.RetrieveProcedure;
 import net.sourceforge.plantuml.tim.stdlib.ReverseColor;
 import net.sourceforge.plantuml.tim.stdlib.ReverseHsluvColor;
--- a/src/net/sourceforge/plantuml/tim/TContext.java
+++ b/src/net/sourceforge/plantuml/tim/TContext.java
@@ -209,7 +209,7 @@ public class TContext {
 		functionsSet.addFunction(new LogicalNor());
 		functionsSet.addFunction(new LogicalNxor());
 		functionsSet.addFunction(new Ord());
-		functionsSet.addFunction(new Random());
+		functionsSet.addFunction(new RandomFunction());
 		// %standard_exists_function
 		// %str_replace
 		// !exit
--- a/src/net/sourceforge/plantuml/tim/stdlib/RandomFunction.java
+++ b/src/net/sourceforge/plantuml/tim/stdlib/RandomFunction.java
@@ -36,6 +36,7 @@ package net.sourceforge.plantuml.tim.stdlib;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Random;
 import java.util.Set;
 
 import net.sourceforge.plantuml.tim.EaterException;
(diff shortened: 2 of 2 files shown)


arnaudroques:
> Then I use: `EaterException.located("Error on Random: Too many argument")`...
> I don't know if that good or not... Perhaps return 0 in this case...

Now, I am not so sure, but I think that, because of:

```
	public boolean canCover(int nbArg, Set<String> namedArgument) {
		return nbArg == 0 || nbArg == 1 || nbArg == 2;
	}
```

I think that the function executeReturnFunction() is called only with 0, 1 or 2 arguments. But, as I said, I am not so sure: perhaps you should test this and tell us :-)

You may also add some tests about this function (in the incoming days :-) )

And finally, you should change:

`final Random random = new Random();
` 

in 

`private final Random random = new Random();
` 

This change encapsulates random within the class, adhering to the principle of least privilege. It's generally a good practice in object-oriented programming to restrict access to class members as much as reasonably possible. To summarize, it is a good practice.

It's really a pity that everything is not private by default in Java, but that's a personal opinion...

philCryoport:
@The-Lum thank you for changing the function name from `Random` to `RandomFunction` to remove the chance of conflict with `java.util.Random`

The-Lum:
Hi all,

To @philCryoport
>thank you for changing the function name from `Random` to `RandomFunction` to remove the chance of conflict with java.util.Random

I had to, I took a model from [`DateFunction`](https://github.com/plantuml/plantuml/blob/master/src/net/sourceforge/plantuml/tim/stdlib/DateFunction.java).

To @arnaudroques 

What are the difference between `EaterException.unlocated` and `EaterException.located`?
_I'm training on the preprocessor..._

Regards,
Th.

arnaudroques:
> What are the difference between `EaterException.unlocated` and `EaterException.located`?
> _I'm training on the preprocessor..._

First, I must extend my apologies regarding the current state of the code: the preprocessor was developed incrementally, with minimal initial planning. Consequently, we've reached a point where it would benefit from some refactoring to enhance its simplicity and coherence.

Let me clarify the distinction between `EaterException` and `EaterExceptionLocated`. Both exceptions are used within the preprocessor to signal an error during its operation. The term "Eater" is employed metaphorically because the preprocessor essentially 'consumes' text, transforming it into a preprocessed format. However, `EaterExceptionLocated` is specifically designed to capture and store information about the exact line—the location—where an error occurs, making it invaluable for providing detailed feedback to the user. On the other hand, `EaterException` lacks this capability, typically because the precise location of the error within the text has been obscured or lost during processing.

When an error arises during the 'eating' process (i.e., while preprocessing), the nature of the available information dictates which exception is raised. If the error's location is known, an `EaterExceptionLocated` is raised to provide detailed context. Otherwise, we resort to raising an `EaterException`.

We have an hack (but this is somehow an ugly hack): every single execution goes through `TContext::executeLines` :

```
	public TValue executeLines(TMemory memory, List<StringLocated> body, TFunctionType ftype, boolean modeSpecial)
			throws EaterExceptionLocated {
		final CodeIterator it = buildCodeIterator(memory, body);

		StringLocated s = null;
		try {
			while ((s = it.peek()) != null) {
				final TValue result = executeOneLineSafe(memory, s, ftype, modeSpecial);
				if (result != null)
					return result;

				it.next();
			}
			return null;
		} catch (EaterException e) {
			throw e.withLocation(s);
		}

	}
```

This method processes each line of code sequentially. If an `EaterException` is thrown during the execution of a line, we catch it and transform it into an `EaterExceptionLocated`. This is possible because, at the point of catching the exception, we have access to the location information—specifically, the line of code being processed when the exception occurred. By converting `EaterException` to `EaterExceptionLocated`, we can provide more detailed error information, including the exact location of the error, which significantly enhances debugging and user feedback.

This approach allows us to handle errors more gracefully by leveraging the additional context available at the moment of the exception. However, it's acknowledged that this method of error handling, though effective, might not be the most elegant solution, indicating an area for potential improvement in the system's design.

The proposal to eliminate `EaterException` from the codebase and exclusively utilize `EaterExceptionLocated` is a strategic move towards simplifying error handling within the system. This approach centralizes the error management mechanism, leveraging the detailed location information provided by `EaterExceptionLocated` for a more precise and informative debugging process. The necessary change involves modifying the `TFunction` interface's method signatures to incorporate `StringLocated` objects, which inherently contain location information, thus streamlining the process of throwing exceptions with location data.

The original `TFunction` interface is defined as follows:

```
public interface TFunction {

...
	public void executeProcedure(TContext context, TMemory memory, LineLocation location, String s)
			throws EaterException, EaterExceptionLocated;

	public TValue executeReturnFunction(TContext context, TMemory memory, LineLocation location, List<TValue> args,
			Map<String, TValue> named) throws EaterException, EaterExceptionLocated;

...
}
```

This can be revised to:

```
public interface TFunction {
...
	public void executeProcedure(TContext context, TMemory memory, StringLocated location)
			throws EaterException, EaterExceptionLocated;

	public TValue executeReturnFunction(TContext context, TMemory memory, StringLocated location, List<TValue> args,
			Map<String, TValue> named) throws EaterException, EaterExceptionLocated;
...
}
```

By adopting `StringLocated` in the interface, we ensure that every function capable of throwing an exception has direct access to the precise location where the error occurred. This adjustment allows all implementations of `TFunction` to directly throw `EaterExceptionLocated` with both the error message and the location encapsulated within a `StringLocated` object, significantly enhancing error traceability.

Furthermore, this change eradicates the need for the previously mentioned workaround—the conversion from `EaterException` to `EaterExceptionLocated`—thereby simplifying the codebase. By enforcing the use of `EaterExceptionLocated` across all relevant parts of the system, error handling becomes more consistent, reducing complexity and improving the maintainability of the code.

In conclusion, by modifying the `TFunction` interface to utilize `StringLocated` and standardizing the use of `EaterExceptionLocated`, the system benefits from streamlined error handling processes. This change not only eliminates the need for less elegant solutions but also enhances the system's overall reliability and user experience in diagnosing and understanding errors.

I'm uncertain if this approach will be effective; the only way to know for sure is to attempt the change. While it may seem complex at first glance, it's actually not as complicated as it appears. Your question has sparked my curiosity as well: could this really work? The only way to find out is by giving it a try—I'm inclined to do so. 😊

To summarize, having two different classes often signals poorly designed code. The good news, however, is that this issue can be rectified. 

I hope my explanation hasn't added to any confusion!

@The-Lum, if you're considering giving it a try, I recommend creating a new branch to experiment. This could be a valuable exercise, despite its complexity. Even if you don't succeed, you'll undoubtedly gain insights into PlantUML code. 😊 Please let me know if you decide to proceed, so we avoid duplicating efforts!

The-Lum:
I think I'm going to have to digest all of this... 🍽️ 

Thanks @arnaudroques...

PR #1691, opened 2024-02-16
fix: suppress `@[start|end]uml` on `sunlust` theme

The-Lum:
That fix:
- https://forum.plantuml.net/18683/sunlust-theme-question

_[FYI @theavege]_


theavege:
@The-Lum, Thanks for the improvements!

PR #1696, opened 2024-02-22
feat: add `%get_all_stdlib` builtin function

The-Lum:
Hi,

_(To continue #1689, and thanks to be69889e :+1:)_

Here is a PR in order to add a `%get_all_stdlib` builtin function on PlantUML.

_Inspired by:_
https://github.com/plantuml/plantuml/blob/be69889e7f4450de7458ea45ac206f97ab3c9c79/src/net/sourceforge/plantuml/preproc/Stdlib.java#L435-L443

Here are the different outputs of `%get_all_stdlib`:
- **Without argument**:  it returns a `JsonArray` of all the stdlib libraries names
- **With whatever one param.**: it returns a `JsonObject` with the below structure:
```json
{
"stdlib_name": {"name": "stdlib_name", "version": "stdlib_version" , "source": "stdlib_src"},
...
}
```

---

:round_pushpin: Before merge:
- Could you have just a look?

Because I changed certain class methods from "private" to "public" on:
https://github.com/plantuml/plantuml/blob/be69889e7f4450de7458ea45ac206f97ab3c9c79/src/net/sourceforge/plantuml/preproc/Stdlib.java#L338
https://github.com/plantuml/plantuml/blob/be69889e7f4450de7458ea45ac206f97ab3c9c79/src/net/sourceforge/plantuml/preproc/Stdlib.java#L454
https://github.com/plantuml/plantuml/blob/be69889e7f4450de7458ea45ac206f97ab3c9c79/src/net/sourceforge/plantuml/preproc/Stdlib.java#L458

- Is it possible?

Another question:
- Where is send the `Logme` exception  message?
- Is it necessary to add also `Log` or `Log.error` to the `stdout`?... 

Regards,
Th.

### DIFF
--- a/src/net/sourceforge/plantuml/preproc/Stdlib.java
+++ b/src/net/sourceforge/plantuml/preproc/Stdlib.java
@@ -335,7 +335,7 @@ public class Stdlib {
 		}
 	}
 
-	private static Collection<String> getAll() throws IOException {
+	public static Collection<String> getAll() throws IOException {
 		final Set<String> result = new TreeSet<>();
 		final InputStream home = getInternalInputStream("home", ".repx");
 		if (home == null)
@@ -451,11 +451,11 @@ public class Stdlib {
 		}
 	}
 
-	private String getVersion() {
+	public String getVersion() {
 		return info.get("VERSION");
 	}
 
-	private String getSource() {
+	public String getSource() {
 		return info.get("SOURCE");
 	}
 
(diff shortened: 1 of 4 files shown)


arnaudroques:
> 
> * Could you have just a look?
> 
> Because I changed certain class methods from "private" to "public" on:
> [...]
> 
> * Is it possible?
> 

Yes, that definitively makes sense!

Thanks for this new function!
But now you have some more works: you have to update the documentation :-)


> 
> * Where is send the `Logme` exception  message?
> * Is it necessary to add also `Log` or `Log.error` to the `stdout`?...
> 

This logging management is not very clear: we probably have to merge the two classes.
However, before doing that, I think we have to open a discussion about how PlantUML should log its messages...

PR #1701, opened 2024-02-28
Add a tool-independent XMI format

DaumAlexande:
This PR adds a new XMI filetype XMI_SCRIPT used with the "-txmi:script" option.
This new format is intended to offer a way to export XMI files for custom scripts, that need to consume plantuml Diagrams without being constrained by the specific XMI support of other applications like ArgoUML or StarUML.
The format is supported for Class diagrams and component diagrams (DescriptionDiagram)

For class-diagrams the format is adapted in the following ways from the ArgoUML format:
* Dashed lines are exported as Dependencies rather than Associations
* Dependencies can reference Attributes and Operations inside Classes based on the Port that the Line connects to
* Slightly changed the Format of AssociationEnd to correspond to the UML format specified in the UML 1.4 XMI schema (https://www.omg.org/spec/UML/20010967/01-02-15.xmi)

For description diagrams the changes are:
* Same handling of Connections as in class Diagrams, but Port location is ignored

These changes only add the specific information we needed for our current usecase.

### DIFF
--- a/src/net/sourceforge/plantuml/FileFormat.java
+++ b/src/net/sourceforge/plantuml/FileFormat.java
@@ -73,6 +73,7 @@ public enum FileFormat {
 	XMI_STANDARD("application/vnd.xmi+xml"), //
 	XMI_STAR("application/vnd.xmi+xml"), //
 	XMI_ARGO("application/vnd.xmi+xml"), //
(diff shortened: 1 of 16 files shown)


arnaudroques:
Many thanks for your contribution, and sorry about the delay...

Issue #945, opened 2022-03-08
Chen Notation for ER-diagrams


HeroGamers:
I looked around for quite a while, and tried to see if anyone had made it possible to use the Chen notation in PlantUML yet, and I could only find the old Issue for the Crow's Foot notation (#39), yet no apparent implementation for the Chen notation, which is what I'm being taught in university, and which seems to be used widely (though it might be an older notation..?)

I'd definitely appreciate having the notation added to PlantUML, and I bet it would be great for others as well!

Some references:
https://vertabelo.com/blog/chen-erd-notation/
https://en.wikipedia.org/wiki/Entity%E2%80%93relationship_model#/media/File:ERD_Representation.svg
Chapter 6 of Database System Concepts (https://www.db-book.com/)

ahorn42:
I really like this this idea!

There was already a suggestion here: https://github.com/plantuml/plantuml/issues/39#issuecomment-474405783

PaienNate:
Really like this idea too.

Benjamin-Davies:
I've started work on an implementation for this. It's mostly done, I just need to change the shape of the relationships.
https://github.com/Benjamin-Davies/plantuml/tree/cheneer

It would be good to get some feedback on the syntax.

```
@startchen movies

entity DIRECTOR {
  key attribute Number
  attr Name {
    Fname
    attribute Lname
  }
  attribute Born
  attribute Died
  attribute Age
}

entity CUSTOMER {
  key attribute Number
  derived attribute Bonus
  multi attribute Name
}

entity DVD {
  key attribute Code
}

relationship RENTED_TO {
  -1 CUSTOMER
  -N DVD

  -(N,M) DIRECTOR

  attribute Date
}

@endchen
```

HeroGamers:
@Benjamin-Davies 

Looks nice - I do have some feedback/questions for the syntax:
- Would it be possible to write relationships like:
```
CUSTOMER -1- RENTED_TO
RENTED_TO -N- DVD
RENTED_TO -(N,M)- DIRECTOR
```
- I saw that you wrote `key` to some of the attributes, but wouldn't it be an idea to keep it similar to how the current Crow's Notation is made in PlantUML, such that the biggest difference actually is the layout, structure and shapes, but otherwise keeps the same way of writing it
- Would this also support adding types, ie. `attribute Born : TIMESTAMP`?

But otherwise great work!

Best regards,
Marcus

Benjamin-Davies:
[...]

PR #1718, opened 2024-03-21
Implement Chen Notation for ER-diagrams

Benjamin-Davies:
Implements #945 

Allows users to conceptually describe database models as Entity-Relational diagrams using Chen's notation. Although this notation appears to have fallen out of use, it is still taught in universities, so it would be nice if those students could use more modern tooling to draw their ER diagrams.

In addition to parsing the new file format, I've also added some new rendering classes as there was no existing way to draw diamonds with text in them or subset symbols in the middle of links.

I've added some simple `nonreg` tests to test the rendering output. I've also added some tests for the behaviour of the new commands.


arnaudroques:
You really did a great job! Many thanks for your contribution.

Now you can think about document it on http://alphadoc.plantuml.com/doc/markdown/en/er-diagram
:-)

Mohammed4077 (approved):

Issue #1726, opened 2024-04-07
[Theme]  `sunlust` theme does not work for JSON or YAML diagram/data


The-Lum:
Hello PlantUML team, (and @theavege),

**Describe the bug**
During [theme galery tests](https://the-lum.github.io/puml-themes-gallery/#sunlust), I discovered this issue:
-  this `sunlust` theme does not work for JSON or YAML diagram/data _[Probably due to some JSON conf. tables on this theme]_

**Tests and Screenshots**
- JSON
```puml
@startjson
!theme sunlust
{
   "fruit":"Apple",
   "size": "Large",
   "color": ["Red", "Green"]
}
@endjson
```
>[![](https://www.planttext.com/api/plantuml/svg/SoWkIImgoIhEp-DIBCb8pKrLA2xDooajBk6gvb9GK51AAohDB56oKd8iACX9LT81YnLdLgKgMIae-IGMfSF4alDpyek0Wj5AGQafIZeAIkv5gQbvIh5SjLmEgNaf80i0)](https://www.planttext.com/?text=SoWkIImgoIhEp-DIBCb8pKrLA2xDooajBk6gvb9GK51AAohDB56oKd8iACX9LT81YnLdLgKgMIae-IGMfSF4alDpyek0Wj5AGQafIZeAIkv5gQbvIh5SjLmEgNaf80i0)

- YAML
```puml
@startyaml
!theme sunlust
fruit: Apple
size: Large
color:
 - Red
 - Green
@endyaml
```
>[![](https://www.planttext.com/api/plantuml/svg/SoWkIImggKpCpU5IBCb8pKrLA2xDooajBk5AAohDBB5IS2meo4dbAiwiIhLIy4aiIa_bIixFoI-oubBGLGXAJG5HxaMfgNbS3gbvAI1p0000)](https://www.planttext.com/?text=SoWkIImggKpCpU5IBCb8pKrLA2xDooajBk5AAohDBB5IS2meo4dbAiwiIhLIy4aiIa_bIixFoI-oubBGLGXAJG5HxaMfgNbS3gbvAI1p0000) 

**PlantUML version**
- Tested on version `1.2024.4`

**Additional context**
_[Origin Ref. or History of this theme: #1677, #1681, #1691]_

The-Lum:
Fixed by #1727.

The-Lum:
Similar request as on #1727:

>Thanks for the first correction.
>But now there is an issue about the size, see example on the original post here:
[...]

PR #1727, opened 2024-04-08
fix json and yaml

theavege:
Thanks for reporting the problem.


The-Lum:
Resolves #1726

Thanks a lot... @theavege 

Regards,
Th.

The-Lum:
Hi @theavege,

Thanks for the first correction.

But now there is an issue about the size, see example on the original post here:
- #1726

- Why the size is too big?

Regards,
Th.

Issue #1735, opened 2024-04-11
Hiding one of multiple stereotypes in containers (e.g. rectangles) hides all stereotypes


travkin79:
**Issue**
The `hide <<some>> stereotype` instruction, in some cases, hides more stereotypes than specified. Consider the following example. Here, we have multiple stereotypes declared for a rectangle and want to hide one of them. instead, all stereotypes are hidden.

If multiple stereotypes are used without hiding one of them, there seems to be no problem (see the package in the following example).

**Steps to reproduce**

1.  Create a diagram with the following source code
```plantuml
@startuml
<style>
root {
    .highlight {
        BackGroundColor LightGray
    }
}
</style>

hide <<highlight>> stereotype

rectangle "System in focus" <<Module>> <<highlight>> {
    class Class
    class ImportantClass <<highlight>>
    class ClassThreadSafe <<thread-safe>> <<Singleton>>
    class ImportantClassThreadSafe <<thread-safe>> <<highlight>>
}

package other <<Stereotype A>> <<Stereotype B>> {
    class ExternalClassThreadSafe <<thread-safe>> <<external>>
    class ExternalImportantClass <<external>> <<highlight>>
}
@enduml  
```
2. Render the diagram with PlantUML, e.g. [using the online server](https://www.plantuml.com/plantuml/uml/ZOzHIWD138RVSuhGT_K2or8iMWHzgXSOTjBkOdOYcGouoDxTx7GfMAWOc8S_-Vylassse5P6XYRRnDI2YXX-0thTzyEfPtzhezPZw3xsAYN5dR0elYxplOQfEcQOeNbOKT0FaR1fBfYsnMoa93Pz4e1IPo6Tc71pc7mmuf3mA5t96q-zIInEQQy0vvzq77B6b-Iq57Ytg5lzvmNu6tBB-juhXNW8HtA_LN6NNVqdKTTSnrpNUnmm0vppJrz_RxmaQJLwU4ifbf5_0000)

**Expected behavior**
The _highlight_ stereotype should be hidden in the rectangle "System in focus" while the _Module_ stereotype should remain visible.

![image](https://github.com/plantuml/plantuml/assets/10887297/8c03ffb4-1247-4716-afe7-55799960e8f9)

**Actual behavior**
Both stereotypes in the rectangle "System in focus" are hidden.

![image](https://github.com/plantuml/plantuml/assets/10887297/7ec0e1a7-cacc-4abe-93ef-b1905db6d03b)

[...]

Issue #1580, opened 2023-10-17
Hiding stereotypes fails if used with multiple stereotypes per element


travkin79:
**Issue**
The hide stereotype instruction doesn't work if an element has more than one stereotype.

**Steps to Reproduce**
Take a look at the following example.

1. Create a diagram with the following source code
```plantuml
@startuml
    hide <<focus>> stereotype
    class Access <<Entity>> <<focus>>
    footer PlantUML version %version()
@enduml
```
2. [Render](https://www.plantuml.com/plantuml/uml/SoWkIImgAStDuL80WepCb5G56vkq_EJIOZixXUAIrABK_9BAWbImR79EOd6nWcDoSYgGih5npIl9BAcqiqDe0AjAoyy7QbC8o4dCAmdrzL4eIoqgpipFKr25CZGqkHnIyrA0LW80) the diagram

**Expected behavior**
The focus stereotype should be hidden while the Entity stereotype should remain visible.

![image](https://github.com/plantuml/plantuml/assets/10887297/e40e9a26-e4fe-4555-bd0f-124f572758cd)

**Actual behavior**
Both stereotypes are shown. The hide instruction is ignored.

![image](https://github.com/plantuml/plantuml/assets/10887297/ceb3ba57-2101-4b7f-bb91-a9bcd3b5597b)

**Desktop**
- OS: MacOS Sonoma 14.0
- PlantUML 1.2023.10
- Graphviz 8.0.5

**Online Server**
- Browser Chrome 118.0.5993.70
- PlantUML 1.2023.11beta3

**Additional context**
Removing the Entity stereotype [leads](https://www.plantuml.com/plantuml/uml/SoWkIImgAStDuL80WepCb5G56vkq_EJIOZixXUAIrABK_9BAWbImR79EOd6nWcDoSYgGWgi2IwNbvmFLAWJa9EQLXFhwA9Ib5XLdvkSfg48P6ffS3gbvAK1x0000) to a hidden focus stereotype.

![image](https://github.com/plantuml/plantuml/assets/10887297/8691af37-7e7a-446f-aabb-a665f40f8d06)

This issue might be related to #1538.

PR #1770, opened 2024-05-07
Fix #1768 Consider "hide stereotype" statement

### LINKED ISSUE #1768: Stereotype won't be hidden with "hide stereotype"
**Describe the bug**
The stereotypes won't be hidden with `hide stereotype` always if there is a stereotype style defined.

**To Reproduce**
The following code
```
@startuml

hide stereotype

rectangle foo <<boo>> {
	rectangle bar <<boo>>
}

@enduml
```
produces in the version 1.2024.3 the following output
image

The same code produces in version 1.2024.5beta1 (online at the moment) the following output:
image

**Expected behavior**
In my point of view, what version 1.2024.3 produces is the expected behaviour (hide all stereotypes, no matter how the rectangle is used).

**additional context**
The problem only occurs if the element has no sprite symbol (e.g. `rectangle <$icon> foo <<boo>> {...}` doesn't have that problem). Altough it is possible that there are now plantuml-stdlib's with a bit unexpected visual behaviours in some situations (I noted the problem in the edgy lib).

travkin79:
Fix for #1768. Also tested with previous cases from issues #1580 and #1735.

### DIFF
--- a/src/net/atmp/CucaDiagram.java
+++ b/src/net/atmp/CucaDiagram.java
@@ -533,6 +533,7 @@ public abstract class CucaDiagram extends UmlDiagram implements GroupHierarchy, 
 			return null;
 		}
 
+		// collect hide or show statements on stereotypes
 		List<EntityHideOrShow> commands = new ArrayList<>();
 		for (EntityHideOrShow hideOrShowEntry : hideOrShows) {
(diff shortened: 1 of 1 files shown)


arnaudroques:
Thanks for the fix!

Issue #1491, opened 2023-07-24
Line ends (arrow heads and others) missing or misplaced in qualified associations


travkin79:
**Problem description**
We've found a few cases where arrow heads (and other line ends) are not correctly drawn, i.e. are misplaced or not visible at all. That seems to affect only qualified associations (which are really tricky) and a few rare cases in component diagrams.

**Steps to reproduce**
Create a file with the following diagram examples and render them with PlantUML version 1.2023.10.

**Example 1**

*Diagram source code*
```plantuml
@startuml
    class HashMap<K,V>

    HashMap [key: K] -u-> V1
    HashMap [key: K] -r-> V2
    HashMap [key: K] -d-> V3
    HashMap [key: K] -l-> V4
@enduml
```

*Rendered diagram*
![image](https://github.com/plantuml/plantuml/assets/10887297/0b67c772-b52b-4801-8690-54f7e912285d)

*Expected diagram*
![image](https://github.com/plantuml/plantuml/assets/10887297/994d5217-a73b-47fa-9953-2ac0fc1d0b21)

**Example 2**

*Diagram source code*
```plantuml
@startuml
    class HashMap<K,V>
    HashMap [a1] <|-u-> [e] V1
    HashMap [b2] *.r.> [f] V2
    HashMap [c3] o.d.> [g] V3
    HashMap [d4] +-l-> [h] V4

    MoreComplex [x: String] ..> X
    MoreComplex [y: int] --> Y
    MoreComplex [z: boolean] ..> Z
@enduml
```

*Rendered diagram*
![image](https://github.com/plantuml/plantuml/assets/10887297/f1977ee3-f19b-4a60-839e-7086ac0165a9)
Here, the two horizontal line's ends on the `HashMap` side (filled diamond and circle with an X) are completely misplaced. They are drawn on the classes `V1` and `V3`. The other line ends are not drawn as starting / ending on the qualifier labels, but on the source and target classes. The qualifier labels are not sticked to the involved classes, but seem to jump away from them to make space for the arrow ends.

*Expected diagram*
In case of these associations with qualifiers on both ends, the lines with their line endings (arrow heads and others) should be drawn between the two qualifier labels. The qualifier labels are to be placed on the side of the corresponding classes like in example 1.

**Example 3**

*Diagram source code*
```plantuml
@startuml
 
skinparam fixCircleLabelOverlapping true
 
() "Product Dubbo" as product_dubbo
() "Member Dubbo" as member_dubbo
[...]

PR #1772, opened 2024-05-08
Add some non-regression tests for certain issues

travkin79:
Test cases for issues #1580, #1735, #1768, and #1491.

Unfortunately, it's not that easy to make regression tests since style and layout details may influence the resulting SVG / PNG images. Thus, these are non-regression tests.


arnaudroques:
Thanks for those tests.
However, it is not clear to me why `build.gradle.kts` should be changed.
Could you tell us more about this? Thanks!

travkin79:
Thank you, @arnaudroques.

Issue #1679, opened 2024-02-09
<-- + note = the location of the Comment is INCORRECT


devwitalik:
**Steps to reproduce the behavior:**
```plantuml
@startuml
Test --> Test: Text
note right: the location of the Comment is correct

Test --> Test: Text
note left: the location of the Comment is correct

Test <-- Test: Text
note right: the location of the Comment is INCORRECT

Test <-- Test: Text
note left: the location of the Comment is INCORRECT
@enduml
```
http://www.plantuml.com/plantuml/uml/SoWkIImgAStDuGf9BIvHqDMrKm0nh81aHGbNNdv9ga9HPde6KA0a8rKX9pyviIGpFqyXFmtCTyxFpKtDAr78B5P8pYygIaqkuS9jJavg6gd6sEZgakaSJpzd_w0WL-SGX4daEGXXa4DgNWeme000

**Screenshots**

![IMG](http://www.plantuml.com/plantuml/png/SoWkIImgAStDuGf9BIvHqDMrKm0nh81aHGbNNdv9ga9HPde6KA0a8rKX9pyviIGpFqyXFmtCTyxFpKtDAr78B5P8pYygIaqkuS9jJavg6gd6sEZgakaSJpzd_w0WL-SGX4daEGXXa4DgNWeme000)

jimnelson372:
Hello.  When working on some related fixes (for #1680 and #839) I saw that comments to the left or right of left directed self messages were off.   Here is how it looks with my current changes.  Please confirm this is what you would expect.  I only changed your script from INCORRECT to [now correct?], otherwise it's the same script you shared:

<img width="682" alt="image" src="https://github.com/plantuml/plantuml/assets/66284321/b0a613de-614e-4e08-a963-b5e9481d6fec">

The-Lum:
Hi all,

_[This is an Issue Review]_ 👀 
This is now fixed on [V1.2024.5](https://github.com/plantuml/plantuml/releases/tag/v1.2024.5).
- Thanks @devwitalik for the report.
- Thanks PlantUML team and @jimnelson372 for your work.

Regards.

Issue #839, opened 2021-12-22
Sequence diagram text misalignment for self-reference/left side/within group


c-george:
# For:
- PlantUML v2.17.0 (within VSCode extension, if that matters)
- sequence diagram

# When:
- participant message to self
- within a group bounding box
- with left-facing arrow

# Erroneous Result:
- group bounding box is rendered as if text were being rendered with right-facing arrow
- bounding box appears to be correct size, just in the wrong horizontal location

# Sample Code:
```
@startuml

group
    a -> a : This works fine
end

group
    a <- a : This does not work
end

@enduml`
```

The-Lum:
Hello @c-george and all,

From:
```puml
@startuml
group
    a -> a : This works fine
end
group
    a <- a : This does not work
end
@enduml
```

Here is the corresponding image:
>[![](http://www.plantuml.com/plantuml/svg/SoWkIImgAStDuKeloYyju580WaG5NJiWOQKGaf5Ph52UNvHTh92McPVAbPgNmeMYpaONeIubFxLO8I-_1Am1hDG1I01D1W00)](http://www.plantuml.com/plantuml/uml/SoWkIImgAStDuKeloYyju580WaG5NJiWOQKGaf5Ph52UNvHTh92McPVAbPgNmeMYpaONeIubFxLO8I-_1Am1hDG1I01D1W00)

Here is a possible workaround, _with adding an hidden arrow_:
```puml
@startuml
group
    a -> a : This works fine
end
group
    a <- a : This does work, but too big
    [<[hidden]- a
end
@enduml
```
>[![](http://www.plantuml.com/plantuml/svg/SoWkIImgAStDuKeloYyju580WaG5NJiWOQKGaf5Ph52UNvHTh92McPVAbPgNWgBCHXUXB2K_5Q9M9oOlgRH4eIG_No4fCnsiCjecEYCp9IKrBnQe0MoA0v00MWe0)](http://www.plantuml.com/plantuml/uml/SoWkIImgAStDuKeloYyju580WaG5NJiWOQKGaf5Ph52UNvHTh92McPVAbPgNWgBCHXUXB2K_5Q9M9oOlgRH4eIG_No4fCnsiCjecEYCp9IKrBnQe0MoA0v00MWe0)

_If that can help,_
Regards.

c-george:
[...]

Issue #1683, opened 2024-02-10
The display of text, line, and arrow on the left side of the Lifeline is incorrect.


devwitalik:
```plantuml
@startuml
skinparam {
   Maxmessagesize 200
}

activate Test
Test <<-- Test : the arrow and text are located inside the Lifeline because they are counted from the right side of the active member's column bar (Lifeline). Which is an incorrect display, right?
Test <<-- Test : also the arrow is not displayed correctly (issue: #1678). (I wonder if the closing of the Lifeline is displayed correctly? Should it also include the arrow before it, i.e. close after it? If not, how do I close the Life Line after the last arrow?)
deactivate Test

@enduml
```
http://www.plantuml.com/plantuml/uml/RP5DwnCn4CRl-HGFUB1_Q4lre58Assj1JmgUPvFPxc1UIYQnhU9tTpTjfQANm0o_vsN8NWlbKeCt-atYYJ85_3G0Fj4bi2eTMUK7u-rcOtuPGxR8TogCBwp5p0zskzMgZTYYZ0pAEPr1qQ7mfKmZmoSxQHmagZXkr4SPs4jaz6ofQbjU6snJZJCyv1GQckKu5ZHb6jgcjM043Ztdrpf9V0qHFMKi7iOlQtmTnOuG1KL8j2bdjWLEzEJfkhpvTlyUGLxJqoMYYEcFZ1tkHlwAXQXMtkBLctVlFqo1YmFEAJhEa5jHwvDAFBRUpqUB_i-lm-SnLUyWfPMOI_lg-AbCpqFAF05Bo9hNBO11GvapIuV3C9TTOfnObt2u0u_mwObtlEqyQRavTo_6yT-_Q_OSNGt-Dm00

**Screenshots**
[...]

PR #1777, opened 2024-05-14
Fixes #1680, #1679, #839 and various other Sequence Diagram layout issues.

### LINKED ISSUE #1680: Grouping message grouping box does not display correctly (if the data is on the left side of the box).
**Steps to reproduce the behavior:**

```plantuml
@startuml
skinparam {
      Maxmessagesize 200
   }

   group Grouping messages
      Test <<-- Test : The group frame does not draw a border around the text (located on the left side), but ignores its presence, and also ignores the presence of a line.
   end
@enduml
```

http://www.plantuml.com/plantuml/uml/FKwxJWD13EplAwO4AK4H9KgHZeekF-1aVSUANVjaUqK4uj_P3GakHYFFmpvua4MhPVAFB2iPLNnFkCuRNIgxqyAUlnZFkzqGVgQ1YsbRyJemouAxxvOyiWVs--tsZxtW-Cwto6nK6KdP8Hf8HfyWdDGI6wXR926wE_WIU2Xwfk04bUkoy1pmdFXnWrCBv4NKUbCEnze9ovate5v1nVKkZ-Y_39txlPA5dyQpB6awT6Yr_080

**Expected behavior**
I expect the group frame to completely enclose the text and line, not ignore their presence and not pass through them as if they didn't exist.  That is, I want the group frame to cover the useful data, and not to take the empty space on the right side into the frame.

**Screenshots**

IMG

jimnelson372:
This collection of fixes all relate to Sequence Diagrams, mostly when using left side self messages, but also in a few cases when using left to right short arrows (e.g ?->D).  

- Issue #839 and #1680 were about groups not properly drawing around left side self messages.  
- issue #1679 was about Notes (both left of and right of these left side self messages) not positioned correctly.  I combined this fix with the group fix, since Notes within groups were equally having layout problems.
- When testing these, I found the Teoz versions were also having problems with Note positioning on Left side self messages, and even had problems drawing a far left left side self message, so found these related fixes.
- Double checking on related Teoz problems, I found a Forum post "Teoz grouping does not embrace all signals", with an example of a short arrow in a group not being drawn within the group.  That issue can be found at:  https://forum.plantuml.net/18832

The changes I made are very isolated to blocks of code dealing with reversed self-arrows.

1. MessageSelfArrow was fixed so that its getStartingX method, used by getMinX, properly accounted for the message being moved the preferred width to the left.  I had to pass a boolean to the component creation to indicate it was a reversal, since that information wasn't obtainable as in other places.
2. ComponentRoseSelfArrow then, instead of drawing itself to the left of its area when drawing a left side self message, the positioning is already managed by the layout, so the drawing stays within the area.   The left side arrow line itself, then, needed to be drawn at the far right of the area.  Very straight forward.
3. Step1Abstract just needed to check if the note was on a reversed self message and instead of pushing the noteBox to the right a positive width, it now pushes a negative width.  This solved the correct positioning of the Left side Note.
4. Step1Message just needed to pass the boolean of whether the self message was reversed to the MessageSelfArrow as mentioned in change 1.
5. To fix the short arrow problem in Teoz's grouping, I just had to add a condition to CommunicationExoTile for isShortArrow in the getPoint1 method, again used by its getMinX to have it positioned correctly so that the Group would know its correct location.
6. CommunicationTileSelf just needed a fix to its getMinX to account for the correct positioning of the left side self message, and the drawU method just needed to use this updated call, rather than calculating the x1 differently.
7. I created a matching pair to the CommunicationTileSelfNoteRight class, calling it CommunicationTileSelfNoteLeft.   It properly positions the left side note when a self message is also left side.  TileBuilder was updated to create this new class under the appropriate conditions (LEFT and self message).
8. Finally, ArrowAndNotBox just removes what appears to be a bug, where noteBox was Right Shifted by getStartingY when the NotePosition was RIGHT.  First, in all my tests it doesn't seem that a right shift was needed there anymore, and second if it was, it should be based on an X coordinate, not a Y coordinate.  Again, all works with that removed.  I suspect that shift was either ignored, or the value of Y was small enough not to make a difference.

While I tested this in cases beyond any of the reported problems, and in cases where there weren't problems to make sure, Here are examples of what the reported issues now will look like:

For #1680, the grouping now is correctly around the message text.
1680testgroupleftself

For #839, the bottom group now properly just surrounds the self message:
839testgroupleftself

For #1679, the bottom two notes are not positioned correctly:
1679testgroupleftself

For Forum 18832, the short arrow (now labled M2 [Grouped[) is correctly captured within the group. This is with Teoz turned on.
18832testgroupleftself

Finally, I had mentioned fixing other problems Teoz had with the left side self messages.  For instance the following script:

```
@startuml
!pragma teoz true
skinparam {
  Maxmessagesize 200
}

group Grouping messages
    Test <- Test    : The group frame [now] does draw a border around the text (located on the left side), [lo longer] ignores its presence, and [no longer]] ignores the presence of a line.
    note right
      A note on the self message
    endnote
end
@enduml
```
Is currently generating:
image

But will now correctly generate (even with a right arrow):
testgroupleftself
It is similar even without the grouping.

 If the note is on the left, it currently generates:
image
But will now generate:
testgroupleftself
Which, of course, matches the right side version nicely (already working):
testgroupleftself

Let me know if you need me to change anything, or if you have any questions about my changes.

(There are a few more group sizing issues related to Teoz that I didn't include here, since they were already much closer to working that the example in the issues above, mostly involving the grouping being to large in some cases.   For now, I focused just on these issues related to the left side self messages and notes.)

Regards,
Jim Nelson
jimnelson372

### DIFF
--- a/src/net/sourceforge/plantuml/sequencediagram/graphic/ArrowAndNoteBox.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/graphic/ArrowAndNoteBox.java
@@ -105,9 +105,6 @@ class ArrowAndNoteBox extends Arrow implements InGroupable {
 		double result = w;
 		for (NoteBox noteBox : noteBoxes) {
 			result += noteBox.getPreferredWidth(stringBounder);
-			if (noteBox.getNotePosition() == NotePosition.RIGHT) {
-				result += noteBox.getRightShift(arrow.getStartingY());
-			}
 		}
 		return result;
 	}
--- a/src/net/sourceforge/plantuml/sequencediagram/graphic/MessageSelfArrow.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/graphic/MessageSelfArrow.java
@@ -51,13 +51,15 @@ class MessageSelfArrow extends Arrow {
 	private final LivingParticipantBox p1;
 	private final double deltaX;
 	private final double deltaY;
+	private final boolean isReverse;
 
 	public MessageSelfArrow(double startingY, Rose skin, ArrowComponent arrow, LivingParticipantBox p1, double deltaY,
-			Url url, double deltaX) {
+													Url url, double deltaX, boolean isReverse) {
 		super(startingY, skin, arrow, url);
 		this.p1 = p1;
 		this.deltaY = deltaY;
 		this.deltaX = deltaX;
+		this.isReverse = isReverse;
 	}
 
 	@Override
@@ -92,7 +94,10 @@ class MessageSelfArrow extends Arrow {
 		// }
 		final double pos2 = p1.getLiveThicknessAt(stringBounder, getArrowYStartLevel(stringBounder)).getSegment()
 				.getPos2();
-		return pos2 + deltaX;
+		if (isReverse)
+			return pos2 + deltaX - getPreferredWidth(stringBounder);
+		else
+			return pos2 + deltaX;
 	}
 
 	@Override
@@ -142,5 +147,9 @@ class MessageSelfArrow extends Arrow {
 	@Override
 	public double getActualWidth(StringBounder stringBounder) {
 		return getPreferredWidth(stringBounder);
+	}
+
+	public boolean isReverse() {
+		return isReverse;
 	}
 }
--- a/src/net/sourceforge/plantuml/sequencediagram/graphic/Step1Abstract.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/graphic/Step1Abstract.java
@@ -119,8 +119,14 @@ abstract class Step1Abstract {
 		final NoteBox noteBox = new NoteBox(arrow.getStartingY(), noteComp, p, null, noteOnMessage.getPosition(),
 				noteOnMessage.getUrl());
 
-		if (arrow instanceof MessageSelfArrow && noteOnMessage.getPosition() == NotePosition.RIGHT) {
-			noteBox.pushToRight(arrow.getPreferredWidth(stringBounder));
+		if (arrow instanceof MessageSelfArrow) {
+			boolean isReverseSelfArrow = ((MessageSelfArrow) arrow).isReverse();
+			if (!isReverseSelfArrow && noteOnMessage.getPosition() == NotePosition.RIGHT) {
+				noteBox.pushToRight(arrow.getPreferredWidth(stringBounder));
+			}
+			if (isReverseSelfArrow && noteOnMessage.getPosition() == NotePosition.LEFT) {
+				noteBox.pushToRight(-arrow.getPreferredWidth(stringBounder));
+			}
 		}
 		// if (arrow instanceof MessageExoArrow) {
 		// final MessageExoType type = ((MessageExoArrow) arrow).getType();
--- a/src/net/sourceforge/plantuml/sequencediagram/graphic/Step1Message.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/graphic/Step1Message.java
@@ -216,7 +216,7 @@ class Step1Message extends Step1Abstract {
 		final ArrowComponent comp = getDrawingSet().getSkin().createComponentArrow(styles, getConfig(),
 				getDrawingSet().getSkinParam(), getMessage().getLabelNumbered());
 		return new MessageSelfArrow(posY, getDrawingSet().getSkin(), comp, getLivingParticipantBox1(), deltaY,
-				getMessage().getUrl(), deltaX);
+				getMessage().getUrl(), deltaX,getConfig().isReverseDefine());
 	}
 
 	private double getHalfLifeWidth() {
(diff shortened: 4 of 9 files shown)


arnaudroques:
Well, you did a great job: those issues are not easy to fix.

However, note placement can be a nightmare :-)

I think there is a regression in the following example:

```
@startuml
A -> B : a
note right: Note
activate B
B --> A : b
deactivate B
@enduml
```

The note is partially hidden. Could you have a look at it?

It would also be great if you could add your examples (and mine :-) to the regression tests https://github.com/plantuml/plantuml/tree/master/test/nonreg/simple

Does it sound good?

jimnelson372:
Thanks @arnaudroques.   

And oh, that is a regression!   Ouch.  That my change caused an active timeline to block a right side note does surprise me, but there it is.  And, unfortunately, I believe it's due to my "fix" to the ArrowAndNoteBox that I considered a bug.  I will take a second look at that later this evening or in the morning.

I'll look at how those regression tests are written and add to those tests.   

Quick question, how did you find the regression example you shared?   I just ran the regression tests with these latest changes in place and I'm not seeing the error showing up there.

Regards.

jimnelson372:
Ah.  I apologize for my complete misunderstanding of that block of code in ArrowAndNoteBox.getPreferredWith(...).  That block was:

```
			if (noteBox.getNotePosition() == NotePosition.RIGHT) {
				result += noteBox.getRightShift(arrow.getStartingY());
			}
```
When I mistakenly thought it was calculating the amount of a right shift based on the input, and I was confused why the input should be  a Y value.   But when I stepped into it and saw method names such as getRightShiftAtLevel(..Y), I finally grasped that it is looking up the appropriate right shift at a given horizontal level in the sequence diagram.  Now it makes sense?  I will create a Pull Request to revert that class back to what it was.   I should have left it alone anyway since it ultimately turned out not to impact the rest of my fixes in this pull request.

Though I'll review the test classes to see if I can include some tests in that pull request as well.

Regards,
Jim Nelson

arnaudroques:
> Quick question, how did you find the regression example you shared? I just ran the regression tests with these latest changes in place and I'm not seeing the error showing up there.

Over the past 15 years, I have compiled a collection of a few thousand diagrams that I use for regression testing. As bugs and updates arise, I save example diagrams of varying complexity.

These diagrams are not published: some of them contain data from users who have sent them to me, and I cannot share them as is. Sorting them manually is quite tedious.

Another major challenge is that these regression tests are not fully automated. I generate an HTML file with the images that have changed, and **it then requires visual inspection** to determine if a diagram is significantly different or not.

It is probably time to industrialize this process. However, putting all these tests in https://github.com/plantuml/plantuml/tree/master/test/nonreg/simple does not seem like a good idea. First, because the tests themselves take a lot of time. Second, because every code modification always results in small differences, and I don't see how to completely automate the comparison.

We need a system that can intelligently compare two diagram images.

If you are interested, we could try a new GitHub project, "plantuml-nonreg," dedicated solely to calculating the delta between two versions of PlantUML. Perhaps with a scoring system?

I don't have much time, but it could be an interesting project, and I am willing to commit the publishable part of my diagram collection to this non-regression project.

jimnelson372:
> We need a system that can intelligently compare two diagram images.
> 
> If you are interested, we could try a new GitHub project, "plantuml-nonreg," dedicated solely to calculating the delta between two versions of PlantUML. Perhaps with a scoring system?

It is an interesting issue.   Right now, I am focused on learning more about the layout processes itself, while helping to fix reported issues.  Maybe after I have more experience with the code and the variety of diagrams themselves I can give this more thought.

arnaudroques:
> Hi @jimnelson372
> 
> > I am focused on learning more about the layout processes itself, while helping to fix reported issues
> 
> If you search for a new bug (in area of sequence diagram and teoz) 😏 can you have a look on [teoz produces OutOfMemoryError](https://forum.plantuml.net/18882/pragma-produces-outofmemoryerror-times-slower-memory-added) too?
> 
> Thank you Helmut

The good news is that I've found where the issue come from :-)
It's quite easy to fix: adding some cache in `getCurrentValueInternal()` method from `RealMin` and `RealMax` class.

I cannot commit the correction right now, but that will be done before the end of next week.
Stay tuned!

jimnelson372:
Hello @arnaudroques,

While working on issue #1683, I discovered that I left incomplete the positioning of the "o" arrow decorators in my fix to ComponentRoseSelfArrow.  My change is causing the o's to draw to the left side of the comment, rather than along the timeline.   I can include that fix when I do the other arrow drawing fixes for the left side self messages for issue #1683, but if you need this earlier, I can do a smaller pull request just to correct the decorator positioning.   Just let  me know if you do need it quickly.   If all goes well, I'll have the #1683 fixes completed before the end of the weekend.

Regards,
Jim Nelson

arnaudroques:
> Hello @arnaudroques,
> 
> While working on issue #1683, I discovered that I left incomplete the positioning of the "o" arrow decorators in my fix to ComponentRoseSelfArrow. My change is causing the o's to draw to the left side of the comment, rather than along the timeline. I can include that fix when I do the other arrow drawing fixes for the left side self messages for issue #1683, but if you need this earlier, I can do a smaller pull request just to correct the decorator positioning. Just let me know if you do need it quickly. If all goes well, I'll have the #1683 fixes completed before the end of the weekend.
> 
> Regards, Jim Nelson

No, don't worry: I do not need it quickly.
Just take your time to fix it: we will release an official version only next week-end.

Good luck in your debug :-)
Thanks again

kirchsth:
Hi @arnaudroques and @jimnelson372

thank you for the fixes.

Helmut

Issue #1494, opened 2023-07-26
Sequence diagram: Wrong start/end of activation boxes in activation boxes while using `teoz`


HeinrichAD:
**Describe the bug**
While using `teoz` in sequence diagrams, the activation boxes are shifted or do not start/end at the correct position.

| Current behavior |  Problems | Expected behavior |
:--:|:--:|:--:
| ![Current behavior](https://github.com/plantuml/plantuml/assets/5962361/54411408-52d9-4e95-b462-7b7c00695bdc) | ![Problems](https://github.com/plantuml/plantuml/assets/5962361/221bb927-1fe5-46c8-8048-871da8e80069) | ![Expected behavio](https://github.com/plantuml/plantuml/assets/5962361/fa778344-e3c4-4bda-9164-1ee49acb7a4c) |

**To Reproduce**
[Example code rendered on plantuml.com](https://www.plantuml.com/plantuml/uml/JOyz3eCm38Ltdy8iNH0Nw820gKiFuGIhYfeGv9XAzFJb9urOBBttlMVBNL8INOA7QXPw1KBb-4MLXI6ztJIJKC2ag-TjYhD75dghKV19taTPaHA6R0tHxD9asNjdUJS8WAow3oblq63JjaZOD7MDTtouCrW17S3ymR3olh62qYXsBjpm9E5QoGn6BZVfAW9qF8tR8py0)

```plantuml
@startuml
!pragma teoz true
skinparam style strictuml

Actor Mallory as m
Actor Bob as b
Actor Alice as a

activate b
b ->> a --++ : Hi Alice!
a ->> b --++ : Bye Bob!

b ->> a ++ : Bye Alice!
& b ->> m --++ : Bye Alice!

deactivate a
deactivate m
@enduml
```

**Desktop**

- OS: Windows 10
- Browser: Firefox
- Version: 102.12.0esr (64-Bit)

jubr:
This teoz bug has been around for a while, see https://github.com/plantuml/plantuml/issues/1099. Could you please take a look? Tnx so much!

jimnelson372:
Looking into these problems now.   I have a fix for the --++ condition, but I'm working on the & conditions.  So, for the moment, my fix improves it to:

![testteozlifeline_006](https://github.com/plantuml/plantuml/assets/66284321/1248bd78-d0c4-4499-ba56-5c103734a3ef)

So just that part on Bob after the Bye Alice messages is still a problem.   Again, working on it now.

[...]

Issue #1099, opened 2022-08-27
sequence lifeline height badly depends on message height


jhoehle:
When connecting arrow messages with &, the height of activation lines is miscomputed except in the simple case where all message texts share the same text height.
```
@startuml
!pragma teoz true
  A <[#red]-o? ++
  A -> B -- : very long\nmulti-line
& B -> C ++ #red: single
note over C: got it
Deactivate C
====
  A <-o? ++ #green
  A -> B -- : very long\nmulti-line
& B -> C ++ #green : same\nheight
note over C: got it
Deactivate C
====
  A <-o? ++ #red
  A -> B -- : single
& B -> C ++ : very long\nmulti-line
note over C: got it
Deactivate C
@enduml
```
![sequence diagram](https://www.plantuml.com/plantuml/png/dS_1QiCm30RWkv_Y7u5TWbyWR5csx2sM7SmY78Cj3qSEj4zVfpxbK4ghewIVJnfMCKboyEhbFnaR38JY6P8oAU0JRpzDelbNnm-qxRMZUtn1QtJOA9tW8zk9G_RYj7TCwdKVznZBFlPiXzMnzQGu2Y6M4COEDWgSg6yoV-8s8uHHlPUgQENGs4J43wirXWARG1ClvEmYp_29vYDUNpbeDo-wwmt4Smx-0W00)

The third case demonstrates how the height of a lifeline must depend on all the following &-connected arrows.
Work-around: Manually adjust all texts to the same height, e.g. using `\nsingle`. Using that trick, I could correct _some_ misplaced lifelines in my diagrams.

Bug too: With Teoz, `<-o?` without text looks awfully short. The standard engine produces a nice view.

Documentation bug: [The online documentation](https://plantuml.com/sequence-diagram) uses &-notation in section "Slanted or odd arrows" without introducing it. Combining arrows using & should be added to documentation bug #261

jubr:
Hi @The-Lum, would it be possible to take a look at this one please? This teoz bug been around for ages and I'm still running into this on a daily basis :'(
Please let me know if more illustrative examples are required, tnx!

jimnelson372:
I have submitted the fix in pull request #1793.   With this fix, your example generates:

![testteozlifeline_005](https://github.com/plantuml/plantuml/assets/66284321/8e607076-d472-4ed6-adfd-30d15558b8dd)

Regards,

Jim Nelson

Issue #739, opened 2021-11-01
teoz: activation breaks parallell lines


chipbite:
Hello!

Strangely, when using activation, it interferes with teoz rendering, so that messages using &, despite that do not render on same row.
**Example:**

```plantuml
@startuml
!pragma teoz true
autonumber 
AAA -> BBB++: Message 0
BBB -[#22A722]> DDD: Message 1
deactivate BBB
activate DDD
& DDD -> EEE: Message 2
note right #red: <-- Expect 1 & 2 to be same row
DDD -[#22A722]> FFF++--: Msg 3
& FFF -> GGG--: Msg 4
note right #red: <-- Expect 3 & 4 to be same row
@enduml
```

![image](https://user-images.githubusercontent.com/559199/139748281-5b6f21b5-e0b9-4c33-9996-e5e4984420ff.png)

Here is rendering:
![image](http://www.plantuml.com/plantuml/png/VP1FIyD04CNlyocU57e92tPR48gK9kJFoKyW7gRD40icMtOdAdvwPmyQzD3Bmhvvx_s6EIRbgFFuHdVNoCF8K0bVq3WByQnXciUJH51Pbd07L5LL57iyIKey2Ex91BZdbVVbW_Sl1zHrlOmtr0kVzVBEAZbAln-pqJg_kRHfcYNZQGfcY9VXLR6Aqk_nw1oQpwkS5HkiuQ419q7YqNpXWwpcpmvjsnQ5SzQP1coDOqBcT5tte-vkKhP6sVsd76NgxKxV)

[Link to this example on plantuml server](http://www.plantuml.com/plantuml/uml/VP1FIyD04CNlyocU57e92tPR48gK9kJFoKyW7gRD40icMtOdAdvwPmyQzD3Bmhvvx_s6EIRbgFFuHdVNoCF8K0bVq3WByQnXciUJH51Pbd07L5LL57iyIKey2Ex91BZdbVVbW_Sl1zHrlOmtr0kVzVBEAZbAln-pqJg_kRHfcYNZQGfcY9VXLR6Aqk_nw1oQpwkS5HkiuQ419q7YqNpXWwpcpmvjsnQ5SzQP1coDOqBcT5tte-vkKhP6sVsd76NgxKxV).

If you comment out activation, (lines 8 & 9 and ++-- on line 12) it is rendered as expected.

Thanks for a great tool!

BR! /mawi

PS 
[...]

Issue #656, opened 2021-09-06
BUG: Activation is offset (Sequence diagram, Teoz)


chipbite:
This was reported previously: https://forum.plantuml.net/13569/activation-is-offset-using-teoz

Hello, 
Using teoz, for a message chain, where each participant gets activated, there is an offset, where only the first participant gets the lifeline drawn correctly (IMO), and the subsequent are drawn differently, with an offset relative to the first. Well, it is easier to see in a diagram, so check below ;-) ...

Code:
```
A -> B++: Get data
& A -> C++: 
& A -> D++: 
B -->> A--: Data
& C -->> A--: 
& D -->> A--:

```
**Current actual result for this:**
![image](https://user-images.githubusercontent.com/559199/132241126-2b48a8c0-beed-45ea-9417-f3032da3a32b.png)

**Actual / Generated result:**
![generated diagram](http://www.plantuml.com/plantuml/png/JO_DIiKm44RtynIV6pUNTEEkYzAVWBZN1uXcKWDD8vbfKP_Uj7AXotFucCFqegxebXPw-2fkJWxA-HTQDgRFw1aXPttFtwHH5mPPfmvHK5XBv9qzGia9Jv0VKKv2Bo8RjnW-DEvEOrwFTGv1M1kyILndlDP2Gxnwd4U91fWEu-tMufaLlYRe4QUS3da7Uy88OxeEWp4jxFzqkgYAzehKrr3zxmy0)

**Expected result:** I would expect activation to be the same across all participants, and preferably (IMO) like that of participant B, the first activated in the chain.

[...]

PR #1790, opened 2024-05-23
Fixes for several teoz active life line and parallel message issues

jimnelson372:
There are a number of issues related to Teoz not displaying active life lines (I sometimes mistakenly type: lifeline) correctly (activating too late, or deactivating too late) when --++ is used (or equivalent activate and deactivate statements) or when parallel messages are used.

Issue #1494 also shows both kinds of problems, with the following example:

image

I solved the problems as seen in this diagram using my fix:

testteozlifeline_006

A forum issue 13409 showed an example with colors, and additional fixes were needed to get the colors to display correctly in these conditions.    (Mostly changes in LiveBoxes.java.)  The problem example looked like:

image

With the fixes, it looks like:

testteozlifeline_003

In recent comments in these issues, mostly on #1494, I've shared other examples.

While investigating the problems reported in #656 and #1099, which I have not yet been able to fix -- related to incorrect activation and deactivation levels when parallel message has text messages with different # of lines (either due to different number of \n or due to some messages having blank text), I found the problem in `GroupingTile.mergeParallel() `that I realized was the cause of several other reported issues with parallel messages, such as in issues #739 where multiple deactivate activate statements between parallel messages not only resulted in incorrect life lines, but also broke the parallel messages into multiple rows.  The example from #739 was:

image

While with my fix (not changing the text at all) it looks like the following.   (Please ignore the difference in the font for the participant names.  I'm not sure why the spacing between the characters is wrong in my test, but it is unrelated to my changes).  Notice that it fixes both the activating and deactivation problem that occurred, but also does keep the parallel messages on the same rows.

testteozlifeline_001

Another issue (forum 15191) had given an example of this same kind of problem, focusing on the Note position on the following part of his script:

```
activate Bob
Bob -> Alice --++: act+deact
& note right of Alice: not ok
deactivate Alice
```
This incorrectly positioned the note in the last arrow in the following diagram:

image

With the fix I made for #739, I discovered this diagram improved as well:

testteozlifeline_009

As I mentioned, I have not yet found a complete solution to issues  #656 and #1099 caused by the different text line sizes on parallel messages, but I decided I should create a pull requests for the fixes I have completed.

Here are the changes:

`LiveBoxes `   :   The --++ issue in the most of the cases was caused by the `getLevelAtInternal` method expecting the activation LifeEvent to immediately follow the current AbstractMessage in the events array, when in cases like --++ it is possible for the deactivation  event on the previous participant to be in the list before the activation event on the current participant.   So, the method needed to be modified (slightly) to loop through any number of LifeEvents to find if there is one that alters the Level on the current message.   Similarly, `getActivationColor `needed to look beyond the event for a LifeEvent that provides the color information due to the current message.

`AbstractMessage` :   It also turned out to be the case that it wasn't just enough to look for the LifeEvents whose message parameter was equal to the current message; the test needed to consider the root parallel message.   For example in the Bob, Alice, Mallory example above, when Bob sends a message to Alice and then in parallel also sends a message to Mallory, with the following script:

```
b ->> a ++ : Bye Alice!
& b ->> m --++ : Bye Alice!
```
The deactivation on b that is in the second message results in a LifeEvent object referencing that second message.  However, the first message needs to be able to find this LifeEvent to properly calculate b's changing levels at the right time.  It was because it wasn't seeing the deactivate at the right time that the lifeline on b was being deactivated later than expected in the original issue image above.   I had to create a method on AbstractMessage to perform an additional test of whether the LifeEvent's associated message was in parallel to the current message.  So I added `isParallelWith(msg)` to AbstractMessage and use it in `LiveBoxes.getLevelAtInterval`.

`GroupingTile`  :  The problem in #739 and Forum 1591 was of a similar kind to the problem in `LiveBoxes`, but was slightly more complex to fix.   It was a problem in `mergeParallel` in that the method assumed that if it encountered a tile such that `isParallel(tile)` was true, then there was only potentially one `LifeEvent` between it and the parallel message to move from the `result` list to the `pending` list.   So it only moved up to 2 events, which sometimes did not actually move the first parallel message into the new TileParallel object.  This was the cause of the strange behavior reported in those issues.  Since there could be more than 1 LifeEvent between the two parallel Message objects in the list, I modified how the method finds how many tiles to move, how to move them and how to remove them from the result list.    Once all the relevant LifeEvents were moved in the pending list correctly, the parallel messages with multiple LifeEvents started to work (as shown above).

`SequenceDiagram` :  When testing the above, I sometimes found that a LifeEvent marked as `parallel` did not have an assigned `parallelBrother`.  This turned out to be a simple fix.  The method `SequenceDiagram.getLastAbstractMessage() `which is used by `addMessage(..)` was counting down to index 1 when it needed to countdown to index 0.   If the first message in a series of parallel messages was the first in the list (e.g. no other messages or LifeEvents happening earlier) then it wouldn't be found.

I'm not sure when I can get back to issues #656 and #1099.  I've tracked it down to the the logic around `getContactPointRelative` used in the `TileParallel.drawU` method, which properly adjusts all the parallel arrow lines to the right level, but which doesn't currently do any calls to `livingSpace.addStepForLivebox(...)` to pass the adjustment so that the affected lifeline activation/deactivation levels are adjusted.   I tried adding the adjustment during the `TileParallel.callbackY_internal()`, but while it did correctly adjust the life lines directly related to an adjusted arrow line, some lifeline down stream did not adjust as I expected.   I still do not fully understand the processing involved in adjusting the lifelines.   Any hints about where it may be best to do those adjustments would be welcome. 

Regards,

Jim Nelson

### DIFF
--- a/src/net/sourceforge/plantuml/sequencediagram/AbstractMessage.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/AbstractMessage.java
@@ -111,6 +111,13 @@ public abstract class AbstractMessage extends AbstractEvent implements EventWith
 		return parallel;
 	}
 
+	public boolean isParallelWith(AbstractMessage message) {
+		boolean hasParallelBrother = parallelBrother != null;
+		return (this == message)
+				|| (hasParallelBrother && parallelBrother == message)
+				|| (hasParallelBrother && parallelBrother.isParallelWith(message));
+	}
+
 	final public Url getUrl() {
 		if (url == null)
 			for (Note n : noteOnMessages)
--- a/src/net/sourceforge/plantuml/sequencediagram/SequenceDiagram.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/SequenceDiagram.java
@@ -191,7 +191,7 @@ public class SequenceDiagram extends UmlDiagram {
 	}
 
 	private AbstractMessage getLastAbstractMessage() {
-		for (int i = events.size() - 1; i > 0; i--)
+		for (int i = events.size() - 1; i >= 0; i--)
 			if (events.get(i) instanceof AbstractMessage)
 				return (AbstractMessage) events.get(i);
 
--- a/src/net/sourceforge/plantuml/sequencediagram/teoz/GroupingTile.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/teoz/GroupingTile.java
@@ -320,17 +320,16 @@ public class GroupingTile extends AbstractTile {
 			if (result.size() > 0 && isParallel(tile)) {
 				if (pending == null) {
 					pending = new TileParallel(stringBounder, null);
-					final Tile tmp = result.get(result.size() - 1);
-					if (tmp instanceof LifeEventTile) {
-						pending.add(result.get(result.size() - 2));
-						pending.add(tmp);
-						// result.set(result.size() - 1, pending);
-						result.set(result.size() - 2, pending);
-						result.remove(result.size() - 1);
-					} else {
-						pending.add(tmp);
-						result.set(result.size() - 1, pending);
-					}
+					int capture = 1;
+					while (result.get(result.size()-capture) instanceof LifeEventTile)
+						capture++;
+
+					for (int i=result.size()-capture; i < result.size(); i++)
+						pending.add(result.get(i));
+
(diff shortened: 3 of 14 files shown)


chipbite:
Very nice, looking forward to getting these improvements! 💪

PR #1793, opened 2024-05-24
fixes for issues #656 and #1099, parallel messages with different text heights.

jimnelson372:
I mentioned at the end of my comments in pull request #1790 that I was having difficulty figuring out how to resolve the specific problem for issues #656 and #1099.

This evening I was finally able to get it working.   As I wrote , the problem was that the TileParallel would properly adjust the positions of the message arrows so that all parallel messages aligned during its drawU(..) method, but it was not also calling any addStepForLiveBox(...) methods so that the lifeline activation levels would also adjust.

My changes were:

`TileParallel` : While `TileParallel `doesn't directly have access to any living spaces, I added to its `callbackY_internal `method, doing the same calculation as is done it drawU to determine how much of an adjustment needs to be made for a given `CommuncationTile` in the list of tiles.   Not being able to directly call an `addStepForLiveBox(...) `method, I create a recalculated `TimeHook` when appropriate to then pass on to the tile.   I also added one additional method, `isParallelWith` that I'll explain below, which allowed me to test if a `CommunicationTile` is parallel to the other tiles already in the current `TileParallel`.   This internally uses the method I added earlier to test their messages.

`GroupingTile` : I discovered that I had to once again enhance the `mergeParallel` method in this class.   In order to solve this multi-height parallel message problem completely, each grouping of parallel messages must all be included in the same TileParallel.   That way, if the 3rd or later message is the one with the extra tall text, the earlier messages do adjust.   In the earlier versions of this method, if there were any LifeEvents between the 2nd and 3rd (or 3rd and 4th, etc.) parallel messages, it would end up grouping the first 2, then grouping that TileParallel with the 3rd, and so on.   I reorganized that method again to simplify how it collects the groups of parallel messages.   Part of the logic in this new version required the ability to test if the current tile (being processed) belongs to the current pending `TileParallel` group, or if it's time to set pending to null.

I've also added several new nonregression tests for the issue examples, plus one of my own.  All my prior nonregression tests were unchanged by these fixes.

Here are the results of these changes:

Issue 656 has an example with the following script:

```
A -> B++: Get data
& A -> C++: 
& A -> D++: 
B -->> A--: Data
& C -->> A--: 
& D -->> A--:
```

Which currently generates:

image

Where the lifeline for C & D begins too early and ends too early due to the problems I mentioned earlier.  Now it will generate:

testteozlifeline

This is a nice example of multiple groups of parallel messages.  The sending messages are grouped and the return messages are grouped.   The messages on C and D have blank test, which is what caused the adjustment that confused the timeline levels.   I also tested what would happen with the larger height messages occurring at different points in the diagram, such as between B and C or between C and D.   I tested a wide variety of conditions to make sure it all works.  Here's one example:

```
@startuml
!pragma teoz true
A -> B++: Get data
& B -> C++:a\n2
& C -> D++:
D -->> C--:a\n2
& C -->> A--:
&B -->> A--: Data
@enduml
```

This currently generates:

image

But with the fix it will now generate:

testteozlifeline_014

Finally, the example from issue #1099 is nice to see.  Here is is the script:

```
@startuml
!pragma teoz true
  A <[#red]-o? ++
  A -> B -- : very long\nmulti-line
& B -> C ++ #red: single
note over C: got it
Deactivate C
====
  A <-o? ++ #green
  A -> B -- : very long\nmulti-line
& B -> C ++ #green : same\nheight
note over C: got it
Deactivate C
====
  A <-o? ++ #red
  A -> B -- : single
& B -> C ++ : very long\nmulti-line
note over C: got it
Deactivate C
@enduml
```

It currently generates the following (even with my fixes from yesterday, since all the problems related to the text height differences):

testteozlifeline_005

It will now generate:

testteozlifeline_005

Let me know if you have any questions or concerns.

Regards,

Jim Nelson

### DIFF
--- a/src/net/sourceforge/plantuml/sequencediagram/teoz/GroupingTile.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/teoz/GroupingTile.java
@@ -317,27 +317,41 @@ public class GroupingTile extends AbstractTile {
 		tiles = removeEmptyCloseToParallel(tiles);
 		final List<Tile> result = new ArrayList<>();
 		for (Tile tile : tiles) {
-			if (result.size() > 0 && isParallel(tile)) {
-				if (pending == null) {
-					pending = new TileParallel(stringBounder, null);
-					int capture = 1;
-					while (result.get(result.size()-capture) instanceof LifeEventTile)
-						capture++;
-
-					for (int i=result.size()-capture; i < result.size(); i++)
-						pending.add(result.get(i));
-
-					for (int i=1; i<=capture; i++)
-						result.remove(result.size()-1);
-					result.add(pending);
-				}
+			if (!isParallel(tile) || result.size() == 0) {
+				result.add(tile);
+				if (tile instanceof LifeEventTile == false)
+					pending = null;
+			} else if (pending == null) {
+				pending = new TileParallel(stringBounder, null);
+				moveRecentParallelTilesToPending(result, pending);
+				pending.add(tile);
+				result.add(pending);
+			} else if (pending.isParallelWith(tile)){
+				moveRecentParallelTilesToPending(result, pending);
 				pending.add(tile);
 			} else {
-				result.add(tile);
 				pending = null;
+				result.add(tile);
 			}
 		}
 		return result;
+	}
+
+	private static void moveRecentParallelTilesToPending(List<Tile> result, TileParallel pending) {
+		if (result.size() == 0) return;
+
+		int capture = 1;
+		while (result.get(result.size() - capture) instanceof LifeEventTile)
+			capture++;
+
+		if (result.get(result.size()-capture) == pending)
+			capture--;
+
+		for (int i = result.size() - capture; i < result.size(); i++)
+			pending.add(result.get(i));
+
+		for (int i = 1; i <= capture; i++)
+			result.remove(result.size() - 1);
 	}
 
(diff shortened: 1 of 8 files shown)


jimnelson372:
I meant to demonstrate that point about what happens if I hadn't updated `mergeParallel` as I did.   Consider the case where there are only two parallel messages and it's the 2nd message that has the taller text, such as in this script:

```
@startuml
!pragma teoz true
A -> B++: Get data
& B -> C++:a\n2
@enduml
```

All now works fine and it will generate the following, even without the `GroupTiling.mergeParallel` fix I made, but just due to the `TileParallel.callbackY_internal change`:

testteoz

But if we have more than 2 parallel messages, such as a C -> D, and make it the tallest, such as in:

```
@startuml
!pragma teoz true
'Issue #656
A -> B++: Get data
& B -> C++:a\n2
& C -> D++: a\n2\n3
@enduml
```
 We now see something resembling the original problem, with the activation of B and C starting too early:
 
testteoz


And this is mainly due to the fact that the D->C event was not getting included in the TileParallel for the 3 parallel messages.

Though, in the relased version, this would work so long as we didn't have the activation on C that ends up in the list of Events between C and D.  Just removing that middle ++, we'd get:

testteoz

However, with the activation of C, the new version of `GroupTiling.mergeParallel` is needed to get it work as expected.   Using the new GroupTiling, and adding back in the activation on C, the result becomes:

testteoz

Issue #591, opened 2021-07-12
`alt`/`else` block shown in parallel using teoz leads to wrong formatting


mxmlnkn:
The alt/else block can't be rendered in parallel by itself, see also #269 but I accidentally got it somewhat rendered in parallel by having the alt/else block in another group. But, then, the header and separation line for the `else` block move to the top instead of being in the middle where it should be:

```plantuml
@startuml
!pragma teoz true

!theme crt-amber
skinparam backgroundColor #000000

participant "Random Name" as foo

hide footbox

foo -> foo : test

& opt message received
    alt REQUEST
        bossrpcp -> bossrpcp : request
    else RESPONSE
        bossrpcp -> bossrpcp : respond
    end
end

@enduml
```

Rendering for the source as shown:

![teoz-alt-else-bug-parallel](https://user-images.githubusercontent.com/6842824/125301519-5ab76780-e32b-11eb-96da-edec62bc0c15.png)

Rendering without the `&`:

![teoz-alt-else-bug-not-parallel](https://user-images.githubusercontent.com/6842824/125301405-407d8980-e32b-11eb-965a-e05ae62edc90.png)

jimnelson372:
Hello @mxmlnkn and @The-Lum, I believe I have found the cause of this issues and I have implemented a solution.   Here is what my fixed version looks like.  (This is the exact script above, but commenting out the theme and backgroundColor instructions, as I only have the default theme in my test environment):

![testteozaltparallel](https://github.com/plantuml/plantuml/assets/66284321/be2d463f-8084-431a-8e72-4f763c9255c8)

I've tested that this fix works regardless of the order of the two parallel elements; in other words, `foo -> foo: test` can be moved to after the opt group and it still works.   My fix will work when an alt/else block is contained within an Opt group.  But there is an additional problem shown in issue #269 when trying to use an alt/else block directly as a parallel element.  I'm looking at that now.

By the way, you can see in the diagram I shared that the upper line of the self message is a little higher than the top of the Opt group.   There is another issue (#1791) where I noticed the  top lines of self messages were not matching up with other messages.   I will see if that can be adjusted, but it may need to wait on later fixes.

Regards,

Jim Nelson

jimnelson372:
Hello @mxmlnkn and @The-Lum, I was able to find a solution for #269 as well.  I wrote a comment on that issue demonstrating it.  Let me know if anything still looks out of place with it.  Thanks.

The-Lum:
Hi all,

[...]

Issue #269, opened 2019-10-24
groups (e.g. `alt`) with `else` cannot be rendered in parallel


da-kami:
This renders correctly:
```plantuml
@startuml
!pragma teoz true
    opt
        A->B: message
    end
    & opt
        C->D: message
    end

    alt
        A->B: message
    end
    & alt
        C->D: message
    end

    alt
        A->B: message
    else default
        B->B: other
    end
    alt
        C->D: message
    else default
        D->D: other
    end

@enduml
```

![image](https://user-images.githubusercontent.com/5557790/67444659-bd4b6a00-f655-11e9-8c4f-cc410ec65878.png)

Once the `alt` with `else` is parallelised it throws a `NullPointer` exception. 
```plantuml
@startuml
!pragma teoz true
    opt
        A->B: message
    end
    & opt
        C->D: message
    end

    alt
        A->B: message
    end
    & alt
        C->D: message
    end

    alt
        A->B: message
    else default
        B->B: other
    end
    & alt
        C->D: message
    else default
        D->D: other
    end

@enduml
```

![image](https://user-images.githubusercontent.com/5557790/67444692-e409a080-f655-11e9-8dff-ad4af43c2233.png)

jimnelson372:
@da-kami and @The-Lum, 

I was able to track down the cause of the exception and find a fix.   It required the changes I made for #591 to properly position the else header line and message, but it works well.   

First, here is what my test environment generates now when running your exact script:

![testteozaltparallel_005](https://github.com/plantuml/plantuml/assets/66284321/2c9438fe-8804-40a8-b88a-120e39136271)

The alt/else blocks are now in parallel, but obviously there is overlap on the two groups.  That's because the internal sides of each group do not push push away other lifelines or events, but merely surround the events they contain.  So in this case the `Alt` groups overlap.  Even the gap in your original parallel `Opt` and `Alt` blocks only were separated due to the needed width of `B->B : other` self message further down in your script.   (I'm somewhat new to PlantUML and the community so I don't know if there are discussions about groups altering (or not altering) the layout.  Though without the parallel (&) feature in Teoz, this wouldn't have been an issue.)

To remove the overlap, I added a hidden message between B and C with sufficient width:

`    B -[hidden]-> C: "              "`

This results in:

![testteozaltparallel_005](https://github.com/plantuml/plantuml/assets/66284321/bcb444fb-b5d1-4c10-99b9-77d40eca169d)

[...]

PR #1802, opened 2024-05-31
Fixes for #269 and #591; alt else problems when used in parallel

jimnelson372:
Issue #269 gave an example where an alt...else...end block was directly used in parallel with another alt..else...end block and resulted in an exception being generated.   Issue #591 presented an example where an the alt..else..end block was contained in an Opt group which itself was in parallel with a self message.  In #591, the script generated a diagram, but the else line was drawn at the top of the alt group rather than near the middle to separate the two sections.

The layout problem was primarily in the `GroupTile` class, in the `fillPositionelTiles` method.   While that method recursively processed GroupTiles, it needed to also process any `GroupTile` located within `TileParallel` tiles.   I separated the common code needed for the `GroupTile` recursion and the TileParallel recursion into methods `fillPositionalSubGroupTiles` and `fillPositionalParallelTiles`.

The cause of the exception in #269 was primarily due to Puma code needed for the Par2..Else...End block, where a Par2 GroupStart is marked as parallel which also made its GroupLeaf (else clause) indicated itself as parallel.  This confused the `GroupingTile.mergeParallel(...)` method, primarily in that the GroupingLeaf (Else Tile) was indicating it was parallel but logically should not have been within the Teoz layout system.  This problem could be found both the prior versions of `mergeParallel` and in the updated version in the latest release.  To fix this issue required:

- Ensuring that the `GroupLeaf.isParallel()` method to ONLY return true for when it's start is a Par2 group.   When not using Par2.  This prevented Alt group ElseTiles from indicating they are parallel just because the Alt group itself is parallel.
- I had looked into fixing the current Par2 problem in Teoz, where it currently throws and exception (for the reasons given above), but since I did not find a complete (yet not too complicated) solution, I decided that can wait for another time.   The changes I did include were needed to allow Teoz to handle Alt groups layout correctly, while also keeping the Par2 code needed by Puma.

Finally, I discovered that part of the change I made to `mergeParallel` in pull request #1793 caused a problem if a Group of any kind was in parallel with more than 1 other message or group.  Specifically, it turned out that the `TileParallel.isParallelWith(tile)` method that I created and used was not actually needed in the final logic I implemented in the `mergeParallel` method.  It had been needed only by earlier versions.   So I removed that test from the last `else if` test making it the final `else`.  The condition to set the `pending = null` the first `if` condition was sufficient.   With this change, groups can be part of a set of more than 2 parallel messages without problem.

I presented the images of the fixed diagrams in the two issue reports today.   But to give an example here, for #591 the layout of the alt...else...end block when put into parallel resulted in the following diagram:

testteozaltparallel_011

With this fix, it looks like:

testteozaltparallel_011

For #269 which resulted in an exception, it wasn't possible to see the layout before the fix, but here is how the diagram looked if the & was removed from between the two alt/else blocks at the bottom:

testteozaltparallel_012

With the fix, the & could be added back in and the result was the following.

testteozaltparallel_012

Since the group's themselves are not creating any pressure on the inner sides, this results in some overlap.   However, it was easy enough to add a hidden message between B and C to push them further apart.  The result is:

testteozaltparallel_012

But please let me know if you want further examples or have any questions.   I tested a variety of adjustments, such as more than one else clause, different orderings of the parallel messages/groups, and more.    

All the prior diagrams I have my own testing were unaffected.  The existing nonreg tests did not change.   I added new nonreg tests for these changes.

Regards,
Jim Nelson
jimnelson372

### DIFF
--- a/src/net/sourceforge/plantuml/sequencediagram/GroupingLeaf.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/GroupingLeaf.java
@@ -98,7 +98,8 @@ final public class GroupingLeaf extends Grouping implements EventWithDeactivate,
 
 	@Override
 	public boolean isParallel() {
-		return start.isParallel();
+		// only Par2 Group Leafs should return true.
+		return start.isPar2GroupStart();
 	}
 
 	private double posYendLevel;
--- a/src/net/sourceforge/plantuml/sequencediagram/GroupingStart.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/GroupingStart.java
@@ -102,11 +102,14 @@ public class GroupingStart extends Grouping {
 
 	@Override
 	public boolean isParallel() {
-		return parallel || getTitle().equals("par2");
+		return parallel || isPar2GroupStart();
 	}
 
 	public void goParallel() {
 		this.parallel = true;
 	}
 
+	public boolean isPar2GroupStart() {
+		return getTitle().equals("par2");
+	}
 }
--- a/src/net/sourceforge/plantuml/sequencediagram/teoz/GroupingTile.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/teoz/GroupingTile.java
@@ -294,15 +294,33 @@ public class GroupingTile extends AbstractTile {
 			full.add((CommonTile) tile);
 			if (tile instanceof GroupingTile) {
 				final GroupingTile groupingTile = (GroupingTile) tile;
-				final double headerHeight = groupingTile.getHeaderHeight(stringBounder);
-				final ArrayList<CommonTile> local2 = new ArrayList<>();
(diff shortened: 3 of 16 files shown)


jimnelson372:
Hi @arnaudroques,

I just wanted to let you know that I used the pdiff tool to compare a jar file generated with the changes in this pull request with the jar file for the master branch (plantuml-1.2024.6beta1.jar) and there were no unexpected differences.   

The changes in this pull request do correct diagram **bn5me** in the existing diagrams in pdiff, as expected.  That diagram had broken timeline activation levels caused by using & with a group (a loop group).   Since that group ends up inside a TileParallel, the fix in the pull request was needed to make sure the tiles inside that group were properly processed for layout.

Let me know if you have any questions.

Regards,

Jim N
@jimnelson372

PR #1803, opened 2024-05-31
fix: minor regression on space instruction before deativation from #1790 changes

jimnelson372:
A small regression issue came to my attention yesterday related to the use of the Space instruction, e.g. ||0||, used after a self message before a delay instruction to ensure the active level remains active until after the end of the arrow.   A simple example is the script:

```
@startuml
!pragma teoz true
activate a
a -> a  : call
||0||
deactivate a
@enduml
```

Which up to release 1.2024.4 would generate:

testdeactivatedelay

When I made the fix to `LiveBoxes.getLevelAtInternal(...)` to resolve the lifeline active level problems resulting from `--++` and also from `LifeEvents` on parallel messages, I needed to make it look further ahead than just the next `LifeEvent` to find all the `LifeEvent`s associated with the current message being evaluated.   However, this causes the search for `LifeEvent`s to skip right over the `||o||` instruction to consider the `deactivate a` instruction first.  This results in that same script behaving as through the `||o|| `wasn't included:

testdeactivatedelay

Looking at alternative ways to solve this regression, I found that the simplest was just to have that method in `LiveBoxes` stop looking for more `LifeEvent`s on the current message when it encounters anything other than a `LifeEvent` or an `AbstractMessage`, which are the only two Event types it ever processes for its calculation.   Now it will exit its search when it encounters the Space instruction, will continue searching when it encounters other `AbstractMessage`s, and will check on any `LifeEvent`s it encounters as needed for the #1790 fixes to work.  You'll see it's just a one line addition to the while loop I created:

 ` if (!(next instanceof LifeEvent || next instanceof AbstractMessage)) break;`

All my test diagrams still work as expected.  All the current nonreg tests I have still work, and the above script now correctly generates:

testdeactivatedelay

And, if the `||o||` comes _after_ the `deactive a`, or a `--` is on the message, the deactivate will occur before that space instruction is applied (most likely with values more than 0 in such cases), as expected, looking very much like the 2nd diagram above.

Let me know if you need any further information.  **NOTE: This pull request is completely independent of file changes in other recent pull request, so can be merged independently.**

One last note, as I discussed with @The-Lum yesterday, there are some questions about how activation levels should work on self messages in general, if perhaps the arrow head type (async or not) should determine whether the activation level extends to the target level or stop at the source level, like is the case for sloped messages.   That is still an unresolved issue, so I wanted to get this simple fix to you quickly.   It makes the `LiveBoxes` method correct regardless of later decisions about self messages.

Regards,

Jim Nelson
@jimnelson372

### DIFF
--- a/src/net/sourceforge/plantuml/sequencediagram/teoz/LiveBoxes.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/teoz/LiveBoxes.java
@@ -109,6 +109,7 @@ public class LiveBoxes {
 				if (current instanceof AbstractMessage) {
 					while (it.hasNext()) {
 						final Event next = nextButSkippingNotes(it);
+						if (!(next instanceof LifeEvent || next instanceof AbstractMessage)) break;
 						if (!(next instanceof LifeEvent)) continue;
 
 						final LifeEvent le = (LifeEvent) next;


arnaudroques:
Non-regression tests can be challenging. To assist users in running extensive non-regression tests on PlantUML, we have published a utility that compares the results of two different versions of PlantUML: [pdiff](https://github.com/plantuml/pdiff/).

While we are currently expanding the diagrams collection, we believe it is already substantial enough to be useful.

We hope this helps!

jimnelson372:
Hi @arnaudroques, I'm testing out the use of pdiff this morning, trying to test the changes in my recent alt/else pull request (that you haven't merged) to see if this reveals a problem with that change.   A few observations:

- I had not been using ELK, so I was getting errors from not having that jar file in place, but I worked through that.
- Some of the diagrams show differences merely because they are diagrams showing current time, or showing the PlantUML version, creating an inherent difference between the two version runs.   Those diagrams should probably be modified not to display these non-essential differences.
- I am getting errors with diagrams needing graphviz.  I'm on an iMac, so will check my installed version.   
- Can there be some way to compare two current branches?  I was trying to compare the current master branch with my modified branch.  Even though I renamed the jars so I could run the pdiff on each one...but they both generate the same 1.2024.6beta1.html file (since it is using their internal version information).   I will try to change the coded version before I generate the jars, but it seems like we should be able to specify the naming used by pdiff rather than it using the coded version.

Nice tool though!  Thanks.

Jim Nelson
@jimnelson372

arnaudroques:
> * Some of the diagrams show differences merely because they are diagrams showing current time, or showing the PlantUML version, creating an inherent difference between the two version runs.   Those diagrams should probably be modified not to display these non-essential differences.

Could you point out those diagrams? I am supposed to have removed them from my push :-)



> * Can there be some way to compare two current branches?  I was trying to compare the current master branch with my modified branch.  Even though I renamed the jars so I could run the pdiff on each one...but they both generate the same 1.2024.6beta1.html file (since it is using their internal version information).   I will try to change the coded version before I generate the jars, but it seems like we should be able to specify the naming used by pdiff rather than it using the coded version.

When running pdiff in "Run" phase, you should be able to use the "-r" option to override the name of the run and put any name you'd like to. However, this has not be really tested :-)

```
java -cp plantuml-xxx.jar:build/libs/pdiff-all.jar com.pdiff.Main run -r branch1
```


> * I am getting errors with diagrams needing graphviz.  I'm on an iMac, so will check my installed version.

I may add an option to use the internal Smetana layout engine (which is not finished but which is a possible replacement for graphviz).

arnaudroques:
> * I had not been using ELK, so I was getting errors from not having that jar file in place, but I worked through that.

I realize that since you are modifying code specifically for Sequence Diagrams, there is no chance that your changes will affect other diagrams (such as class or component diagrams) which use ELK, GraphViz, or Smetana for layout.

Therefore, I should probably add an option to select only certain diagrams (Sequence Diagrams, in your case) when using pdiff.

jimnelson372:
> Could you point out those diagrams? I am supposed to have removed them from my push :-)

Yes.  They are:

- 798qj    A Gantt diagram, title
- cltdq     Test of non-existing-font...version in the footer
- de05f    An activity diagram, title
- e1a67    Saying diagram not supported, showing version info.
- gu87f    A class diagram?   The final class names seem to be the specific version (e.g. 5 vs 6beat1)

It turns out I'm not actually having problems with Graphviz as such, as I do have the correct version, proved by testdot, and I can generate graphviz based documents generally.  

What I'm seeing, though, is exceptions in my alt/else changes, the master branch (i.e. 1.2024.6beta1) and in 1.2024.5.  So, it's not the result of my alt/else change. These exceptions show up because in the exception report, the PlantUML version is different.  I haven't looked at versions further back.  These exceptions are on the following diagrams:

- 5afd9
- biz8z
- d3yum

I'm not sure if I should just ignore these exceptions, when they are in the prior version as well.   I will try to test 1.2024.4, but when I just did, it overwrote the 1.2024.4.html file in the pdiff directory and caused the diff command to exit with exceptions.   I had to revert it back to continue.

Jim N.

jimnelson372:
Oh, the ones that generate exceptions/stack dumps on their diagrams don't only show up because the version may be different, but there seems to be randomly changed text message just after the line that says an exception has occurred.  Since the quote is different in each case, it shows a difference in the comparison.  

Jim

PR #1816, opened 2024-06-09
Fix parallel group active levels

jimnelson372:
My changes for pull request #1802 did fix the reported issues with parallel alt-else blocks.  However, while working on another issue related to activation levels, I discovered that my code changes were correcting positions for everything in the parallel groups except for the activation levels.   Fortunately, the solution ended up simplifying the code overall.

It turns out the changes I had made to `TileParallel.callbackY_internal(...)` that were needed to solve the original issues related to --++ and activity levels were better performed in the `GroupingTile.fillPositionalParallelTiles(...) `method I had created to fix the alt-else parallel problems.  When I wrote the change to `TileParallel.callbackY_internal`, that new `GroupingTile` method didn't exist, and when I wrote the `fillPositionalParallelTiles(...)` method, putting all the adjustments in it was conflicting with the changes I had in `TileParallel`.   Well, to get it all working together, all the adjustments needed to be in the same place.  So, now that the `fillPositionalParallelTiles(..)` method did exist, I moved all the parallel adjustment logic there.

What did this fix.   Here is an example of a Group that is NOT in parallel with another message, to see how the activation levels inside the group should look:

testteozgroupoldparallel

Now, if I made the group parallel with the message (adding & in front of it), with the following script:

```
@startuml
!pragma teoz true
participant a
participant b
participant c
b ->> c ++: Outside Group
&opt test
  activate a
  a ->> b --++ : Inside Group
end
deactivate a
deactivate c
deactivate b
@enduml
```

The activation levels in the group generated the wrong image, where the activation levels had not adjusted.

testteozgroupoldparallel_001

With the new fix, everything else still works as expected (I also tested with pdiff),  but now the activation levels within the group draw correctly:

testteozgroupparallel_001

I was initially planning to include this change with other work I am doing related to activity levels, but I decided it could be merged on its own, to keep the change simpler.   I was very happy with this fix, since you may recall when I first made the changes to `TileParallel.callbackY_internal(...)` I had mentioned that it seemed to be an awkward place to do the adjustments I needed, but I didn't have a better place to put them.   I finally did.  

Please let me know if you have questions.

Regards,

Jim Nelson
@jimnelson372

### DIFF
--- a/src/net/sourceforge/plantuml/sequencediagram/teoz/GroupingTile.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/teoz/GroupingTile.java
@@ -314,12 +314,18 @@ public class GroupingTile extends AbstractTile {
 	}
 
 	private static void fillPositionalParallelTiles(StringBounder stringBounder, TimeHook yArg, List<CommonTile> full, TileParallel tileParallel) {
+		final double yPointAll = tileParallel.getContactPointRelative();
 		for (Tile tile : tileParallel.getTiles()) {
-			if (tile instanceof GroupingTile) {
-				GroupingTile groupingTile = (GroupingTile) tile;
-				final double headerHeight = groupingTile.getHeaderHeight(stringBounder);
-				fillPositionalSubGroupTiles(stringBounder, new TimeHook(yArg.getValue()), full, groupingTile);
-			}
+
+			final double yPoint = tile.getContactPointRelative();
+			final double adjustment = yPointAll - yPoint;
+			TimeHook yAdjusted = new TimeHook(yArg.getValue() + adjustment);
+
+			tile.callbackY(yAdjusted);
+			full.add((CommonTile) tile);
+
+			if (tile instanceof GroupingTile)
+				fillPositionalSubGroupTiles(stringBounder, yAdjusted, full, (GroupingTile) tile);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


jimnelson372:
I wanted to point out that if both Tiles were Groups, prior to this fix, and were the same height, then this problem was not apparent as the activation levels seemed to be where they should be.  It was only when I was testing this condition of a group parallel with a non-group that I saw the missing adjustment.

arnaudroques:
Thanks for your contribution!

Should we add this diagram to pdiff?

```
@startuml
!pragma teoz true
participant a
participant b
participant c
b ->> c ++: Outside Group
&opt test
  activate a
  a ->> b --++ : Inside Group
end
deactivate a
deactivate c
deactivate b
@enduml
```

arnaudroques:
> Fortunately, the solution ended up simplifying the code overall

Yes, sometimes it's magic!


> Please let me know if you have questions.

No, that's just great!
Thanks again

jimnelson372:
`Should we add this diagram to pdiff?`

Yes, please do add it.

The-Lum:
FYI
It is done by:
- [3bp8voiq](https://github.com/plantuml/pdiff/blob/05802ae0f539993e883c7aca5755f915b6ec518b/db/3b/3bp8voiqzjrc6q79fyee8fmsjpgy9od.puml)

On:
- https://github.com/plantuml/pdiff/commit/05802ae0f539993e883c7aca5755f915b6ec518b

Thanks @arnaudroques.

Issue #1702, opened 2024-02-29
Smetana is not supporting colors in UML


martinviszlai:
When using `!pragma layout smetana`, relation tags formatting is ignored.

Example of a simple diagram:
```plantuml
@startuml
!include https://raw.githubusercontent.com/plantuml-stdlib/C4-PlantUML/master/C4_Container.puml

AddRelTag("API", $lineColor="red", $textColor="red", $lineStyle="dashed")

System(service, "service", "My awesome service")
System(client, "client", "Client service")

Rel(client, service, "make API calls", "HTTP", "calls methods from interfaces", $tags="API")
@enduml
```

kirchsth:
with **elk layout** the colors,... are ignored too. Maybe it is the same skinparam related problem 

https://forum.plantuml.net/18723/elk-layout-ignores-arrow-skinparam

https://github.com/plantuml-stdlib/C4-PlantUML/issues/340

[![](https://www.plantuml.com/plantuml/svg/bP3DJeH048Jl_HJbUb0JyG5qGtQ1T_1WF7HaWeJvsJHDoCRuxWwpekFFPM-LhleefdQZikZaB5s1HSACSUZzWOKTvY33YE3j4NElhs5I63l0yZ7AcByu27UEJuVaqQ_Jw8ooPwBlZwQIxLPCsmHLqvObtWW0wc23u38Q3xRtvgvDDht_3__5KbJCIopihC4-qrMcwqmtH7iKJmjyUwfxBafKkCVcyVF1NzqRgb14jyQFdr-psMpNorXfAHzq6J4wtStADua_ZzmPtqxEVW00)](https://www.plantuml.com/plantuml/uml/bP3DJeH048Jl_HJbUb0JyG5qGtQ1T_1WF7HaWeJvsJHDoCRuxWwpekFFPM-LhleefdQZikZaB5s1HSACSUZzWOKTvY33YE3j4NElhs5I63l0yZ7AcByu27UEJuVaqQ_Jw8ooPwBlZwQIxLPCsmHLqvObtWW0wc23u38Q3xRtvgvDDht_3__5KbJCIopihC4-qrMcwqmtH7iKJmjyUwfxBafKkCVcyVF1NzqRgb14jyQFdr-psMpNorXfAHzq6J4wtStADua_ZzmPtqxEVW00)

Thank you and BR
Helmut

The-Lum:
Hello @arnaudroques, and all contributors,

Here is a minimal not working example:

[...]

Issue #1230, opened 2022-12-22
Smetana: Allow full style on label of arrows


The-Lum:
Hello PlantUML team, @arnaudroques, @matthew16550, and all,

---

**Observation:**
It seems that label of arrows is always black with Smetana layout.

Example observed with `amiga` theme:
```puml
@startuml
!pragma layout smetana
!theme amiga

class A
class B

A -> B : that A2B
@enduml
```
>[![](https://www.planttext.com/api/plantuml/svg/BSmn3e0W40J0_bwnFi36qi80Fzae0HCE3Ho5lzV2QhfntTXiQ97fRanA53xNC7IDnagPB4UDe9w98djXx_2_GSHZtX2mmZ8DVWdYOZs-xmK0)](https://www.planttext.com/?text=BSmn3e0W40J0_bwnFi36qi80Fzae0HCE3Ho5lzV2QhfntTXiQ97fRanA53xNC7IDnagPB4UDe9w98djXx_2_GSHZtX2mmZ8DVWdYOZs-xmK0)

---

Then here is a **minimal example**:
```puml
@startuml
!pragma layout smetana

<style>
root {
  BackgroundColor gray
  Linecolor white
  
  FontColor white
}
</style>

class A
class B

A -> B : that A2B
@enduml
```
>[![](https://www.planttext.com/api/plantuml/svg/JSv12i8m40NG_PnYF808BgMK9WLNNc98GrfCCZAPK89uTuDq8Npuy1QVFnKbqPeYd5v28H56QbmLI_9AcG26eYtw4OHPyGs8bjmp2DUyp1nPC0YrpeyjU_U3VTtKT-cvSzRv3pymN8uzS95AGNEq1J1u7j7Y3NKbHNErCFcyz6zV)](https://www.planttext.com/?text=JSv12i8m40NG_PnYF808BgMK9WLNNc98GrfCCZAPK89uTuDq8Npuy1QVFnKbqPeYd5v28H56QbmLI_9AcG26eYtw4OHPyGs8bjmp2DUyp1nPC0YrpeyjU_U3VTtKT-cvSzRv3pymN8uzS95AGNEq1J1u7j7Y3NKbHNErCFcyz6zV)

---

_[Pseudo-similar question/issue or related issue/PR on smetana: #448, #738]_
Thanks for the future enhancement,
Regards,

The-Lum:
[...]

PR #1833, opened 2024-06-22
Elk: Allow full style on arrows (except fontcolor)

kirchsth:
Hello PlantUML team, @arnaudroques, @The-Lum 

related to #1702, [forum elk/smetana layout ignores arrow skinparam](https://forum.plantuml.net/18723/elk-layout-ignores-arrow-skinparam), [C4-PlantUML Line styling not recognized when using ELK](https://github.com/plantuml-stdlib/C4-PlantUML/issues/340) 

**Observation:**
It seems that stereotype/style specific linecolor, linethickness and linestyles of arrows is always ignored with Elk layout.
E.g. 
```plantuml
@startuml
'!pragma layout smetana
!pragma layout elk

<style>
.a {
  LineThickness 3
  Linecolor red
  Linestyle 3-2

  Fontcolor green
}
</style>

card a
card b
a -> b <<a>> : Should be green
@enduml
```

produces without the fix following output
image

After the fix the output is
image

(The fontcolor of the arrow label is not fixed with this PR, but I didn't found a solution for that; #1230 has the same problem)

My [1702_elk_arrow_stylesheet_incl_test branch](https://github.com/kirchsth/plantuml/tree/fix/1702_elk_arrow_stylesheet_incl_test) has a more complex test [Test_1702_ColorElk](https://github.com/kirchsth/plantuml/blob/fix/1702_elk_arrow_stylesheet_incl_test/test/test/example/Test_1702_ColorElk.java)
with following output (arrow label should be green, but this is not part of the fix)
image
but for this test I added `libs\elk-full.jar` therefore I didn't add it to the pull request.

BR Helmut

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ out
 /bin
 #vscode files
 **/bin/
+/.vs
 
 # Ant result file
 plantuml.jar
--- a/src/net/sourceforge/plantuml/elk/ElkPath.java
+++ b/src/net/sourceforge/plantuml/elk/ElkPath.java
@@ -5,12 +5,12 @@
  * (C) Copyright 2009-2024, Arnaud Roques
  *
  * Project Info:  https://plantuml.com
- * 
+ *
  * If you like this project or if you find it useful, you can support us at:
- * 
+ *
  * https://plantuml.com/patreon (only 1$ per month!)
  * https://plantuml.com/paypal
- * 
+ *
  * This file is part of PlantUML.
  *
  * PlantUML is free software; you can redistribute it and/or modify it
@@ -30,7 +30,7 @@
  *
  *
  * Original Author:  Arnaud Roques
- * 
+ *
  *
  */
 package net.sourceforge.plantuml.elk;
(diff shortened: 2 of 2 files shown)


arnaudroques:

I have included your test in the new [pdiff project](https://github.com/plantuml/pdiff/commit/def0621dd18e7c81261dab9556c1b91ce88566b6), which we are now using for non-regression tests.

I understand that these tests require ELK to run, and I am considering integrating ELK into the main PlantUML library by default. This would mean that, in the future, these tests could work out-of-the-box without the need to download the ELK JAR file separately.

The-Lum:
Hi @kirchsth,

For traceability, I just open an issue about `fontcolor`...
- https://github.com/plantuml/plantuml/issues/1834

Regards,
Th.

StephanPirnbaum:
> Thank you very much for your contribution.
> 
> I have included your test in the new [pdiff project](https://github.com/plantuml/pdiff/commit/def0621dd18e7c81261dab9556c1b91ce88566b6), which we are now using for non-regression tests.
> 
> I understand that these tests require ELK to run, and I am considering integrating ELK into the main PlantUML library by default. This would mean that, in the future, these tests could work out-of-the-box without the need to download the ELK JAR file separately.

Hi @arnaudroques 
integrating ELK into the default distribution of PlantUML would be really great. For most of the customer projects, I'm using ELK by default as GraphViz brings challenges for the installation especially in corporate CI/CD environments and ELK often gives more appealing results. Having ELK in the standard distribution would make the setup much easier, especially for the IDE tooling.

Issue #1097, opened 2022-08-26
inconsistent computation of group/alt/opt box size given left or right arrows


jhoehle:
The computed size of alt/opt boxes differs on whether left or right arrows are used, as seen from the following examples.
```puml
@startuml
!pragma teoz true
skinparam ParticipantPadding 18
skinparam BoxPadding 20
box
participant A
end box
  A -> B
& B -> C
opt#transparent #pink bad /'in teoz'/
?o-> B : Pre
& B <-o? : yolo
'& C -[hidden]> C /' enlarge opt box '/
end
opt#transparent expand?
[o-> B : Pre
& B <-o? : yolo
'& C -[hidden]> C /' enlarge opt box '/
'& C -[hidden]> B
end
opt#transparent #20DD0090 OK
  A o-> B
& B <-o? : yolo
'& C -[hidden]> C /' enlarge opt box '/
end
opt#transparent #pink bad?
  A o-> B
& B <-o] : yolo
'& C -[hidden]> C /' enlarge opt box '/
end
@enduml
```
![Server view](https://www.plantuml.com/plantuml/png/fP8nJyCm48Nt-nKFHMgcgAOJ84JQj1i3sQiELsm5gubjEQuK-FLS8Y6G0ajOelZytlkUJjxsYMEwTYtTXCXDnqZ6Vo35gw7-ObtWo1rgqTXN6zYbchMshi7jtO_xoWzVyusYinyeV3km8-CqnYcmG_64YbQenfyz-P2o5DdrOZ2YpO9r5vnPOvrRDvN9rrJwqOK7rD5CtiV2bt9yzwsdV8KzYkERrTguauIA5SQr71i3oH_9a10fCSEPGHhgaeuJO6dqJ5Vz2iiswd1GwbxXvNbQnV-WWlXxVUKSS5e2sCf7NiSd)
Behaviour should be consistent. Summary of findings:

Teoz engine:
 - Short `?o-> B` is handled like the long arrow `[o->` causing _surprisingly large boxes_.

Standard engine:
 - The o (of o-> or ?o->) appears on the outside rim of the box.
 - The box grows more and more downwards for each hidden arrow
   (as if drawn transparently, instead of being there for effect only).

In both engines:
 - Uncommenting the hidden & arrow enlarges the box downwards.
   Explanation: because of the invisible U-turn self-referential arrow.
   => Better use the pattern `C-[hidden]>B` instead of `C-[hidden]>C` where possible!
[...]

PR #1842, opened 2024-06-28
fix for issue #1097

jimnelson372:
Here is a fix for issue #1097, improving how left edge message arrows are drawn when in a group, by positioning the text of the message as close to the right side of the arrow as possible (matching the positioning of the text in short arrows).

The changes were:

## Coding changes

`CommunicationExoTile` -- 
- the `getPoint1(..)`, `getPoint2(...)`, `getPoint1Value(...)`, `getPoint2Value(...)` and `getMaxX()` methods were modified to ensure that the message layout position and sizing of edge arrows was the same as for short arrows, while the drawing values extended to the edges.   
- Two methods were added to determine if it is a Right or Left edge message: `isFromRightBorderMessage()` and `isFromLeftBorderMessage()`.   
- To properly position the anchor arrows to their original position before these changes, the `getMiddleX()` method was overridden.
- Finally, so that `ComponentRoseArrow` could properly position the message text despite being passed the border position for these edge arrows, a new value `textDeltaX` is calculated and assigned the the Area class.  This is similar to how Self message arrows pass `deltaX1` in the Area class for information about the activity level.  I thought to reuse deltaX1, but the semantics were too different that I chose to add a new value to Area just for this purpose.

`ComponentRoseArrow` -- The `drawInternalU` method now adjusts the message text position based on `Area.textDeltaX`.  For non-edge arrows, this value is 0 and there is no adjustment.

`Area` -- just added the `textDeltaX` value and its getter and setter.

`CommonTitle` -- removed the final from `getMiddleX` so this could be overridden in `CommunicationExoTile`.

## nonreg and pdiff changes

- Two nonreg tests were changed, since text positioning on left edge arrows changed.
- Several diagrams in pdiff that used left edge arrows were changed, for the better, by this fix: 
-- 2vwir  (a group within a group example)
-- 4bsiu   (the example I shared in today's comment about anchor positioning)
-- 4bsiu
-- 90zhp 
-- a4ksd   (another group within a group example)
-- go1md  (a larger diagram that had one left edge arrow where the text was shifted right)
-- i7m1f  (The original issue diagram for 1097)
-- imwcw
-- tq3kq  (A larger example I had shared when demonstrating a different proposed fix, now working better.)

Here is what that last diagram, tq3kq, now generates, with the improvement being in the "left edge arrow" section.

image

## Two additional, new, diagrams.

The comments for the issue have examples, but I wanted to provide two more scripts to point out some possible additional improvements that could be worked on, though not needed to resolve the current issue.

1. The following script makes clear that the grouping of the edge arrows matches the sizing and positioning of the short arrows on both sides.   (It also, though, demonstrates that there is a separate remaining issue with the actual text positioning on left and right arrows relative to the ends of the arrows, for all arrow types, not just edge arrows, so that can be looked at later.)

```
@startuml
!pragma teoz true
skinparam ParticipantPadding 18
  A -> B
& B <- C
== short arrows ==
opt#transparent
?o-> B : word
& B <-o? : word
end
?o-> B : word
& B <-o? : word
== left edge arrow ==
opt#transparent
[o-> B : word
end
[o-> B : word
== right edge arrow==
opt#transparent
& B <-o] : word
end
B <-o] : word
@enduml
```
I'll just show the diagram as fixed.  Prior to the fix, the grouping of the first left edge arrow was all the way at the border, rather than just the end of the arrow extending to the border.   

testteozgroup_013

The next script demonstrates what happens, though, if the text message on the edge arrows is long enough to push out the borders so that the grouping box is drawn near the border.   You can see that for both my fix to the edge arrows and for the original short arrow grouping layout, the group boxes are drawn the same.  But with the edge arrows, it becomes clear that the padding on the group box is slightly different for the left and right sides relatives to the borders.   I just wanted to show this example to show how the diagrams will look, but like the text position adjustments I suggested on the previous diagram, changes to the group padding can be done in a later pull request.

```
@startuml
!pragma teoz true
participant B
== extra long text both sides, short vs edge ==
opt#transparent OK
   [o-> B : This text extends the arrow length\nright edge arrow
& B <-o] : This text extends the arrow length\nright edge arrow
end
opt#transparent OK /'in teoz'/
?o-> B : This text extends the arrow length\nshort arrow
& B <-o? : This text extends the arrow length\nshort arrow
end
@enduml
```
testteozgroup_010

Let me know if you have any questions or concerns.

Jim Nelson
@jimnelson372

### DIFF
--- a/src/net/sourceforge/plantuml/sequencediagram/teoz/CommonTile.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/teoz/CommonTile.java
@@ -65,7 +65,7 @@ public abstract class CommonTile implements Tile, UDrawable {
 		return stringBounder;
 	}
 
-	final public double getMiddleX() {
+	public double getMiddleX() {
 		final double max = getMaxX().getCurrentValue();
 		final double min = getMinX().getCurrentValue();
 		return (min + max) / 2;
--- a/src/net/sourceforge/plantuml/sequencediagram/teoz/CommunicationExoTile.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/teoz/CommunicationExoTile.java
@@ -105,13 +105,19 @@ public class CommunicationExoTile extends AbstractTile {
 		final XDimension2D dim = comp.getPreferredDimension(stringBounder);
 		double x1 = getPoint1Value(stringBounder);
 		double x2 = getPoint2Value(stringBounder);
+
+		final double textDeltaX = isFromLeftBorderMessage()
+																? getPoint1(stringBounder).getCurrentValue() - x1
+																: isFromRightBorderMessage()
+																		? getPoint2(stringBounder).getCurrentValue() - x2
+																		: 0;
+
 		final int level = livingSpace.getLevelAt(this, EventsHistoryMode.IGNORE_FUTURE_DEACTIVATE);
 		if (level > 0) {
 			if (message.getType().isRightBorder())
 				x1 += CommunicationTile.LIVE_DELTA_SIZE * level;
 			else
 				x2 += CommunicationTile.LIVE_DELTA_SIZE * (level - 2);
-
 		}
 
 		final ArrowConfiguration arrowConfiguration = message.getArrowConfiguration();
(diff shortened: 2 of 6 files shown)


arnaudroques:
Thank you very much for your contribution!

Would it be possible to add your last two examples to `pdiff`? If you'd like, I can take care of it.

jimnelson372:
Thank you @arnaudroques!   

Yes, please add those two examples to `pdiff` for me.  I haven't yet used the import feature of `pdiff` and I'm not yet sure when I'll have time to do that.   I'll make sure to learn to do that on future fixes and create pull requests on that project at the same time.

The-Lum:
Hello all,

_[Just for the record]_
This is this corresponding commit:
- https://github.com/plantuml/pdiff/commit/0ac2100fc5f0ab0bb23f5be6333abb3193a8bbb3

_Thanks to @arnaudroques and @jimnelson372 ._
Regards,
Th.

PR #1843, opened 2024-06-28
- set RenderingHints.VALUE_FRACTIONALMETRICS_ON in DriverTextG2

kirchsth:
Hello PlantUML team, @arnaudroques, @The-Lum,

related to [C4-Stdlib discussion:Why is font rendering for line text so bad?](https://github.com/plantuml-stdlib/C4-PlantUML/discussions/347) and  [Forum entry: svg exports of diagrams with maxMessageSize/wrapWidth have spacing issues](https://forum.plantuml.net/18858/exports-diagrams-maxmessagesize-wrapwidth-spacing-issues)

**Observation:**
It seems that the distance/layout between the letters is not correct and should be improved.
I tried 3 different scenarios and combined it in a svg output file (first line orig implementation; second line all characters are written as one text entry and the third with the new implementation: RenderingHints.VALUE_FRACTIONALMETRICS_ON).

SampleWithExactXValues 

```plantuml
@startuml
skinparam arrow {
    FontSize 8
}

a -> a : **Perform the usual session auth**
@enduml
```

I think the implementation with the VALUE_FRACTIONALMETRICS_ON produces a much better layout.

My [TextUsesFractionalMetrics_inclTests branch](https://github.com/kirchsth/plantuml/tree/feat/TextUsesFractionalMetrics_inclTests) has a more complex test [Test_TextFractionalMetric](https://github.com/kirchsth/plantuml/blob/feat/TextUsesFractionalMetrics_inclTests/test/test/example/Test_TextFractionalMetric.java) with following output

I found KEY_FRACTIONALMETRICS and KEY_TEXT_ANTIALIASING on multiple places maybe it should be set on another place
or via the UChange pattern like the UAntiAliasing class.

BR Helmut

PS.: I removed additional the duplicated lines in the if/else path.

### DIFF
--- a/src/net/sourceforge/plantuml/FileFormat.java
+++ b/src/net/sourceforge/plantuml/FileFormat.java
@@ -127,8 +127,8 @@ public enum FileFormat {
 	final static private BufferedImage imDummy = new BufferedImage(100, 100, BufferedImage.TYPE_INT_RGB);
 	final static public Graphics2D gg = imDummy.createGraphics();
 	static {
-		// KEY_FRACTIONALMETRICS
 		gg.setRenderingHint(RenderingHints.KEY_TEXT_ANTIALIASING, RenderingHints.VALUE_TEXT_ANTIALIAS_ON);
+		gg.setRenderingHint(RenderingHints.KEY_FRACTIONALMETRICS, RenderingHints.VALUE_FRACTIONALMETRICS_ON);
 	}
 
 	public StringBounder getDefaultStringBounder() {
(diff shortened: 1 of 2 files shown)


kirchsth:
In previous diagram I compared the old online and new local version. Following image shows the real difference between the different local jar versions:
- the 3. line is calculated via local jar with old implementation
- the 4. line is calculated via local jar with new implementation
    
SampleWithExactXValues

arnaudroques:
I agree that text rendering looks better now.
Let's see what other people say about this! :-)

kirchsth:
Thank you for the merge.
I hope we get a lot of (positive ;-) )feedback too.
I asked @rcdailey in the [C4-Stdlib discussion](https://github.com/plantuml-stdlib/C4-PlantUML/discussions/347#discussioncomment-9911822) too.

Issue #1804, opened 2024-06-02
[Seq] Allow `>>` and `>` on the same arrow


The-Lum:
## Context
To follow:
- #1678
- [QA-18958 - [Seq] Cannot have two end of Arrow having different style](https://forum.plantuml.net/18958/sequence-diagram-cannot-have-arrow-having-different-style)

## Wanted feature
- Could you allow `>>` and `>` on the same arrow, on seq. diag.?

## Example (on V1.2025.5)

```puml
@startuml
Test -->> Test: OK
Test <<-- Test: OK
Test <<-->> Test: the arrow is KO
Test <<->> Test: the arrow is KO
Test <<--> Test: the arrow is KO
Test <->> Test: the arrow is KO
Test <<-> a: the arrow is KO
Test <->> a: the arrow is KO
Test <<--> a: the arrow is KO
Test <-->> a: the arrow is KO
@enduml
```
>[![](https://www.planttext.com/api/plantuml/svg/SoWkIImgAStDuGf9BIvHqDMrirC0iQmK_BqXGZOskhfOXE3gIZ9I5HABYlBB5JAB5RpzuIe8Aj35hu086nGIyMd6AGcs6hTM77eTKlDIWE440000)](https://www.planttext.com/?text=SoWkIImgAStDuGf9BIvHqDMrirC0iQmK_BqXGZOskhfOXE3gIZ9I5HABYlBB5JAB5RpzuIe8Aj35hu086nGIyMd6AGcs6hTM77eTKlDIWE440000) 

_[FYI @jimnelson372]_
Regards,
Th.

jimnelson372:
Hi @The-Lum,

Yes, I see in the code that when the message instruction for two ended arrows is being processed, it is currently assuming that if either arrow is async, then both are async.   I believe that can be fixed without too much difficulty, other than verifying that nothing else depends on this.   Right now, I'm unclear about the semantics of a double arrow on a sequence diagram, but if users find it to be meaningful and useful, then it should support the variety of arrows.

[...]

PR #1859, opened 2024-07-14
fix for issue #1804

jimnelson372:
This is a fix for issue #1804, to allow double head arrows to have different arrow types on each side.   Previously, if either arrow head was async then both would be async and self arrows would only draw both ends if both were non-async.

## The changes involved

`CommandArrow.java` --  the `executeArg` method now identifies each arrow head as async independently of each other.  It also reverses them for reverse arrows.

`ArrowConfiguration.java` -- the original `isAsync` method is now split into two methods: `isAsync1` and `isAsync2`.

`Step1Message.java` -- I had originally modified the `getSelfArrowType` method to use the new `ArrowConfiguration` `isAsync1` and` isAsync2` methods, but later removed that `isAsync()` test altogether since the `withHead1` and `withHead2` were called afterwards anyway.

`ComponentRoseSelfArrow.java` -- the `drawRightSide` method for right side self messages now also draws a starting arrow if one is given.

I've added the following diagram script (both with and without teoz) to pdiff and I will create a pull request for them shortly.

## Test Script
```
@startuml
Test -->> Test: OK
Test <<-- Test: OK
Test <<-->> Test: the arrow is KO
Test <<->> Test: the arrow is KO
Test <<--> Test: the arrow is KO
Test <->> Test: the arrow is KO
Test <-> Test: the arrow is KO
Test <<-> a: the arrow is KO
Test <->> a: the arrow is KO
Test <<--> a: the arrow is KO
Test <-->> a: the arrow is KO
@enduml
```

## New Results 

I believe it is the same as in the  issue, though I added the <-> case ((in the middle) to demonstrate that working before and after.   This now will generate:

For Puma:

testarrow2_001

For Teoz:

testarrow2_002

Please let me know if you have any questions or concerns.   I've run this change through pdiff and only the new diagram results in a change -- as expected.

Jim Nelson
@jimnelson372

### DIFF
--- a/src/net/sourceforge/plantuml/sequencediagram/command/CommandArrow.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/command/CommandArrow.java
@@ -227,7 +227,8 @@ public class CommandArrow extends SingleLineCommand2<SequenceDiagram> {
 		final Participant p2;
 		final boolean circleAtStart;
 		final boolean circleAtEnd;
-
+		final boolean sync1;
+		final boolean sync2;
 		if (reverseDefine) {
 			// Keep the order
 			// See https://github.com/plantuml/plantuml/issues/1819#issuecomment-2158524871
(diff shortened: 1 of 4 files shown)


arnaudroques:
This is perfect!
Thanks!
Arnaud

PR #1863, opened 2024-07-22
Added background color to arrows and draw background for circles

lprc:
Recently I created [this feature request in the forum (#19075)](https://forum.plantuml.net/19075/request-support-displaying-found-messages-filled-black-circle). I'd like to have the ability to fill circles of lost and found messages in sequence diagrams. Not knowing PlantUML internals at all, today I had the time to look into it myself. With my modifications you can use the following style to fill circles created with an "o".

    <style>
    sequenceDiagram {
        arrow {
            BackgroundColor Black;
        }
    }
    </style>

This is probably not the intended way to implement such a feature (it would be better if BackGroundColor was set for circle instead of arrow, but that would need more changes I guess). However maybe someone who has more knowledge can use this as a starting point...

### DIFF
--- a/skin/plantuml.skin
+++ b/skin/plantuml.skin
@@ -2,7 +2,7 @@ root {
   --common-background: #f1f1f1;
   --note-background: #FEFFDD;
   --grey-blue: #e2e2f0;
-  
+
   FontName SansSerif
   HyperLinkColor blue
   HyperLinkUnderlineThickness 1
@@ -90,7 +90,7 @@ element {
   composite {
     title {
       FontStyle bold
-    }  
+    }
   }
 }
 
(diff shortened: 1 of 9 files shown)


arnaudroques:
Thanks for your contribution.
It looks like JUnit tests do not pass: did you try them locally?

You can also try https://github.com/plantuml/pdiff project to see if they are any regression.

Tell us if you need some help!

lprc:
Oh sorry, yeah I didn't run them locally before.

I just did that and 

- fixed a bug (forgot self-messages before) and
- made black background color the default for arrows, i.e. circles at arrow tips should now always be filled.

If the default behavior is undesired, I can change that again, just tell me if you want that. With my changes, the default `--common-background` color was used instead of `NULL_COLOR`. 

Now the Unit Tests are running without failure locally.

arnaudroques:
> If the default behavior is undesired, I can change that again, just tell me if you want that.

Well, I'm not sure which behavior is the best one :-)
Let's see what other users say!

Anyway, thanks for your contribution!

Issue #1392, opened 2023-04-13
Produce graalvm reachability metadata in the CI and contribute it to the central graalvm database


bcopy:
**Is your feature request related to a problem? Please describe.**

I would like to compile native graalvm applications producing plantuml diagrams, but the native compilation process fails out of the box, this is made more complicated because graalvm reachability metadata for plantuml is not available centrally (at runtime, the native executable complains that classes are missing).

**Describe the solution you'd like**

I would like for the metadata to be generated as part of the Plantuml CI and advertised to https://github.com/oracle/graalvm-reachability-metadata (contribution guidelines are available here : https://github.com/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md). This would greatly simplify the compilation of plantuml-based applications with the Maven or Gradle native plugins, as well as the Spring Framework.

**Describe alternatives you've considered**

I have found inspiration in existing successful native compilations of plantuml, but haven't come across one for Spring Boot, nor is the Plantuml reachability metadata listed in https://github.com/oracle/graalvm-reachability-metadata/blob/master/library-and-framework-list.json

The-Lum:
Hi @bcopy 

- Could you re-test with last release v1.2026.5?
_in which there was a lot of refactoring of the natives building_ (Thanks @arnaudroques)

And new config files on:
- https://github.com/plantuml/plantuml/tree/master/src/main/resources/META-INF/native-image

Regards,
Th.

PR #1360, opened 2023-03-25
Build the native image as part of the release process

ggrossetie:
### What I Did

- Export `release_version` as a job output using `gradle properties -q | grep "version:" | awk '{print $2}'`
- Replace `POM_VERSION` by `RELEASE_VERSION` in `release-snapshot.sh` and `release.sh`
- Add the Linux binary (amd64) in the release directory
- Introduce a job named `build_artifacts` to build PlantUML
- Add a new job named `native_images` to build native images from the artifacts (i.e., jar files) produced by `build_artifacts`
- Use `actions/cache/save@v3` and `actions/cache/restore@v3` to share artifacts between jobs
- Call `.github/workflows/native-image.yml` (reusable workflow) in `ci.yml`

Here's the build artifact uploaded on the CI workflow:

image

As you can see, there' a `plantuml-linux-amd64-M.YYYY.m-SNAPSHOT` binary included.

PR #1868, opened 2024-07-31
Native image build

asm0dey:
Hi folks, I've slightly changed the native image build to build automatically the full-fledged GUI flavor of the application. I did it only for the GPLv2 flavor because it's a native binary, so nobody can depend on its source and won't have to disclose binaries of their own applications. I've also added Mac amd64 builds.
I had to change the build slightly: it's beneficial to use the Gradle GraalVM plugin because it handles the application resource better, and PlantUML needs a lot of resources to operate according to docs (skins, localization, icons, etc.). However, be aware that the build of native image will work only on [Liberica NIK](https://bell-sw.com/pages/downloads/native-image-kit/#nik-23-(jdk-21)), since it's the only GraalVM native image distribution supporting Swing without much additional tuning.
Therefore, if a developer wants to run `./gradlew plantuml-gplv2:nativeCompile` they will need to have Liberica NIK installed and the env GRAALVM_HOME pointing to it, or it should be the default JDK.

To test the result, you can check out the [releases section of my fork](https://github.com/asm0dey/plantuml/releases/tag/1.2024.7beta4)

### DIFF
--- a/.github/workflows/native-image.yml
+++ b/.github/workflows/native-image.yml
@@ -1,11 +1,10 @@
 name: Native Image
 
 on:
-  workflow_call:
-    inputs:
-      release-version:
-        required: true
-        type: string
+  workflow_dispatch:
+  push:
+    branches:
+      - master
 
 jobs:
   build_non_win_images:
@@ -17,6 +16,8 @@ jobs:
           - os: 'ubuntu-latest'
             platform: 'linux-amd64'
           - os: 'macos-latest'
+            platform: 'darwin-arm64'
+          - os: 'macos-13'
             platform: 'darwin-amd64'
           - os: 'windows-latest'
             platform: 'win-amd64'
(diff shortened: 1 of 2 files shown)


asm0dey:
Currently, the build fails, because we can't build native image for Java 8, is it possible to work this issue around? My belief is we don't actually care about the Java version of the native binary because a user is not intended to have Java installed to run the native image.

arnaudroques:
Yes, you can try with a more recent version of java for the native build.

asm0dey:
Yes, if you look at my changes in the PR, the native image uses Java 21 and works correctly (I think). But here I can't force 8 on ubuntu-latest to pass

arnaudroques:
> I would be grateful for any comments on what I should do to have this PR integrated

Yes, sorry, the team is off for a couple of weeks...
We are going to look at this PR carefully at the end of the month.

Sorry again for the delay, but thanks for your work!

asm0dey:
Nothing to be sorry for, I just wasn't sure if anybody is interested (which is also perfectly fine, I can support my own fork for this matter).
Thank you for attention!

asm0dey:
Folks, I'm sorry, but I am not sure that the workflow file should remain as I changed it. Currently it pushes the native image to the tag automatically on every push. Is it a desired behavior?

arnaudroques:
So, we ended up doing something very simple: we excluded Java 8 from the GPLv2 tests and build!  
We're open to better ideas if anyone have any suggestions.

Anyway, thanks again for your contribution and patience!

> Folks, I'm sorry, but I am not sure that the workflow file should remain as I changed it. Currently it pushes the native image to the tag automatically on every push. Is it a desired behavior?

I think it's fine because it is then published as a beta on [GitHub Releases](https://github.com/plantuml/plantuml/releases).
I guess that when the official 1.2024.7 release is published, further pushes won't overwrite it because we'll first increase the version number right after the release.

@The-Lum, any thoughts on this?

asm0dey:
It should be absolutely safe, no releases will be lost, but there is a chance that the source code in the release won't match binaries

The-Lum:
Hi @asm0dey, @arnaudroques, and all,

About:
>@The-Lum, any thoughts on this?

- What are your main goal?
- What is the architecture? on each pushes? on each pre-release? with tag? on release? on pre-release?

Perhaps to avoid to affect the normal release _(see #1881)_, it will be necessary to create, as the [snapshot](https://github.com/plantuml/plantuml/releases/tag/snapshot), 
a tag `native` _(and not a release, but more a pre-release)_ and put the native output on each pushes or changes?

_To debate,_
Regards,
Th.

The-Lum:
Hi @asm0dey, @arnaudroques, and all,

FYI there are some issues on "win native" creation:
- https://github.com/plantuml/plantuml/actions/runs/10557634039/job/29245545352#step:13:13

```
/usr/bin/tar: C\:/Users/runneradmin/.gradle/caches/modules-2/modules-2.lock: Read error at byte 0, while reading 38 bytes: Device or resource busy
/usr/bin/tar: Exiting with failure status due to previous errors
Warning: Failed to save: "C:\Program failed with error: The process 'C:\Program Files\Git\usr\bin\tar.exe' failed with exit code 2
Cache saved with the key: setup-graalvm-Windows-gradle-1178db426b67b175898d7651ff7b915c5dbbd49cc98a00eabfb545b9d0a36773
```

If you can check?

Regards,
Th.

asm0dey:
@The-Lum it's to be reported to the upstream, I think. As you can see it's in the "post" phase of the "setup graalvm" action, it's not under our control and we can do nothing about it as far as I understand.
As long as windows native build works correctly it's not a big issue too

The-Lum:
@asm0dey Thanks for this precision. 👍 

I just saw the error (wihtout any analyse) on:
- https://github.com/plantuml/plantuml/actions/runs/10557634039



Regards,
Th.

asm0dey:
> What is the architecture? on each pushes? on each pre-release? with tag? on release? on pre-release?

The current implementation publishes artifacts on push, I think it's totally fine, there is no need in any complex procedure like Maven central publication and it allows users to try pre-release versions early of they will.
However I also think we should publish native builds in releases too! SaaS's like kroki, planttext and others might be interested in this in my opinion.

The-Lum:
> However I also think we should publish native builds in releases too!

You are right...

_[And Just for the record]_
Here are some attempts to add native version on Release:
- #1345
- #1360
- #1362
- #1385
- #1391

See aslo this old or standby PR:
- #1395

And here is the original requests about _native_:
- #1344
- #1392


_[FYI @ggrossetie, @evantill, @bcopy: current work about native version of PlantUML...]_

_Thanks for all your contribution,_
Regards,
Th.

The-Lum:
@asm0dey 
> It should be absolutely safe, no releases will be lost, but there is a chance that the source code in the release won't match binaries

It seems there is a issue.
Native images ([_native_](https://github.com/plantuml/plantuml/releases/tag/native)) are updated (and overridden) well on each push but not the sources.

Do you know if with [svenstaro/upload-release-action](https://github.com/svenstaro/upload-release-action) it is possible to update source also?

Regards,
Th.

asm0dey:
I mentioned it :)

> the source code in the release won't match binaries

I think for this to work we should delete a tag and set it on the new HEAD. It's just how GitHub releases work

PR #1873, opened 2024-08-15
feat: add builtin function `%boolval`

dragondive:
This change adds the `%boolval` function to improve usability of boolean values and conditionals. This is especially helpful when the diagram data is generated in a previous workflow step, using a conventional programming or scripting language.

In this first implementation, the boolean strings are defined as:

* True values: `true`, `1`
* False values: `false`, `0`

These values are case-sensitive, meaning `True` `TRUE` are also treated as true values (likewise for false values).

This builtin function helps mitigate some user confusion with the use of boolean values, specifically that all strings including `"0"` and `"false"` are treated as true values.

Existing behaviour (integer `0` is false, all other numbers and strings are true) is not affected, and can be used by simply not calling the new `%boolval` function.

Other minor improvements:

1. Parsing error in `%intval` will now raise an exception. This improves tool behaviour consistency, for example, when `-noerror` flag is used from the CLI.
2. Functions in `TContext.java` are arranged in sorted order. This code readability improvement does not change the tool behaviour from the user's perspective.


dragondive:
Many programming languages have a dedicated boolean type. Languages not having a dedicated boolean type usually use integers instead. In those languages, when converting a boolean value to integer, usually for comparison with another boolean value, the integers `1` and `0` generally represent `true` and `false`.

PlantUML does not have a dedicated boolean type. Instead, it provides utility functions, such as `%not()`, `%true()` and `%false()`, to assist with boolean operations. The integer `0` is treated as `false` and all other numbers and strings (including `"0"` and `"false"`) are treated as `true`. This causes some inconsistencies and surprising behaviour, particularly for users familiar with one or more other programming languages (which describes majority of PlantUML users).

Some reasonable expectations from a boolean type are:

1. Given a set `S`, if a condition `C` filters subset `X` of `S`, then the complement condition `not(C)` should filter the complement subset `S - X`.
2. Complement of `true` should equal `false` (and vice versa) in all cases, regardless of how either operand is specified or computed.

Additionally, interfacing with data generated by other programming languages should be easy, which can be facilitated by:

1. A facility to treat conventional truthy values (`true` - C++/Java, `True` - Python, `1` - C) or falsy values (`false` - C++/Java, 'False` - Python, `0` - C) as users normally expect.
2. Treating boolean-like string values (`"true"`, `"false"`, ...) like boolean values, which is particularly relevant for data read from JSON file.

The proposed `%boolval` builtin function improves the usability of boolean type in PlantUML, complementing the existing above-mentioned utility functions. It does not replace the current behaviour, which is available by simply not using this function. Thus, existing code will not be affected, although users may consider "upgrading" it using `%boolval`.

Below are the example code, json data and some generated diagrams to illustrate the `%boolval` function usage, as well as the limitations of the current behaviour. **The title of the diagram shows the conditional expression used, rest of the plantuml code is the same.**

Note that JSON has a dedicated boolean type, the `0` and `1` used in the example data below is only for convenience. Also, some conditional expressions are bad coding style, such as inverted semantics (if highlight is false, then highlight, else don't), comparing `not(X)` to `false`, etc. They're used here for demonstration only.

```plantuml
@startuml languages
allowmixing
!$data = %load_json("languages-data.json")
!assert %size($data) != 0, "failed to load data."

title "%<U+200B>boolval($item.is_highlighted) == %true()"
legend "<color:green>OK</color> 😃"
json "<b>Languages Data</b>" as Data $data

!$highlight_color = $data.highlight_color

rectangle Languages {
    !foreach $item in $data.items
        !if %boolval($item.is_highlighted) == %true()
            rectangle "$item.name" #$highlight_color
        !else
            rectangle "$item.name"
        !endif
    !endfor
}

Data -[hidden]down-> Languages

@enduml
```

```json
{
    "highlight_color": "Gold",
    "items": [
        {
            "name": "C++",
            "is_highlighted": false
        },
        {
            "name": "Java",
            "is_highlighted": 1
        },
        {
            "name": "Python",
            "is_highlighted": 0
        },
        {
            "name": "PlantUML",
            "is_highlighted": true
        }
    ]
}
```

**`%boolval` usage examples**

%U+200Bboolval($item is_highlighted) == %U+200Btrue()

%U+200Bboolval($item is_highlighted)

%U+200Bboolval($item is_highlighted) == %U+200Bfalse()

%U+200Bnot(%U+200Bboolval($item is_highlighted) == %U+200Btrue()

%U+200Bnot(%U+200Bboolval($item is_highlighted) == %U+200Bfalse()

**limitations of current behaviour examples** 

$item is_highlighted

$item is_highlighted == %U+200Btrue()

$item is_highlighted == %U+200Bfalse()

$item is_highlighted == true

%U+200Bnot($item is_highlighted) == %U+200Btrue()

%U+200Bnot($item is_highlighted) == %U+200Bfalse()

%U+200Bnot($item is_highlighted)

%U+200Bnot($item is_highlighted) == true

src/net/sourceforge/plantuml/tim/stdlib/Modulo.java
```diff
@@ -30,6 +30,7 @@
  *
  *
  * Original Author:  Arnaud Roques
+ * Contribution:  Aravind Pai
```
dragondive, line 33: I didn't realize I could do this in my previous PR, so I'd like to squeeze it in with this one. 😄

arnaudroques:
Many thanks for yours contributions!
On thing you could also do is to add yours examples to https://github.com/plantuml/pdiff
This would be useful for non-regression tests.

Thanks again!

The-Lum:
Thank @dragondive, 👍 

That's now fixed:
- https://forum.plantuml.net/14930/why-cant-i-compare-json-variable-with-plantuml-boolean

Regards,
Th.

Issue #1344, opened 2023-03-19
Produce a native image using GraalVM


ggrossetie:
**Is your feature request related to a problem? Please describe.**
I would like to use PlantUML CLI as a native image:

```console
$ echo 'Alice->Bob: Hello' | ./plantuml -tpng -pipe > diagram.png
```

**Describe the solution you'd like**
I was able to produce a native image using GraalVM on Ubuntu:

```console
$ /path/to/graalvm-ce-java17-22.3.1/bin/native-image -H:ConfigurationFileDirectories=./native-image-config-dir --no-fallback --report-unsupported-elements-at-runtime -jar plantuml-1.2023.4.jar
========================================================================================================================
GraalVM Native Image: Generating 'plantuml-1.2023.4' (executable)...
========================================================================================================================
[1/7] Initializing...                                                                                    (2,5s @ 0,22GB)
 Version info: 'GraalVM 22.3.1 Java 17 CE'
 Java version info: '17.0.6+10-jvmci-22.3-b13'
 C compiler: gcc (linux, x86_64, 11.3.0)
 Garbage collector: Serial GC
[2/7] Performing analysis...  [******]                                                                  (12,1s @ 3,33GB)
  11 171 (88,31%) of 12 650 classes reachable
  21 887 (67,71%) of 32 323 fields reachable
  58 772 (61,05%) of 96 267 methods reachable
     200 classes,    96 fields, and 1 098 methods registered for reflection
     129 classes,   200 fields, and   115 methods registered for JNI access
       7 native libraries: dl, freetype, m, pthread, rt, stdc++, z
[3/7] Building universe...                                                                               (1,3s @ 3,25GB)
[4/7] Parsing methods...      [*]                                                                        (1,0s @ 2,32GB)
[5/7] Inlining methods...     [***]                                                                      (0,7s @ 2,75GB)
[6/7] Compiling methods...    [***]                                                                     (10,5s @ 3,33GB)
[7/7] Creating image...                                                                                  (1,8s @ 1,87GB)
  31,77MB (50,30%) for code area:    37 017 compilation units
  26,68MB (42,24%) for image heap:  292 532 objects and 109 resources
   4,72MB ( 7,47%) for other data
  63,17MB in total
------------------------------------------------------------------------------------------------------------------------
Top 10 packages in code area:                               Top 10 object types in image heap:
 953,14KB java.util                                            6,70MB byte[] for code metadata
 796,22KB sun.font                                             3,09MB java.lang.String
[...]

PR #1882, opened 2024-08-28
fix: make `native` release as `snapshot` release

### LINKED ISSUE #1881: Make latest release point to a general release
My [PlantUML installer script](https://github.com/iainelder/dotfiles/blob/master/programs/plantuml/install.bash) gets the latest release from GitHub.

It requests this URL: `https://api.github.com/repos/plantuml/plantuml/releases/latest`

Today that points to [1.2024.7beta5](https://github.com/plantuml/plantuml/releases/tag/1.2024.7beta5), which doesn't contain `plantuml.jar`. [Today's CI run failed](https://github.com/iainelder/dotfiles/actions/runs/10572252416/job/29302973796).

The latest general release would be [v1.2024.06](https://github.com/plantuml/plantuml/releases/tag/v1.2024.6).

Please apply the latest label to the latest general release, never a beta or pre-release version.

The-Lum:
Hi @arnaudroques, @asm0dey,

To follow:
- #1868

Awaiting arhitecture design about native release and debate about it _(see https://github.com/plantuml/plantuml/pull/1868#issuecomment-2312603215)_

Here is a proposal to make `native` as `snapshot`:
- only one tag (`native`)
- override (on each time)
- not a latest release, fixes #1881

That fixes:
- https://github.com/plantuml/plantuml/pull/1868#issuecomment-2309820122
- #1881

### DIFF
--- a/.github/workflows/native-image.yml
+++ b/.github/workflows/native-image.yml
@@ -53,6 +53,7 @@ jobs:
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: "plantuml-gplv2/build/native/nativeCompile/plantuml-${{ matrix. platform }}-${{ env.VERSION }}.zip"
-          tag: ${{ env.VERSION }}
+          tag: native
+          release_name: native (~${{ env.VERSION }})
           overwrite: true
-          make_latest: true
+          make_latest: false


asm0dey:
I think this can be augmented with a small addition: if a tag is present on the commit already than native should be published to the tag too, this way all the major releases will have native versions too

The-Lum:
@asm0dey 
> I think this can be augmented with a small addition: if a tag is present on the commit already than native should be published to the tag too, this way all the major releases will have native versions too

- Do you know what code should be added to do this?

I just propose this minor PR in order to fix the current defects...
And I understand that the topic of PlantUML major version and native release should be taken into account.
_(to add on my questions on https://github.com/plantuml/plantuml/pull/1868#issuecomment-2312603215)_

Regards,
Th.

asm0dey:
I have an idea, but need to check it. This PR is totally fine

The-Lum:
> I have an idea, but need to check it. This PR is totally fine

Good.

---

_[Just for the record]_
Here are some attempts to add native version on Release:
- #1345
- #1360
- #1362
- #1385
- #1391

See aslo this old or standby PR:
- #1395

And here is the original requests about _native_:
- #1344
- #1392

_Thanks for your contribution,_
Regards,
Th.

asm0dey:
I'm not sure about the reachability metadata, probably we can extract something useful from the build. The first feature request is fixed, it seems

Issue #1879, opened 2024-08-26
Crash happened on plantuml 1.2024.7beta5, related to "headless"


sunlin7:
Hi,

I tried the latest release but it always crash for no `DISPLAY` available.
I run the plantuml on Ubuntu 20.04 server without DISPLAY, or with the `headless` mode.
```
$ plantuml "-Djava.awt.headless=true" -headless ... 
```
And the error logs:

> Exception in thread "main" java.awt.AWTError: Can't connect to X11 window server using ':0.0' as the value of the DISPLAY variable.
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jni.functions.JNIFunctions$NewObjectWithObjectArrayArgFunctionPointer.invoke(JNIFunctions.java)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.jni.functions.JNIFunctions.ThrowNew(JNIFunctions.java:907)
        at java.desktop@21.0.2/sun.awt.X11GraphicsEnvironment.initDisplay(Native Method)
        at java.desktop@21.0.2/sun.awt.X11GraphicsEnvironment$1.run(X11GraphicsEnvironment.java:105)
        at java.base@21.0.2/java.security.AccessController.executePrivileged(AccessController.java:129)
        at java.base@21.0.2/java.security.AccessController.doPrivileged(AccessController.java:319)
        at java.desktop@21.0.2/sun.awt.X11GraphicsEnvironment.initStatic(X11GraphicsEnvironment.java:64)
        at java.desktop@21.0.2/sun.awt.X11GraphicsEnvironment.<clinit>(X11GraphicsEnvironment.java:59)
        at java.desktop@21.0.2/sun.awt.PlatformGraphicsInfo.createGE(PlatformGraphicsInfo.java:37)
        at java.desktop@21.0.2/java.awt.GraphicsEnvironment$LocalGE.createGE(GraphicsEnvironment.java:93)
        at java.desktop@21.0.2/java.awt.GraphicsEnvironment$LocalGE.<clinit>(GraphicsEnvironment.java:84)
        at java.desktop@21.0.2/java.awt.GraphicsEnvironment.getLocalGraphicsEnvironment(GraphicsEnvironment.java:106)
        at java.desktop@21.0.2/java.awt.image.BufferedImage.createGraphics(BufferedImage.java:1182)
        at net.sourceforge.plantuml.FileFormat.<clinit>(FileFormat.java:162)
[...]

PR #1883, opened 2024-08-29
Fixes headless build

asm0dey:
Not two versions of native images are generated: full and headless.


asm0dey:
This hopefully addresses the issue #1879

arnaudroques:
> PS: Sorry, took some time to polish it until it works

No problem: it rarely works on the first try, especially on complex subjects!

Thank you very much for your contribution; this will definitely be very useful for users who want to use native builds.

However, could you please explain why we need both full and headless versions? I'm not entirely clear on this myself :-)

asm0dey:
@arnaudroques that's simple: if you wanna run it s desktop app - you will need the full version, otherwise headless. Since it's a native image app - it's very strict about its requirements. And the full version requires the real awt implementation, while the headless version requires the headless implementation.

Did I answer the question?

PR #1898, opened 2024-09-09
update gradle to gradle-8.10, which permits java-22, java-23.

soloturn:
compile works, tests run through.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.10-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists


The-Lum:
Hi @soloturn, @arnaudroques, 

- Should we also update the [gradle-wrapper.jar](https://github.com/plantuml/plantuml/blob/master/gradle/wrapper/gradle-wrapper.jar) jar?
  - On https://github.com/plantuml/plantuml/tree/b056f45bcd6269f3b455112336735089c04a3576/gradle/wrapper
- or should we exclude it from the repo.?

Regards,
Th.

arnaudroques:
> * Should we also update the [gradle-wrapper.jar](https://github.com/plantuml/plantuml/blob/master/gradle/wrapper/gradle-wrapper.jar) jar?
>   
>   * On https://github.com/plantuml/plantuml/tree/b056f45bcd6269f3b455112336735089c04a3576/gradle/wrapper
> * or should we exclude it from the repo.?

I am really not sure.
Any idea about the best practice? What other Java project are doing?

The-Lum:
Here are some elements:
- https://stackoverflow.com/questions/25769536/how-when-to-generate-gradle-wrapper-files
- https://docs.gradle.org/current/userguide/gradle_wrapper.html
- https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper

Regards,
Th.

The-Lum:
Hi @arnaudroques, @soloturn,

After rapidity search, in fact the CI does not use `gradle-wrapper.jar`, it uses the local `gradle` *_(and it is already on 8.10!)_*... 
- https://github.com/plantuml/plantuml/actions/runs/10768938506/job/29859074155#step:4:19

So there is no impact on the CI, but only on the use by external users of the repository fork.
Then:
- @arnaudroques you can accept this PR as it.
- and we will see later if it is necessary to update `gradle-wrapper.jar`...

Regards,
Th.

The-Lum:
We will ready for gradle 9....
Next step...

soloturn:
@The-Lum  @arnaudroques the wrapper does not change a lot, so it is ok to [upgrade it from time to time, but not always](https://docs.gradle.org/current/userguide/gradle_wrapper.html#:~:text=This%20is%20usually%20fine%20as,wrapper%20task%20a%20second%20time.).

The-Lum:
We can wait Gradle v9...

PR #1909, opened 2024-09-15
Add support for colors in Chen EER diagrams

### LINKED ISSUE #1890: Sometimes difficult to detect to which arrow a text in a Chen diagram belongs
We are doing [these](https://plantuml.com/en/er-diagram) at school, and actually I am supposed to use DIA, but I prefer PlantUML. My teacher is confused by the fact that it is sometimes difficult to see which number belongs to which arrow. In [use case diagrams](https://plantuml.com/en/use-case-diagram#39a64d49d97a1ea0) I colored the arrows and their text in different colors to make it clear, but I did not find a way to do this in Chen diagrams. Could we have this here too (e.g. a syntax like `MyRelation =N=<<red>> MyEntity`)? Alternatively, could we bring the arrow text closer to the relationship or to the entity?

Benjamin-Davies:
Adds support for colors in Chen EER diagrams using existing color-parsing code.

Example:

```
entity CUSTOMER #lightblue;line:blue {
  Number <<key>> #lime;line:orange
  Bonus <<derived>>
  Name <<multi>>
}

relationship RENTED_TO #pink;line:red {
  Date
}

RENTED_TO -1- CUSTOMER #lime
```

Fixes #1890

### DIFF
--- a/src/net/sourceforge/plantuml/cheneer/command/CommandAssociate.java
+++ b/src/net/sourceforge/plantuml/cheneer/command/CommandAssociate.java
@@ -43,6 +43,8 @@ import net.sourceforge.plantuml.command.CommandExecutionResult;
 import net.sourceforge.plantuml.command.SingleLineCommand2;
 import net.sourceforge.plantuml.decoration.LinkDecor;
 import net.sourceforge.plantuml.decoration.LinkType;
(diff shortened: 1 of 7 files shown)


nilslindemann:
Thanks @Benjamin-Davies!

Issue #1734, opened 2024-04-10
LaTeX labels not centered


capmlbpkel:
**Describe the bug**
When using output format .tex instead of .png, labels which used to be centered, are not centered anymore.

**To Reproduce**

PNG
1. Create png output file via command line

LaTeX
1. Create tex output file via command line
2. Run pdflatex on tex file to obtain a pdf file

**Expected behavior**
Labes in the pdf file have the same alignment as in the png file.

**Screenshots**

PNG
![image](https://github.com/plantuml/plantuml/assets/162835982/56701448-0e4d-46b5-ad6a-39bd45a978e7)

PDF (look at the actor label)
![image](https://github.com/plantuml/plantuml/assets/162835982/72cd9943-38a2-4392-8ed1-9d4845ced82b)

**Source**
```
@startuml
actor Actor
component Component
Actor -- Component
@enduml
```

**Version**
1.2024.4

**Additional context**
Maybe a combination of left anchored nodes in tikz together with a deviating font / font size is the cause.

liudongmiao:
<img width="200" alt="dark" src="https://github.com/user-attachments/assets/3c44897d-e324-4186-bdd2-e378850d8c85">

capmlbpkel:
I can confirm that #1914 solves the issue, thank you very much!

PR #1913, opened 2024-09-21
tikz: make node size to parent, then center it

liudongmiao:
is there any better way to get size of text block and parent?

should resolve https://github.com/plantuml/plantuml/issues/1734


liudongmiao:
#1914 should be a better solution

Issue #1606, opened 2023-11-16
mindmap: node MaximumWidth: weird latex result


fcorneli:
Consider the following `example.puml`:
```plantuml
@startmindmap

<style>
node {
    MaximumWidth 250
}
</style>

* This is a test
** Text is looking weird in Latex
** Second Item here
*** The problem seems to be due to the MaximumWidth setting I guess.

@endmindmap
```
Compile via:
```
java -jar $PLANTUML_JAR -tlatex example.puml
pdflatex example.tex
```
Results is a very weird looking layout. The text is way off.

fcorneli:
All looking good now, thanks.
New result:
<img width="1144" alt="Scherm­afbeelding 2024-09-24 om 20 44 26" src="https://github.com/user-attachments/assets/3ce6b102-d196-460d-b5b0-57ecddb660f0">
Old result:
<img width="1142" alt="Scherm­afbeelding 2024-09-24 om 20 40 31" src="https://github.com/user-attachments/assets/9c4f09fe-f1c2-418b-b4c8-cb20d54ff92f">

PR #1914, opened 2024-09-22
use latex to get string bounder

liudongmiao:
use !pragma to configure tex and preamble, and the default is like this:

```
!pragma TexSystem xelatex
```

latex implementations is inspired by matplotlib

### DIFF
--- a/src/net/sourceforge/plantuml/FileFormat.java
+++ b/src/net/sourceforge/plantuml/FileFormat.java
@@ -224,24 +224,46 @@ public enum FileFormat {
 
 	private StringBounder getTikzStringBounder(final TikzFontDistortion tikzFontDistortion) {
 		return new StringBounderRaw(FileFormat.gg.getFontRenderContext()) {
+
+			private final LatexManager latexManager = new LatexManager(tikzFontDistortion.getTexSystem(), tikzFontDistortion.getTexPreamble());
+
 			public String toString() {
 				return "FileFormat::getTikzStringBounder";
 			}
 
 			protected XDimension2D calculateDimensionInternal(UFont font, String text) {
-				text = text.replace("\t", "    ");
-				final XDimension2D w1 = getJavaDimension(font.goTikz(-1), text);
-				final XDimension2D w2 = getJavaDimension(font.goTikz(0), text);
-				final XDimension2D w3 = getJavaDimension(font.goTikz(1), text);
-				final double factor = (w3.getWidth() - w1.getWidth()) / w2.getWidth();
-				final double distortion = tikzFontDistortion.getDistortion();
-				final double magnify = tikzFontDistortion.getMagnify();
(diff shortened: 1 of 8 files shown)


liudongmiao:
@arnaudroques This PR use latex to get string bounder, please help to review.

liudongmiao:
This should resolve #1734

src/net/sourceforge/plantuml/FileFormat.java
```diff
@@ -224,25 +224,47 @@ public boolean matchesProperty(String propertyName) {
...
+				final boolean italic = font.isItalic();
+				final boolean bold = font.isBold();
+
+				if (italic)
+					sb.append("\\textit{");
+
+				if (bold)
+					sb.append("\\textbf{");
+
+				sb.append(text);
```
liudongmiao, line 259: The tex source need to synced with:
https://github.com/plantuml/plantuml/blob/83a4bfbdb398ee2e15657c77e80d142321b89d16/src/net/sourceforge/plantuml/tikz/TikzGraphics.java#L386-L403

However, to capture box, LatexManager use catcode to avoid protect.

arnaudroques: Thanks for your contribution, but I am not sure to understand your point here.
Could you explain a little?

liudongmiao: Currently, plantuml use Java to calculate dimension then render them with LaTeX, so they don't match.

Codes to calculate dimension for text (master version):
https://github.com/plantuml/plantuml/blob/83a4bfbdb398ee2e15657c77e80d142321b89d16/src/net/sourceforge/plantuml/FileFormat.java#L231-L241

Codes to display text:
https://github.com/plantuml/plantuml/blob/83a4bfbdb398ee2e15657c77e80d142321b89d16/src/net/sourceforge/plantuml/tikz/TikzGraphics.java#L317-L353

As they are unmatched, so it would show weird results.

This PR use LaTeX to calculate dimension for text:
https://github.com/plantuml/plantuml/blob/05112b7bd5ab5c9bbbc7aa43639d9f602e15a3ee/src/net/sourceforge/plantuml/FileFormat.java#L234-L237

However, the LaTeX source to calculate dimension may be slight different from the LaTeX source to render text, as there are `protectedText` in the LaTeX source when redner:

https://github.com/plantuml/plantuml/blob/83a4bfbdb398ee2e15657c77e80d142321b89d16/src/net/sourceforge/plantuml/tikz/TikzGraphics.java#L386-L403

Furthermore, there maybe some other problem. For example in https://github.com/plantuml/plantuml/issues/1606, the text is rendered by every word.

Weird results:
weird

weird

However, as the API returns the `height` and `depth`, it could be fixed if `text height` is specified with the maximum of the entire words, and `text depth` is re-calculated with the actual `text height` and `text depth` of single word (ref https://tikz.dev/tikz-shapes#sec-17.4.4)

PR #1919, opened 2024-09-24
tikz: make all fonts with the same baseline

### LINKED ISSUE #1606: mindmap: node MaximumWidth: weird latex result

liudongmiao:
In AtomText, the minimal font height is set to 10. And in LaTeX, different character may has different height and depth. This PR find the height and descent of the first AtomText, then make sure other AtomText share the same baseline with the first AtomText.

This should resolve #1606

### DIFF
--- a/src/net/sourceforge/plantuml/klimt/creole/Sea.java
+++ b/src/net/sourceforge/plantuml/klimt/creole/Sea.java
@@ -36,18 +36,21 @@
 package net.sourceforge.plantuml.klimt.creole;
 
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
 
 import net.sourceforge.plantuml.klimt.creole.atom.Atom;
+import net.sourceforge.plantuml.klimt.creole.legacy.AtomText;
 import net.sourceforge.plantuml.klimt.font.StringBounder;
+import net.sourceforge.plantuml.klimt.font.UFont;
 import net.sourceforge.plantuml.klimt.geom.MinMax;
 import net.sourceforge.plantuml.klimt.geom.XDimension2D;
 
 public class Sea {
 
 	private double currentX;
-	private final Map<Atom, Position> positions = new HashMap<Atom, Position>();
+	private final Map<Atom, Position> positions = new LinkedHashMap<Atom, Position>();
 	private final StringBounder stringBounder;
 
 	public Sea(StringBounder stringBounder) {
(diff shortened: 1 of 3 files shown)


src/net/sourceforge/plantuml/klimt/creole/Sea.java
```diff
@@ -36,18 +36,21 @@
...
 import net.sourceforge.plantuml.klimt.font.StringBounder;
+import net.sourceforge.plantuml.klimt.font.UFont;
 import net.sourceforge.plantuml.klimt.geom.MinMax;
 import net.sourceforge.plantuml.klimt.geom.XDimension2D;
 
 public class Sea {
 
 	private double currentX;
-	private final Map<Atom, Position> positions = new HashMap<Atom, Position>();
+	private final Map<Atom, Position> positions = new LinkedHashMap<Atom, Position>();
```
liudongmiao, line 53: Change to LinkedHashMap as we want to find the first AtomText.

src/net/sourceforge/plantuml/klimt/creole/Sea.java
```diff
@@ -85,6 +88,32 @@ public void translateMinYto(double newValue) {
...
+		double firstTextHeight = Double.NaN;
+		for (Map.Entry<Atom, Position> entry : new LinkedHashMap<>(positions).entrySet()) {
+			final Atom atom = entry.getKey();
+			if (!(atom instanceof AtomText)) {
+				continue;
+			}
+			AtomText atomText = (AtomText) atom;
+			UFont font = atomText.getFontConfiguration().getFont();
+			String text = atomText.getText();
+			double height = stringBounder.calculateDimension(font, text).getHeight() - stringBounder.getDescent(font, text);
```
liudongmiao, line 102: get the height without descent.

src/net/sourceforge/plantuml/klimt/creole/Sea.java
```diff
@@ -85,6 +88,32 @@ public void translateMinYto(double newValue) {
...
+			if (height == 0.0) {
+				continue;
+			}
+			if (Double.isNaN(firstTextHeight)) {
+				firstTextHeight = height;
+				continue;
+			}
+			double delta = firstTextHeight - height;
+			if (delta != 0.0) {
+				positions.put(atom, entry.getValue().translateY(delta));
```
liudongmiao, line 112: now we find the height delta.

src/net/sourceforge/plantuml/klimt/creole/SheetBlock1.java
```diff
@@ -140,13 +140,16 @@ private void initMap(StringBounder stringBounder) {
...
 			final double width = sea.getWidth();
 			widths.put(stripe, width);
 			minMax = sea.update(minMax);
 			final double height = sea.getHeight();
 			heights.put(stripe, height);
 			y += height;
+			if (stringBounder.matchesProperty("TIKZ")) {
+				sea.doAlignTikzBaseline();
+			}
+			sea.exportAllPositions(positions);
```
liudongmiao, line 152: sea.update, and sea.getHeight would re-calculate the height of positions. However, we only want to change the translation without the actual height, so move the adjust at last.

PR #1922, opened 2024-09-25
tikz: use the same protectText when calc dimension and render

liudongmiao:
- `~` / `\u00AB` / `\u00BB`, use the replacement as render
- `<` / `>`, ignore as non special characters

### DIFF
--- a/src/net/sourceforge/plantuml/FileFormat.java
+++ b/src/net/sourceforge/plantuml/FileFormat.java
@@ -246,6 +246,9 @@ public enum FileFormat {
 			}
 
 			protected String styleText(UFont font, String text) {
+				if (font == null) {
+					return "$" + text + "$";
+				}
 				StringBuilder sb = new StringBuilder();
 				final boolean italic = font.isItalic();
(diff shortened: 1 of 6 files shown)


liudongmiao:
@arnaudroques Please help to review.
Finally, Latex format in PlantUML use the same tex source to calc dimension and render, would avoid weird results.

And the document should be updated too:
- `skinparam TikzFont` is removed, we should document it, and remove the codes
- `pragma  TexSystem` and `pragma TexPreamble` is added, we should document it

arnaudroques:
> @arnaudroques Please help to review. Finally, Latex format in PlantUML use the same tex source to calc dimension and render, would avoid weird results.
> 
> And the document should be updated too:
> 
>     * `skinparam TikzFont` is unused, we should document it, and remove the codes
> 
>     * `pragma  TexSystem` and `pragma TexPreamble` is added, we should document it

Thanks for your contribution!
Yes, I will do a review: this is in the TODO list :-)

Note that you can go on with your modifications: I am sure that you know LaTeX better than us!

Issue #1628, opened 2023-12-18
[math/latex] Improve horizontal alignment of `math`/`latex`


The-Lum:
Hi PlantUML team,

## Context
- https://forum.plantuml.net/15401/font-size-math-bigger-than-rest-could-this-configured-same-font

Observed on:
- Version `V1.2023.13`
- Activity diagram

## Wanted feature
- Could you align `math`/`latex` horizontally with the other text?
_(on a first step: middle align will be good...)_

## Example
```puml
@startuml
:This is a test with <math>y^2</math>and math <math>a*x^2+b*x+c</math>inside text;
:Let the formula: <math>int_0^1f(x)dx</math>;
@enduml
```
>[![](https://www.planttext.com/api/plantuml/png/SoWkIImgAStDuR8AoSWiLW2YH8MIrE8IXVBCaWm5czp4aWoxoZWZ6tqmAp4lHG74W4eaQbN46MadQLLe9qCLPEOLPwQa0aseAB7cil99BL4eoKXLICiloYtDIRI2QilCAuatY3DCqwZGJAc0QhJcSaZDIm4w1000)](https://www.planttext.com/?text=SoWkIImgAStDuR8AoSWiLW2YH8MIrE8IXVBCaWm5czp4aWoxoZWZ6tqmAp4lHG74W4eaQbN46MadQLLe9qCLPEOLPwQa0aseAB7cil99BL4eoKXLICiloYtDIRI2QilCAuatY3DCqwZGJAc0QhJcSaZDIm4w1000)

Regards,
Th.

The-Lum:
Hi me, 😸 

Here is a possible workaround, using _full math sans serif text_:
```puml
@startuml
split
:This is a test with <math>y^2</math>and math <math>a*x^2+b*x+c</math>inside text;
:<math>sf"This is a test with " y^2 sf" and math " a*x^2+b*x+c sf" inside text"</math>;
kill
split again
:Let the formula: <math>int_0^1f(x)dx</math>;
:<math>sf"Let the formula:"int_0^1f(x)dx</math>;
kill
@enduml
```
[...]

PR #1923, opened 2024-09-25
tikz: optimize latex render for #1628

liudongmiao:


src/net/sourceforge/plantuml/AbstractPSystem.java
```diff
@@ -136,7 +136,7 @@ final public long seed() {
 		return getSource().seed();
 	}
 
-	public int getNbImages() {
+	public int getNbImages(FileFormatOption formatOption) {
```
The-Lum, line 139: Hi @liudongmiao,

A simple dummy question:
- Why `getNbImages` needs now of `formatOption`?

Regards,
Th.

arnaudroques: I agree with @The-Lum : this is somehow a breaking change, and does not sound very logical to me.

Could you tell us more about this?

Thanks!

liudongmiao: In AthomMath, the dimension is cached.

https://github.com/plantuml/plantuml/blob/f186e7a6864933629014681345017e6f42ddce8e/src/net/sourceforge/plantuml/klimt/creole/atom/AtomMath.java#L74-L79

There are two way to fix:
- pass the real `StringBounder` when possible (previous commit)
- don't cache for debug (I have reverted the commit, and make a new commit)

liudongmiao:
@arnaudroques I have walked the open issues with latex and examples in https://plantuml.com/zh/ascii-math.
Now the tikz should work now.

The only issue is Embedded Diagram #1926.

The-Lum:
Hello @liudongmiao @arnaudroques 

Here is a global question about package architecture and folder design:
- Is it plan to put together all the file/package (LatexManager, TikzFontDistortion) about latex and tikz in a same folder?

Hint:
- https://github.com/plantuml/plantuml/tree/master/src%2Fnet%2Fsourceforge%2Fplantuml%2Fklimt%2Fdrawing%2Ftikz
Or
- https://github.com/plantuml/plantuml/tree/master/src%2Fnet%2Fsourceforge%2Fplantuml%2Ftikz
[Or in a new folder
- `plantuml/latex` ]

_To debate..._

Regards,
Th.

liudongmiao:
@The-Lum @arnaudroques Yes, it should. And, actually, three things:
- remove `TikzFontDistortion`. `Pragma` to configure `TexSystem` and `TexPreamble` shouldn't be put in file named as `TikzFontDistortion`. Now we use `LaTeX` to get width/height/depth, there should be no font distortion.
- add copyright template in `LatexManager`.
- add test cases. However, these cases would requires runtime TexLive.

The-Lum:
Hello @liudongmiao

FYI, here are some latex test cases from `asciimath2tex`:
- 
https://github.com/plantuml/plantuml/blob/master/test/net/sourceforge/plantuml/math/ASCIIMathTeXImgUnitTest.java

Regards,
Th

liudongmiao:
@The-Lum Not the same. I means the same input latex uml, produces the same tex output.
As the width, height, depth is related to LaTeX, maybe the font and TeX version, so the tex output may not the same.

@The-Lum @arnaudroques Do you know is there any reason to redefind `~`?
https://github.com/plantuml/plantuml/blob/d2b2bcf1722b8705f7f01a556dc96751e7739f7d/src/net/sourceforge/plantuml/tikz/TikzGraphics.java#L398-L401
IMO, it should be converted to `~`, we don't need to ajdust it. If user want to custom it, there is other way.
// Current LatexManager copied this settings too, however, it seems strange.
https://github.com/plantuml/plantuml/blob/bc0becdb83acda69904066e745821032a7690740/src/net/sourceforge/plantuml/LatexManager.java#L115-L116

PR #1936, opened 2024-10-01
test: test ExportOnUText or test `Utext` for 'class' or 'component' 

The-Lum:
Hello PlantUML team,

- How to test `Utext` for 'class' or 'component' diagram without `dot`?

Here is an attempt with some tests with this PR.

--- 
Here is a minimal example:
```puml
@startuml
class a
@enduml
```

✔️  On environment with `dot` we observe:
```
,-.
|a|
|-|
`-'

```
[_Link to output_](https://www.planttext.com/api/plantuml/txt/SoWkIImgAStDuKhEIImkLaZaSaZDIm7o0G00)

❌ On environment without `dot` we observe this exception:

```cy
    java.lang.IllegalStateException
        at net.sourceforge.plantuml.dot.AbstractGraphviz.createFile3(AbstractGraphviz.java:103)
        at net.sourceforge.plantuml.posimo.GraphvizSolverB.solve(GraphvizSolverB.java:107)
        at net.sourceforge.plantuml.dot.CucaDiagramTxtMaker.<init>(CucaDiagramTxtMaker.java:120)
        at net.atmp.CucaDiagram.createFilesTxt(CucaDiagram.java:392)
        at net.atmp.CucaDiagram.exportDiagramInternal(CucaDiagram.java:416)
        at net.sourceforge.plantuml.classdiagram.ClassDiagram.exportDiagramInternal(ClassDiagram.java:83)
        at net.sourceforge.plantuml.UmlDiagram.exportDiagramNow(UmlDiagram.java:139)
        at net.sourceforge.plantuml.AbstractPSystem.exportDiagram(AbstractPSystem.java:207)
        at net.sourceforge.plantuml.SourceStringReader.outputImage(SourceStringReader.java:189)
        at net.sourceforge.plantuml.Pipe.generateDiagram(Pipe.java:105)
        at net.sourceforge.plantuml.Pipe.managePipe(Pipe.java:96)
        at test.test.ExportOnUTextTest.renderViaPipe(ExportOnUTextTest.java:51)
        at test.test.ExportOnUTextTest.assertRenderExpectedOutput(ExportOnUTextTest.java:59)
        at test.test.ExportOnUTextTest.RenderTest(ExportOnUTextTest.java:33)
```

- Is it normal or not?
- Is `dot` is mandatory?

_Thanks for your lookout,_
Regards,
Th.

### DIFF
--- a/test/test/test/ExportOnUTextTest.java
+++ b/test/test/test/ExportOnUTextTest.java
@@ -1,0 +1,62 @@
+package test.test;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.util.Lists.newArrayList;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+import java.util.Collections;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+import net.sourceforge.plantuml.ErrorStatus;
+import net.sourceforge.plantuml.Option;
+import net.sourceforge.plantuml.Pipe;
+
+/**
+ * Tests the Render UText
+ */
+class ExportOnUTextTest {
+
+    @ParameterizedTest(name = "[{index}] {1}")
+	@CsvSource(value = {
+        "'@startuml\na -> b\n@enduml\n', a",
+        "'@startuml\nclass a\n@enduml\n', a",
+        "'@startuml\n!pragma layout smetana\nclass a\n@enduml\n', a",
+	})
+	void RenderTest(String input, String expected) throws Exception {
+		assertRenderExpectedOutput(input, expected);
+	}
+
+    // TODO: to Factorize on a specific test package...
+    //
+    private static final String[] COMMON_OPTIONS = {"-tutxt"};
+
+    private String[] optionArray(String... extraOptions) {
+        final List<String> list = newArrayList(COMMON_OPTIONS);
+        Collections.addAll(list, extraOptions);
+        return list.toArray(new String[0]);
+    }
+
+    private String renderViaPipe(String diagram, String... extraOptions) throws Exception {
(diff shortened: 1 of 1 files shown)


The-Lum:
For the UTEXT export...

Hints, see:
https://github.com/plantuml/plantuml/blob/090efa98697d2f9b8ff5000d50ef9765a134b249/src/net/sourceforge/plantuml/dot/CucaDiagramTxtMaker.java#L150

In order to manage UText perhaps it will be necessary to use... `drawBoxSimpleUnicode` :-)

arnaudroques:
Even in the case of ASCII Art diagrams, we need dot because the positioning calculations are done by `dot`.  
Now, there is the degenerate case where there is only one class: obviously, in this case, the positioning is trivial, so we could very well generate this case in ASCII Art.  
(By the way, that's what happens with PNG/SVG, I believe we don’t use `dot` in this case.)

And there might be a second issue unrelated to `dot`, which is that the drawing of boxes purely in Unicode may simply not be implemented.  
I need to look into that :-)

The-Lum:
Hello @arnaudroques 

Thanks for the answer.

1/ `dot`
> Even in the case of ASCII Art diagrams, we need dot because the positioning calculations are done by `dot`.  

- Is `Smetana` would take the job?
- Could you plan to use `smetana` in alternative if we use `!pragma layout smetana`? 

2/ Unicode for Class
> And there might be a second issue unrelated to `dot`, which is that the drawing of boxes purely in Unicode may simply not be implemented.  

With my previous hint, I Can take this action, and propose a new PR.

Thanks,
Regards,
Th.

The-Lum:
Hi @arnaudroques,
>Even in the case of ASCII Art diagrams, we need dot because the positioning calculations are done by dot.
Now, there is the degenerate case where there is only one class: obviously, in this case, the positioning is trivial, so we could very well generate this case in ASCII Art.
(By the way, that's what happens with PNG/SVG, I believe we don’t use dot in this case.)

1. Dot or smetana 
    - Is Smetana would take the job?
    - Could you plan to use smetana in alternative if we use `!pragma layout smetana`?

2. Not use of `dot` or `smetana` if there is only one class
    - Could you, _a minima_, don't use `dot` when there is only one element?

Regards,
Th.

Issue #1788, opened 2024-05-22
exception on two transitions to the same deep history node


fjouault:
**Describe the bug**
The following PlantUML code yields an exception with versions >= v1.2023.2, but works as expected with versions <= v1.2023.1:

```plantuml
@startuml
state B
A --> B[H*] : ev1
A --> B[H*] : ev2
@enduml
```

There is no issue with regular history nodes, only with deep history nodes.

For instance, with version v1.2023.2, the exception is:
```cy
Error java.lang.IllegalStateException
java.lang.IllegalStateException
	at net.atmp.CucaDiagram.reallyCreateLeaf(CucaDiagram.java:180)
	at net.sourceforge.plantuml.statediagram.StateDiagram.getDeepHistory(StateDiagram.java:164)
	at net.sourceforge.plantuml.statediagram.command.CommandLinkStateCommon.getEntity(CommandLinkStateCommon.java:165)
	at net.sourceforge.plantuml.statediagram.command.CommandLinkStateCommon.getEntityEnd(CommandLinkStateCommon.java:151)
	at net.sourceforge.plantuml.statediagram.command.CommandLinkStateCommon.executeArg(CommandLinkStateCommon.java:82)
	at net.sourceforge.plantuml.statediagram.command.CommandLinkStateCommon.executeArg(CommandLinkStateCommon.java:60)
	at net.sourceforge.plantuml.command.SingleLineCommand2.execute(SingleLineCommand2.java:150)
	at net.sourceforge.plantuml.command.ProtectedCommand.execute(ProtectedCommand.java:57)
	at net.sourceforge.plantuml.AbstractPSystem.executeCommand(AbstractPSystem.java:166)
	at net.sourceforge.plantuml.command.PSystemCommandFactory.executeFewLines(PSystemCommandFactory.java:125)
	at net.sourceforge.plantuml.command.PSystemCommandFactory.createSystem(PSystemCommandFactory.java:108)
	at net.sourceforge.plantuml.PSystemBuilder.createPSystem(PSystemBuilder.java:145)
	at net.sourceforge.plantuml.BlockUml.getDiagram(BlockUml.java:187)
	at net.sourceforge.plantuml.SourceFileReaderAbstract.getGeneratedImages(SourceFileReaderAbstract.java:170)
	at net.sourceforge.plantuml.Run.manageFileInternal(Run.java:524)
	at net.sourceforge.plantuml.Run.processArgs(Run.java:407)
[...]

PR #1959, opened 2024-10-24
fix: Allow multiple transitions to the same deep history (on state diagram) 

The-Lum:
Hello PlantUML Team,

Attempt to resolve plantuml/plantuml#1788

Report of:
- https://github.com/plantuml/plantuml/commit/5d38bb6ef4a325135d8b64a8caadd8056211b786 _(correction on `getHistorical(String idShort)` for History issue https://forum.plantuml.net/18282)_

But here for Deep History on `getDeepHistory(String idShort)`.

---

@arnaudroques:
- Could you confirm me the good line? _with `ident` or `""`_

Because I hesitated between:
```java
result = reallyCreateLeaf(ident, Display.getWithNewlines(ident), LeafType.DEEP_HISTORY, null);
```
or
```java
result = reallyCreateLeaf(ident, Display.getWithNewlines(""), LeafType.DEEP_HISTORY, null);
```

---
Regards,
Th.

### DIFF
--- a/src/net/sourceforge/plantuml/statediagram/StateDiagram.java
+++ b/src/net/sourceforge/plantuml/statediagram/StateDiagram.java
@@ -135,7 +135,7 @@ public class StateDiagram extends AbstractEntityDiagram {
 		final Quark<Entity> ident = quarkInContext(true, tmp);
 		final Entity result;
 		if (ident.getData() == null)
-			result = reallyCreateLeaf(ident, Display.getWithNewlines(ident), LeafType.PSEUDO_STATE, null);
+			result = reallyCreateLeaf(ident, Display.getWithNewlines(""), LeafType.PSEUDO_STATE, null);
 		else
 			result = ident.getData();
 		endGroup();
@@ -169,7 +169,7 @@ public class StateDiagram extends AbstractEntityDiagram {
 		final Quark<Entity> ident = quarkInContext(true, cleanId(tmp));
 		final Entity result;
 		if (ident.getData() == null)
-			result = reallyCreateLeaf(ident, Display.getWithNewlines(ident), LeafType.DEEP_HISTORY, null);
+			result = reallyCreateLeaf(ident, Display.getWithNewlines(""), LeafType.DEEP_HISTORY, null);
 		else
 			result = ident.getData();
 		endGroup();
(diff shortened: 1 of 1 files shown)


arnaudroques:
Thank you for your debugging!

Both pieces of code work because the "Entity" objects do not have any "Display" associated with them. Only `H*` is shown.

However, the more logical approach is:
```
result = reallyCreateLeaf(ident, Display.getWithNewlines(""), LeafType.DEEP_HISTORY, null);
```

This is because they truly do not have any display.

Congratulations on your work!

The-Lum:
@arnaudroques:
Thank you for this feedback.
	
For standardization, should we do the same on `getHistorical(String idShort)` for:
https://github.com/plantuml/plantuml/blob/8eb791f39478778788fd47a9195dc1b2feb3eade/src/net/sourceforge/plantuml/statediagram/StateDiagram.java#L138

- Can I make the change at the same time _(in order to change `ident` into `""`)_?

Regards,
Th.

arnaudroques:
> * Can I make the change at the same time _(in order to change `ident` into `""`)_?

Yes, sure.

And by the way, you should try with `Display.NULL`, I am not sure it will work, but if it does, it's even a better solution.

The-Lum:
Hello @arnaudroques,

On `getDeepHistory(String idShort)`, `Display.NULL` seems to be ok.

But as soon as I replace `Display.getWithNewlines("")` by `Display.NULL` elsewhere..

As on `getDeepHistory()`, I get:
```cy
Error java.lang.IllegalArgumentException
java.lang.IllegalArgumentException
    at net.atmp.CucaDiagram.reallyCreateLeaf(CucaDiagram.java:208)
    at net.sourceforge.plantuml.statediagram.StateDiagram.getDeepHistory(StateDiagram.java:158)
    at net.sourceforge.plantuml.statediagram.command.CommandLinkStateCommon.getEntity(CommandLinkStateCommon.java:163)
    at net.sourceforge.plantuml.statediagram.command.CommandLinkStateCommon.getEntityStart(CommandLinkStateCommon.java:145)
    at net.sourceforge.plantuml.statediagram.command.CommandLinkStateCommon.executeArg(CommandLinkStateCommon.java:78)
    at net.sourceforge.plantuml.statediagram.command.CommandLinkStateCommon.executeArg(CommandLinkStateCommon.java:60)

```
Then I will keep `Display.getWithNewlines("")`.

Regards,
Th.

PR #1995, opened 2024-12-04
Add floating participants header to interactive SVG sequence diagram

michaelomichael:
When you're scrolling through a very large sequence diagram, it's very difficult to know which participants each message refers to because the participants labels at the top are scrolled off the top of the screen and those at the bottom aren't visible yet:

image

This PR introduces new behaviour to sequence diagrams when the diagram has `!pragma svginteractive true`, such that the participant labels/headers at the top will be "sticky": they will remain at the top of the screen as the user scrolls down through the diagram:

image

The change introduces new `<g>` elements to group together and identify participants in the SVG, plus fixes to the existing non-sequence-diagram JS files so that they play nicely when you have multiple embedded interactive `<svg>`s in the same HTML page.

I added some tests for the grouping, although I suspect that they might be a pain to maintain as they will fail if the expected positions of each SVG element change at all, so let me know if there's another approach that you'd prefer.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -78,6 +78,8 @@ sourceSets {
 		}
 		resources {
 			srcDirs(".")
+			include("svg/**/*.css")
+			include("svg/**/*.js")
 			include("skin/**/*.skin")
 			include("themes/**/*.puml")
 		}
--- a/src/net/atmp/ImageBuilder.java
+++ b/src/net/atmp/ImageBuilder.java
@@ -393,14 +393,14 @@ public class ImageBuilder {
 		if (titledDiagram != null)
 			option = option.withTitle(titledDiagram.getTitleDisplay());
 
-		if ("true".equalsIgnoreCase(pragma.getValue("svginteractive")))
+		if (pragma.isSvgInteractive())
 			option = option.withInteractive();
 		if (skinParam != null) {
 			option = option.withLengthAdjust(skinParam.getlengthAdjust());
 			option = option.withSvgDimensionStyle(skinParam.svgDimensionStyle());
 		}
 
-		final UGraphicSvg ug = UGraphicSvg.build(option, false, seed, stringBounder);
+		final UGraphicSvg ug = UGraphicSvg.build(option, false, seed, stringBounder, titledDiagram != null ? titledDiagram.getUmlDiagramType() : null);
 		return ug;
 
 	}
--- a/src/net/sourceforge/plantuml/klimt/UGroupType.java
+++ b/src/net/sourceforge/plantuml/klimt/UGroupType.java
@@ -36,5 +36,5 @@
 package net.sourceforge.plantuml.klimt;
 
 public enum UGroupType {
-	ID, CLASS
+	ID, CLASS, PARTICIPANT_NAME, PARTICIPANT_1_NAME, PARTICIPANT_2_NAME
 }
--- a/src/net/sourceforge/plantuml/klimt/drawing/svg/SvgGraphics.java
+++ b/src/net/sourceforge/plantuml/klimt/drawing/svg/SvgGraphics.java
@@ -221,7 +221,7 @@ public class SvgGraphics {
 
 	private Element getStylesForInteractiveMode() {
 		final Element style = simpleElement("style");
-		final String text = getData("default.css");
+		final String text = getData(getInteractiveModeAddOnsBaseFilename() + ".css");
 		if (text == null)
 			return null;
 
@@ -229,6 +229,13 @@ public class SvgGraphics {
 		style.setAttribute("type", "text/css");
 		style.appendChild(cdata);
 		return style;
+	}
+
+	private String getInteractiveModeAddOnsBaseFilename() {
+		if (UmlDiagramType.SEQUENCE.equals(diagramType)) {
+			return "sequencediagram";
+		}
+		return "default";
 	}
 
 //	private Element getStylesForDarkness() {
@@ -261,7 +268,7 @@ public class SvgGraphics {
 
 	private Element getScriptForInteractiveMode() {
 		final Element script = document.createElement("script");
-		final String text = getData("default.js");
+		final String text = getData(getInteractiveModeAddOnsBaseFilename() + ".js");
 		if (text == null)
 			return null;
 
--- a/src/net/sourceforge/plantuml/klimt/drawing/svg/SvgGraphics.java
+++ b/src/net/sourceforge/plantuml/klimt/drawing/svg/SvgGraphics.java
@@ -60,6 +60,8 @@ import javax.xml.transform.TransformerException;
 import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.stream.StreamResult;
 
+import net.sourceforge.plantuml.skin.UmlDiagramType;
+
 import org.w3c.dom.CDATASection;
 import org.w3c.dom.Comment;
 import org.w3c.dom.Document;
@@ -124,6 +126,7 @@ public class SvgGraphics {
 	private final String gradientId;
 
 	private final SvgOption option;
+	private final UmlDiagramType diagramType;
 
 	private Element pendingBackground;
 	private boolean robotoAdded = false;
@@ -137,7 +140,7 @@ public class SvgGraphics {
 
 	}
 
-	public SvgGraphics(long seed, SvgOption option) {
+	public SvgGraphics(long seed, SvgOption option, UmlDiagramType diagramType) {
 		try {
 			this.document = getDocument();
 
@@ -145,7 +148,12 @@ public class SvgGraphics {
 			final XDimension2D minDim = option.getMinDim();
 			ensureVisible(minDim.getWidth(), minDim.getHeight());
 
+			this.diagramType = diagramType;
+
 			this.root = getRootNode();
+			if (option.isInteractive() && diagramType != null) {
+				root.setAttribute("data-diagram-type", diagramType.name());
+			}
 
 			// Create a node named defs, which will be the parent
 			// for a pair of linear gradient definitions.
@@ -1126,8 +1134,22 @@ public class SvgGraphics {
 		for (Map.Entry<UGroupType, String> typeIdent : typeIdents.entrySet()) {
 			if (typeIdent.getKey() == UGroupType.ID)
 				pendingAction.get(0).setAttribute("id", typeIdent.getValue());
-			if (option.isInteractive() && typeIdent.getKey() == UGroupType.CLASS)
-				pendingAction.get(0).setAttribute("class", typeIdent.getValue());
+			if (option.isInteractive()) {
+				switch (typeIdent.getKey()) {
+					case CLASS:
+						pendingAction.get(0).setAttribute("class", typeIdent.getValue());
+						break;
+					case PARTICIPANT_NAME:
+						pendingAction.get(0).setAttribute("data-participant", typeIdent.getValue());
+						break;
+					case PARTICIPANT_1_NAME:
+						pendingAction.get(0).setAttribute("data-participant-1", typeIdent.getValue());
+						break;
+					case PARTICIPANT_2_NAME:
+						pendingAction.get(0).setAttribute("data-participant-2", typeIdent.getValue());
+						break;
+				}
+			}
 		}
 	}
 
--- a/src/net/sourceforge/plantuml/klimt/drawing/svg/UGraphicSvg.java
+++ b/src/net/sourceforge/plantuml/klimt/drawing/svg/UGraphicSvg.java
@@ -57,12 +57,14 @@ import net.sourceforge.plantuml.klimt.shape.UPixel;
 import net.sourceforge.plantuml.klimt.shape.UPolygon;
 import net.sourceforge.plantuml.klimt.shape.URectangle;
 import net.sourceforge.plantuml.klimt.shape.UText;
+import net.sourceforge.plantuml.skin.UmlDiagramType;
 import net.sourceforge.plantuml.url.Url;
 
 public class UGraphicSvg extends AbstractUGraphic<SvgGraphics> implements ClipContainer {
 	// ::remove file when __HAXE__
 
 	private final boolean textAsPath;
+	private final UmlDiagramType diagramType;
 	private /* final */ SvgOption option;
 
 	public double dpiFactor() {
@@ -71,21 +73,22 @@ public class UGraphicSvg extends AbstractUGraphic<SvgGraphics> implements ClipCo
 
 	@Override
 	protected AbstractCommonUGraphic copyUGraphic() {
-		final UGraphicSvg result = new UGraphicSvg(getStringBounder(), textAsPath);
+		final UGraphicSvg result = new UGraphicSvg(getStringBounder(), textAsPath, diagramType);
 		result.copy(this);
 		result.option = this.option;
 		return result;
 	}
 
-	private UGraphicSvg(StringBounder stringBounder, boolean textAsPath) {
+	private UGraphicSvg(StringBounder stringBounder, boolean textAsPath, UmlDiagramType diagramType) {
 		super(stringBounder);
 		this.textAsPath = textAsPath;
+		this.diagramType = diagramType;
 		register();
 	}
 
-	public static UGraphicSvg build(SvgOption option, boolean textAsPath, long seed, StringBounder stringBounder) {
-		final UGraphicSvg result = new UGraphicSvg(stringBounder, textAsPath);
-		result.copy(option.getBackcolor(), option.getColorMapper(), new SvgGraphics(seed, option));
+	public static UGraphicSvg build(SvgOption option, boolean textAsPath, long seed, StringBounder stringBounder, UmlDiagramType diagramType) {
+		final UGraphicSvg result = new UGraphicSvg(stringBounder, textAsPath, diagramType);
+		result.copy(option.getBackcolor(), option.getColorMapper(), new SvgGraphics(seed, option, diagramType));
 		result.option = option;
 		return result;
 	}
--- a/src/net/sourceforge/plantuml/klimt/drawing/svg/UGraphicSvg.java
+++ b/src/net/sourceforge/plantuml/klimt/drawing/svg/UGraphicSvg.java
@@ -141,13 +141,12 @@ public class UGraphicSvg extends AbstractUGraphic<SvgGraphics> implements ClipCo
 				// create so we can register
 				// the event handlers on them we will append to the end of the document
 				getGraphicObject().addStyle("onmouseinteractivefooter.css");
-				getGraphicObject().addScriptTag("https://cdn.jsdelivr.net/npm/@svgdotjs/svg.js@3.0/dist/svg.min.js");
 				getGraphicObject().addScript("onmouseinteractivefooter.js");
 			}
 
 			getGraphicObject().createXml(os);
 		} catch (TransformerException e) {
-			throw new IOException(e.toString());
+			throw new IOException(e);
 		}
 	}
 
--- a/src/net/sourceforge/plantuml/sequencediagram/graphic/Arrow.java
+++ b/src/net/sourceforge/plantuml/sequencediagram/graphic/Arrow.java
@@ -35,6 +35,7 @@
  */
 package net.sourceforge.plantuml.sequencediagram.graphic;
 
+import net.sourceforge.plantuml.klimt.UGroupType;
 import net.sourceforge.plantuml.klimt.drawing.UGraphic;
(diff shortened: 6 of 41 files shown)


arnaudroques:
Thank you for your contribution! This is a nice PR and introduces a much-needed improvement for sequence diagrams, especially for large ones. The feature is well-thought-out and addresses a real usability issue.  

That being said, since the changes in this PR are substantial, I suggest that instead of merging it directly into `master`, we first create a new branch called `SequenceDiagramFloatingHeader` to incorporate these modifications. This will give us a chance to further refine and review the changes together before merging them into `master`.  

I’ve already created the `SequenceDiagramFloatingHeader` branch, so could you please update the PR to target this branch instead?  

### Points to Discuss:  
1. **Dependency Concerns**:  
   There’s a dependency on `net.sourceforge.plantuml.skin.UmlDiagramType` in the following files:  
   - `src/net/sourceforge/plantuml/klimt/drawing/svg/SvgGraphics.java`  
   - `src/net/sourceforge/plantuml/klimt/drawing/svg/UGraphicSvg.java`  

   While PlantUML’s architecture isn’t perfect, this introduces a coupling of layers that are conceptually at different levels. We should aim to avoid such inter-layer dependencies to maintain cleaner architecture.

2. **Method Signature**:  
   The new parameter `boolean shouldGroupComponents` in  
   ```java
   protected void drawInternalU(UGraphic ug, double maxX, Context2D context, boolean shouldGroupComponents)
   ```  
   is not ideal. It complicates the signature and could be simplified.  

   In fact, would it make sense for this flag to always be `true`? This could help reduce complexity in the code.  

### Proposal:  
Let’s merge your code into the `SequenceDiagramFloatingHeader` branch first. Once it’s there, we can collaboratively refactor and address the points mentioned above.  

Thank you again for your hard work on this! Let me know if you have any questions or if there’s anything I can help with during the branch update.

The-Lum:
FYI and for comment or review _(to all the original authors)_:
- @ioplker 
- @blipper
- @diguage
- @rubikscuber

Regards,
Th.

michaelomichael:
Hi @arnaudroques. I've updated the PR to point to the branch you created.

Regarding the `UmlDiagramType` proliferation: I don't know the architecture well enough yet to say what the best solution would be for this. Do you have any suggestions?

I'd be happy to always group the components. The only reason I didn't do this was in case some users had their own custom JS scripts that expected the SVG XML to follow certain patterns (I have some such scripts myself!), and adding the groups might break these people's scripts. On the other hand, I don't believe that the structure of the SVG is defined as an official API anywhere, so there should be no expectation of backwards compatibility for the SVG structure. Again, I'm happy to go with whatever you think is best here.

Let me know about any other changes you think should be considered too.

arnaudroques:
Thank you for updating the PR to target the new branch.  

### Regarding the `UmlDiagramType` Proliferation:  
You’re absolutely right; the architecture isn’t well-documented, and this can make it challenging to understand the best approach. I’m currently working on improving this—see [this commit](https://github.com/plantuml/plantuml/commit/f181f140411117522dca85987a7ae596161fd2ed) for some progress on this front.

### Always Grouping Components:  
You make a good point about the potential impact on custom JS scripts. However, as there is no official API guaranteeing the structure of the SVG, I agree that we should simplify things by always grouping components.  

Feel free to start working on this change directly in the `SequenceDiagramFloatingHeader` branch if that’s simpler for you—it’s entirely up to your preference. It’s a good step forward while we continue addressing the first point.  

Let me know if you have any other questions or need further clarification!

ioplker:
Hello, @michaelomichael, @arnaudroques and everyone! Glad to see the interactive mode improving! :hugs: 
A bit late here, but I can suggest another solution for this UX problem: tooltips.

We can add `<title>` element to each lifetime line and activation box.
What it gives us:
  - the diagram is not "dynamic" (sometimes it's annoying when you scroll the page and it suddenly changes)
  - we don't eat additional users' screen space (I think it will be always a relevant point, even when "everyone has 4k monitors nowadays")
  - `<title>` is an element with its own semantics and such a use case screams that it should be applied here
  - `<title>` is widely used already (don't need to cover any edge-cases like hovering on the edge or corners of the screen)
  - I suppose, that this kind of solution is more robust and simple to implement and support

That being said, this solution has some drawbacks:
  - we need to somehow increase the "hit box" for lifetimes as they are a couple pixels wide (I think it's pretty much doable with.. grouping :sweat_smile:)
  - tooltip's appearance (the built-in, let's not add js dependencies, please :pray:) depends on the browser and (as far as I'm aware) cannot be changed to suit diagram's theme (on the other hand, those tooltips aren't permanent parts of the diagram and most of us can live with this style inconsistency)

Also, we can add `<title>` to calls. In it we can state the sender and receiver. Although, I suspect it could be not so trivial to distinguish who is who (due to flexible arrow types).

P.S.
@The-Lum thank you for tagging me! :heart:

michaelomichael:
Hello @ioplker, 

Your suggestion about using `<title>` is excellent! Is there any reason why it shouldn't be the default behaviour for all sequence diagrams, even when `svginteractive` is not enabled? 

The tooltip will work really well for the case where you already _think_ you know what a given participant/lifeline is, and want to double-check its identity.
Similarly, it will be very helpful when there are one or two calls whose arrows point to a lifeline off-screen: hovering over the call will reveal the name of the receiver without the need to scroll horizontally.

I don't imagine that many people would be concerned about the styling of the tooltip, and I'd be happy to add the code for `<title>` and the extended hitbox in a separate PR.

That being said, some other downsides of a tooltips-only approach are:
1. There is a short delay before the tooltip appears, making it tedious to check the identity of more than one or two participants' lifelines.
2. It still requires the user to try to maintain a mental map of all of the different lifelines, as they can only see one lifeline's identity at a time.
3. It assumes that the person viewing the diagram has control over the mouse.

For background to this PR, the use case in which I find myself almost daily is where I'm looking at the centre of a huge sequence diagram with 15-20 participants and many more calls than can fit on a screen, and I want to locate the next call between, say, participant F and participant P, in order to see which calls P makes in response. Having to hover the mouse over each of the participants (and/or calls) in turn would make this possible, but it would be time consuming and frustrating. 

To compound this, these diagrams are often being presented on a Teams/Zoom call, so most of the viewers wouldn't have the opportunity to trigger an arbitrary tooltip due to not being able to move the cursor.

With regard to the screen real estate, the example I've shown above is a worst-case scenario, using lots of differently-shaped actors. When using the default rectangular participant shapes, the amount of space taken up by the floating header is much reduced.

I should add that I will also be creating a subsequent PR involving filters for the participants, in a similar vein to the `svginteractive` functionality that exists today for other diagram types. Again, having the floating header will make this future filtering much more usable.

I hope I've explained that well enough, and thanks again for the great suggestion.

ioplker:
@michaelomichael,

> Is there any reason why it shouldn't be the default behaviour for all sequence diagrams, even when `svginteractive` is not enabled?

I personally think it would be a sane default :)

---

In regard to use cases you've provided. If I've understood it well, we can summarize them as _big scheme exploration_ and _public presentation_.
Both cases will benefit from the floating header, you are right (_let's not argue the decomposition / grouping / coloring to simplify the diagrams as it's highly subjective and not possible from time to time_).

Nonetheless, the reason I've proposed the `title` solution is it's simplicity and robustness (_sorry, I feel somewhat responsible for starting this JS interactiveness thing_). And I'm afraid that at some point this can go wrong and break somebody's workflow (_as we are introducing some dynamic JS-logic to the client, which is more complex than a static image_). PlantUML has a dozen of plugins in all sorts of software from offline editors to online wiki's, and most of the time their users can't update or rollback them.

So, what do you say, if we "cover" this floating header appearance under some trigger?
By default, the header acts as is (not floating). But after clicking on the header or any of the lifelines it instantly sticks in place and becomes floating.
This way we can:
  - give the viewer to choose, if (_and when_) he wants the header to be floating or not
  - show perfectly valid diagram right away, even if JS somehow messes up after triggering the header to float


P.S.
Don't get me wrong - I appreciate the work you've done and would love to use this feature myself!
But, as we continue this interactive story, it's in our hands to make it painless for the end users (_both authors and viewers_).

michaelomichael:
@ioplker,

You're right to be cautious. Last thing I want is for someone to have to try to read an "svginteractive" sequence diagram and have a misbehaving header obscure the part they want to read!

I started out thinking that, if there are problems with the rendering/JS, then it's easy enough to remove the `!pragma svginteractive true` line and re-generate the diagram, but you're right to point out that not every person viewing the diagram will be in a position to do this. There's an option to have the JS loaded from a CDN at runtime, which would allow for centralised, on-the-fly fixes to any bugs, but that would involve versioning the SVG XML that we produce, and having to support every such flavour of XML that we produce from that point onward, and I _really_ don't want to go there.

I like your idea of a trigger, provided we can make the source of the trigger obvious to the user. If we hide the behaviour behind some esoteric sequence of clicks on otherwise innocuous parts of the diagram then few people will ever discover it, and fewer still will understand what they did to make it happen the first time.

I'm leaning towards the following approach:
* Leave the original header (and footer/tail) in place, and have the "floating" header be a clone of the original header;
* Add to the header a "pin" button that is usually hidden, but which appears (semi-transparent) when you hover over any part of the header, and is fully opaque when you hover over the button itself;
    * If the mouse continues to hover over the header (but doesn't hover directly on the button itself), then the button should eventually fade from view;
    * If the mouse moves away from the header completely, then the button should disappear immediately;
* Clicking the button will toggle the floating header behaviour on or off;
* The button's state should be remembered (a flag in `LocalStorage`) so the user doesn't have to click the button again every time they navigate to a different page.

Here's a quick proof of concept:
Toggle Example

The PoC is missing a means of indicating whether the floating header is enabled or not. My immediate thought is to do this using one version of the pushpin icon that has a solid fill ("enabled"), and another version in which the pushpin is just an outline, without any fill ("disabled").

Top-left feels like the most natural location for the button. When it's visible it might slightly obscure the first participant name, but the "button disappears again after a few seconds" behaviour helps mitigate this, and I think it strikes a good balance overall in favour of discoverability without being too in-your-face.

The JS code that's there can be made more robust too. Off the top of my head, we can:
* Perform checks at the start to make sure that all of the elements and methods we need are available;
* Disable the scrolling behaviour and hide the floating header (the original header will always be visible at the top of the document) if an error is detected at any point.

Let me know if you have any other suggestions.

ioplker:
@michaelomichael, PoC looks great! But can you, please, upload an example where the header actually floats? Sorry, if I've missed it somewhere :sweat_smile:
Am I right, that the pin button is placed on the floating header (when it is enabled)? I mean, could we see and press it after we scrolled down and the floating header is shown?

The pin button could use a tooltip, btw :upside_down_face:
Something like: "Toggle floating header.". Although, I'm not sure how we should handle localization for such things :thinking:
But an eng tooltip is better than a "silent" button.

Also, I'm a bit triggered when we talk about CDN and dependencies from the web for just viewing the diagram. I strongly think the svg must be fully self-contained. Dependencies could fail loading, be not accessible (e.g. in restricted intranet of a bank etc) or even be malicious (_welcome to my cozy paranoia corner, heh_).
And yes, as you've said, it would be a nightmare for supporting all the versions.

michaelomichael:
@ioplker I hadn't gotten as far as doing a floating version! 😁 The PoC was just to try out the mechanics of conditionally displaying a button. Yes, I anticipate the button being both on the floating header and on the non-floating version. (In both cases, it would only be shown when you hover over a part of either header.)

A tooltip would be great. The Gantt diagram type has a `language <xx>` command (see section 16.32 in the [reference guide](https://pdf.plantuml.net/PlantUML_Language_Reference_Guide_en.pdf)), but that's not available for other diagram types, and I can't see anyone bothering to add it to their diagrams just for the sake of a tooltip. I suppose we could include some strings in different languages in the embedded JS file, and use `navigator.language` to select one of them for the tooltip at runtime. Not sure how we'd go about actually getting (or verifying) translations though. Any thoughts?

Glad we're in agreement about not introducing external dependencies! 😁 

I'll put together a PoC with the button on a floating header and tooltip.

ioplker:
@michaelomichael, I'd love to see this PoC when you are ready :blush: 

About the translation.. I can see several paths on solving it:

1. Don't add a tooltip at all. No text to translate - no problems. The button will remain mysterious for new users, but a pushpin icon gives a good hint.

2. Don't bother translating the tooltip. That's ugly for non-english users, but somewhat more functional than previous option.

3. Translate the tooltip. We need to support god knows how many languages (_Gantt supports only 5.._). And this can lead to a global localization of other "interfaces" of other diagrams - for better or worse.

3.1. Translate with `language xx` as in Gantt diagram. This requires the author to add the `language xx` command to the diagram. And, as you've mentioned, not everybody would do this for an element, which is not on the screen all the time. So, we would have the same problem when this command isn't included (_however I suppose, we will use english as default_).

3.2 Translate with JS. JS `navigator.language` produces `BCP47` language tags according to the spec ([here](https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-language-dev) and [here](https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-language)). But Gantt diagram, as I can see, [uses ISO639](https://github.com/plantuml/plantuml.js/blob/main/plantuml-docs/modules/ROOT/pages/gantt-diagram.adoc#language-of-the-calendar). So we would need to handle and support the mapping between `BCP47` and `ISO639` for localization to work this way in the browser. And I think it's not so cost-efficient. At least for now when we are talking about a single tooltip.

---

Sooo.. I think it's more complicated than we want to handle right now :sweat_smile:
The first solution seems to be the optimal one.

It's certainly a good idea to make this svg-tooltips-and-other-gui-stuff localization one day. And maybe further svg interactiveness will greatly motivate to do this step.

That's my opinion btw, maybe someone else (_with more experience with PlantUML developing_) has something to offer. @arnaudroques what do you think?

---

P.S.
It's the section 16.34 in the reference guide, not 16.32 :sweat_smile:

arnaudroques:
Hi everyone,  

First, I want to thank you all for this fascinating and engaging discussion. It’s inspiring to see how much thought is going into improving the usability and versatility of PlantUML.  

### Regarding "Fully Self-Contained" SVGs  

I must admit, I’m not deeply experienced with SVGs or JavaScript, but achieving a fully self-contained approach might be challenging in today’s ecosystem. Dependencies have become increasingly common, and fully isolating functionality is complex.  

An alternative could be to create a dedicated project under the [PlantUML organization](https://github.com/plantuml/) that packages all the dependencies required by the generated SVGs. This project could be distributed via GitHub’s CDN. While it wouldn’t eliminate dependencies entirely, it would mitigate concerns about malicious sources and reduce the number of versions we’d need to support.  

That said, this approach may still introduce some complexity, so I leave it to all of you to evaluate its feasibility. I recognize this is a bit outside my comfort zone, but I hope the idea is helpful.  

### Regarding Localization (i18n)  

For i18n, I think we could take an interesting approach:  

1. **Extend the `language <xx>` Command**  
   - Make the `language <xx>` command (currently used in Gantt diagrams) available for all diagrams.  
   - Specifically, we could make it part of the preprocessor, perhaps using something like `!language <xx>`.  

2. **Expose Language Information in the Standard Library**  
   - Introduce a function in the standard library (e.g., `%language()`) that returns the diagram’s language.  

3. **Default Language Detection**  
   - Attempt to infer the user's language by:
     - Reading HTTP headers like `Accept-Language` (on www.plantuml.com or other hosting environments).  
     - Using JVM configuration when generating diagrams offline, so `%language()` provides a meaningful default even when not explicitly set in the diagram.  

While the JavaScript `navigator.language` is compelling, it has a significant limitation: the user's language is determined *after* the diagram has been generated (i.e., at the time of rendering in the browser). This delay creates a mismatch between the language used during diagram generation and the one shown to the user, which could lead to inconsistencies.  

By contrast, the approach I suggest ensures that the language is determined as early as possible—during diagram generation. This makes it possible for the entire diagram (including tooltips or labels) to reflect the chosen or detected language consistently.  

However, I acknowledge that my proposed method has its own limitation: the diagram will not dynamically adapt to the language of the person viewing it. This could be a significant drawback, especially if the goal is to create diagrams that are responsive to the user’s environment at the time of display.  

This trade-off is worth considering carefully. While generating static diagrams with consistent language settings may simplify some aspects, it may not align with the vision of offering truly dynamic and user-responsive diagrams.  

### Final Thoughts  

I hope this provides some clarity and sparks further discussion. Let me know if anything here is unclear or if there’s a way to improve these ideas further.

The-Lum:
Hello all,

_[Just for the record]_
FYI, this enhancement for _Sequence diagram_ will be perhaps use also for _Timing diagram_, as requested here:
- https://forum.plantuml.net/17984/timing-diagram-to-fix-the-title-when-scrolling-to-the-right

Regards,
Th.

michaelomichael:
@arnaudroques: Thanks for taking the time to share your thoughts on this. There are, thankfully, no 3rd party dependencies to include in the interactive SVG files - I've strived to use just standard APIs that are available on all modern browsers - so it's feasible to keep the SVGs fully self-contained for the current interactive features. It's great that you're willing to consider more sophisticated solutions for the future, should the need arise. 😄 

The language aspect is an interesting one. I think of the floating header controls as something that would, in an ideal world, be the responsibility of the _web browser_ rather than custom script within the SVG. (Imagine if web browsers understood the semantics of e.g. `<g floatable="true">` in an SVG, and automatically added UI elements to let you toggle that on and off. Maybe someday... 😄) If that were the case, then the tooltip language would definitely be driven by the setup of the user who is _viewing_ the diagram, and wouldn't change just because the web page you're visiting declared that it was in another language. (As an analogue, it would be upsetting for the user if, when they right-click on an amazon.com page, the context menu shows "Copy", but when they right-click on an amazon.fr page it changes to "Copier".) Of course we can never hope to have our tooltip translated into so many languages as an OS, but we ought to be able to cover at least some of the major ones, and people have the option of submitting a translation for new languages over time.

@The-Lum: Great spot! 😄  I haven't used Timing diagrams before but yes, it does seem like floating headers would be a great fit here (albeit scrolling horizontally rather than vertically). We should make this a separate PR, once the sequence diagram code is done and dusted.

michaelomichael:
@ioplker I've attached an example of the latest version. It shows multiple embedded interactive SVGs in an HTML page, with the floating header and toggle button active. 

[SvgsInScrollableDivs.html.zip](https://github.com/user-attachments/files/18197098/SvgsInScrollableDivs.html.zip)

diguage:
It's great. 👍👍👍

ioplker:
> @ioplker I've attached an example of the latest version. It shows multiple embedded interactive SVGs in an HTML page, with the floating header and toggle button active.
> 
> [SvgsInScrollableDivs.html.zip](https://github.com/user-attachments/files/18197098/SvgsInScrollableDivs.html.zip)

Nice one, thank you!
I'd like to propose some enhancements though:
1. We can place the pin button fixed at top left corner (_not in the header, but on page itself_) - so viewers can quickly check the header without the need to scroll up or down.
2. On the `SVG #1` you've showed a situation which is unusual to me. When the actor is placed down the diagram, not at the top with all others. Is it.. legit? :sweat_smile: If it is, then we should consider to put such actors to the floating header too.

image

And also a nit: I think it would be beneficial to minify the JS. It's always good to save some bandwidth :upside_down_face: 

---

P.S.
Maybe there is something wrong with my archiver, but all `<`, `>` and `&` were URL-encoded (_like `&amp;` for `&`_). JFYI.

michaelomichael:
@ioplker, thanks for your feedback on the example. 😀

1. Thoughts on pinning the toggle button:
	* It's an interesting idea, but I think it has the same drawbacks as the original "always on" floating header - i.e. it runs the risk of breaking the user's workflow if the positioning code misbehaves. If I'm missing something do let me know. 😀

2. Actor appearing mid-diagram:
	* Yes, this is valid, apparently! I guess it's used to represent an ephemeral actor which didn't exist until that point in time. I only noticed it when I was digging through the code, and I've never seen it used myself in the real world.
	* I included an example in my test diagram because I wanted to make sure that the basic floating header implementation wouldn't freak out if it came across one of these.
	* I agree that it would be great if these mid-diagram actors could also float. I don't see any blockers, implementation-wise, but I do consider it a separate requirement for a future roadmap; the current branch caters for 99% of the use cases, and I'd like to get the baseline implementation merged into master first, then look to see if the demand is there.

3. Minification:
	* I agree that this would be a great addition. It's something that applies to the existing (non-sequence diagram) scripts too. 
	* Again, in the interests of limiting the scope of this initial version of the feature, I'll raise minification as a separate feature request to be addressed in the next phase.

4. URL-encoding:
	* This is just a consequence of the `<script>` tag's content being embedded in another XML document, so that the XML is well-formed despite the use of reserved XML chars in the JS code. We could use a `CDATA` block instead, but it's the same correct escaping either way. I trust you didn't have to make any changes to the file after you unzipped it?

PR #1997, opened 2024-12-05
Extend error message with diagram type

travkin79:
Sometimes it's not obvious why something is considered being a syntax error. We had some cases where it was helpful to know the diagram type that PlantUML assumed to detect in our PlantUML source code. Thus, we'd like to add that detail (the diagram type) to the error message. It would be even better if we had an error message pointing us to the `allowmixing` option for our example. Could we do that?

Hi @arnaudroques,
Could we also make PlantUML accept something like `rectangle rec121` or `package rec121` without the brackets `{ }` in all diagram types, incl. class diagrams? It seems that missing brackets are ok in, e.g., deployment (composite?) diagrams, but not ok in class diagrams.

Example: https://www.plantuml.com/plantuml/uml/TOvH3i8m34F_UmfVeJt1hXAL324bHKgoByJTOTK0ZMu_aMNBUPuyo69HIKZ6DQXTXV5M6Oy407lpuux9BW9kBTXcglnJ-Hi_qtOF2EMKKF9MMukpaJBwxU3nKpdTMlws0bN87NvNxceOkX8dRfT5vGK0

```plantuml
@startuml
 
rectangle rec1 {
    rectangle rec11 {
        rectangle rec111 {
          interface interface1
        }
    }
 
    rectangle rec12 {
        rectangle rec121
    }
}
 
frame frame1 {
 
    rectangle rec21 {
        rectangle rec211 {
            class someclass
        }
    }
}
 
@enduml
```

### DIFF
--- a/src/net/sourceforge/plantuml/command/PSystemCommandFactory.java
+++ b/src/net/sourceforge/plantuml/command/PSystemCommandFactory.java
@@ -44,6 +44,7 @@ import net.sourceforge.plantuml.AbstractPSystem;
 import net.sourceforge.plantuml.EmbeddedDiagram;
 import net.sourceforge.plantuml.ErrorUml;
 import net.sourceforge.plantuml.ErrorUmlType;
+import net.sourceforge.plantuml.TitledDiagram;
 import net.sourceforge.plantuml.core.Diagram;
 import net.sourceforge.plantuml.core.DiagramType;
 import net.sourceforge.plantuml.core.UmlSource;
@@ -120,7 +121,7 @@ public abstract class PSystemCommandFactory extends PSystemAbstractFactory {
 			return buildEmptyError(source, location, it.getTrace());
 		}
 		sys.makeDiagramReady();
-		if (sys.isOk() == false)
+		if (!sys.isOk())
 			return null;
 
 		return sys;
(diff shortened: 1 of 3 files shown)


arnaudroques:
> Sometimes it's not obvious why something is considered being a syntax error. We had some cases where it was helpful to know the diagram type that PlantUML assumed to detect in our PlantUML source code. Thus, we'd like to add that detail (the diagram type) to the error message. It would be even better if we had an error message pointing us to the `allowmixing` option for our example. Could we do that?

Yes, that's a great idea!
Your commits have set us on the right path, so I suggest you don’t waste more time on this: we’ll take it from here.



> Hi @arnaudroques,
> Could we also make PlantUML accept something like `rectangle rec121` or `package rec121` without the brackets `{ }` in all diagram types, incl. class diagrams? It seems that missing brackets are ok in, e.g., deployment (composite?) diagrams, but not ok in class diagrams.

Ok! Sorry for all those inconsistencies :-)
We’ll make sure to fix that too.

travkin79:
That's great, @arnaudroques,
Thank you very much!

arnaudroques:
> I'd also suggest to replace "description diagram" with "component diagram" since there is no UML description diagram type in the [UML 2.5.1 specification](https://www.omg.org/spec/UML/2.5.1/PDF) (see figure) and no such diagram type on plantuml.com. We could just rename the entry `DESCRIPTION` to `COMPONENT` in `UmlDiagramType` or we could add a human-readable label to each diagram type entry from `UmlDiagramType`.

Thanks for pointing that out! I've made the change in a quick and dirty way for now.

travkin79:
> Thanks for pointing that out! I've made [the change in a quick and dirty way](https://github.com/plantuml/plantuml/commit/8824e08a088c32250aef0e46567207120be9c5b9) for now.

Thank you, @arnaudroques.

PR #2009, opened 2024-12-16
Add possibility to use custom graphviz implementation

### LINKED ISSUE #2008: Add possibility to use custom graphviz implementation
Currently it is not possible to add custom implementation for Graphviz. I suggest to add possibility to add graphviz via service (I mean service will be found using ServiceLoader) in 

https://github.com/plantuml/plantuml/blob/1d6d965496122097ceb34b78dee62b3e351f39de/src/net/sourceforge/plantuml/dot/GraphvizUtils.java#L75

PavelTurk:
Fixes #2008

### DIFF
--- a/src/net/sourceforge/plantuml/dot/GraphvizUtils.java
+++ b/src/net/sourceforge/plantuml/dot/GraphvizUtils.java
@@ -5,12 +5,12 @@
  * (C) Copyright 2009-2024, Arnaud Roques
  *
  * Project Info:  https://plantuml.com
- * 
+ *
  * If you like this project or if you find it useful, you can support us at:
- * 
+ *
  * https://plantuml.com/patreon (only 1$ per month!)
  * https://plantuml.com/paypal
- * 
+ *
  * This file is part of PlantUML.
  *
  * PlantUML is free software; you can redistribute it and/or modify it
@@ -30,7 +30,7 @@
  *
  *
  * Original Author:  Arnaud Roques
- * 
+ *
  *
  */
 package net.sourceforge.plantuml.dot;
@@ -38,7 +38,9 @@ package net.sourceforge.plantuml.dot;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
+import java.util.Iterator;
 import java.util.List;
+import java.util.ServiceLoader;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


arnaudroques:
That's a nice move. I'm curious to see where this takes us :-)

However, I suggest that we work on a dedicated branch for this https://github.com/plantuml/plantuml/tree/custom_graphviz

Could you please update your pull request accordingly?

arnaudroques:
> com.foo.TempGraphviz;

Are you considering committing this (or other classes) to the codebase?

arnaudroques:
> No, we will create our own project that uses graphviz-java, but it will be under ASL.
> 
Understood, no problem. Tell us when you will have an URL for this project!

> The reason is that we don’t want to be tied to PlantUML and prefer to make releases whenever we need.
>
That makes perfect sense.

> But that’s exactly the beauty of the solution I’m proposing. From now on, everyone will be able to use PlantUML with the specific version of Graphviz that suits them. Isn’t that wonderful and doesn’t it provide incredible flexibility to PlantUML?
> 
Absolutely! We’re ready to merge your pull request into the main branch. However, I do have a concern. The interface `net.sourceforge.plantuml.dot.Graphviz` is not particularly well-designed at the moment. It is subject to changes in the future, which could potentially break your integration.

On a related note, we’ve been considering refactoring parts of our code to make it cleaner. Specifically, we might take advantage of `java.util.ServiceLoader` to improve some of our existing implementations.


> For example, we are working on a program that draws graphs. Everyone who downloads the program should have the ability to use this feature. Of course, telling everyone to install a specific Graphviz version isn’t a great approach. We tried Smetana, but it has some unpleasant bugs. Today, I tested ELK and PlantUML/Vizjs. Vizjs threw a bunch of errors. ELK worked, but the results are still not as good as with Graphviz. That’s why we decided to use graphviz-java.
>

Thanks for sharing! Since we’re on the topic of Graphviz implementations, you might want to explore https://github.com/jamisonjiang/graph-support
It’s a minimalist Graphviz re-implementation in pure Java with almost no external dependencies. This might be worth considering for your use case.

arnaudroques:
> Now, I am thinking - maybe it is better to load not `Graphviz`, but `GraphvizFactory`. I think it will provide more flexibility. I mean:
> 
> ```
> Iterator<GraphvizFactory> iterator = ServiceLoader.load(GraphvizFactory.class).iterator();
> if (iterator.hasNext()) {
>     Graphviz graphviz = iterator.next().create();
>     Log.info("Using service provider " + graphviz.getClass().getName());
>     return graphviz;
> }
> ```
> 
> What do you think? I can make PR.

Yes, this does seem like a more flexible and improved solution.
Please feel free to create a new PR or update the current one. We look forward to reviewing your contribution!



> graphviz244onWindows()

This is quite outdated at this point and should probably be removed entirely.
We’ll handle the removal, but we’d prefer to wait for your PR first to streamline the process.

PR #2012, opened 2024-12-17
Add possibility to create Graphviz instance using service factory

PavelTurk:
Important. If a decision is made to make PlantUML an explicit JPMS module (by adding a `module-info`), then `module-info` must specify `uses net.sourceforge.plantuml.dot.GraphvizFactory;` or it will not work.

### DIFF
--- a/src/net/sourceforge/plantuml/dot/GraphvizFactory.java
+++ b/src/net/sourceforge/plantuml/dot/GraphvizFactory.java
@@ -1,0 +1,54 @@
+/* ========================================================================
+ * PlantUML : a free UML diagram generator
+ * ========================================================================
+ *
+ * (C) Copyright 2009-2024, Arnaud Roques
+ *
+ * Project Info:  https://plantuml.com
+ *
+ * If you like this project or if you find it useful, you can support us at:
+ *
+ * https://plantuml.com/patreon (only 1$ per month!)
+ * https://plantuml.com/paypal
+ *
+ * This file is part of PlantUML.
+ *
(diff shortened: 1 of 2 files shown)


arnaudroques:
> Important. If a decision is made to make PlantUML an explicit JPMS module (by adding a `module-info`), then `module-info` must specify `uses net.sourceforge.plantuml.dot.GraphvizFactory;` or it will not work.

Since we still want to support Java 8, I am wondering if this is truly feasible. Any thought?

arnaudroques:
> Therefore, the question is whether you are willing to drop support for Java 8.

At this time, we still have a significant number of users relying on Java 8, so we aim to continue supporting it for as long as possible. Given this, we’ve decided to postpone introducing `module-info` for now.  

PS: Your contribution has been merged via [this pull request](https://github.com/plantuml/plantuml/pull/2013). Thank you for your effort and valuable input!  

By the way, is there any other refactoring or additional improvement you'd like us to consider at this stage? We're happy to collaborate further.

arnaudroques:
Thank you so much for taking the time to share these thoughtful ideas. They are very interesting and certainly provide valuable insights for the project’s evolution.

That said, we intend to approach these changes with caution and a conservative mindset. For now, we have decided to stick with a single jar structure, as it remains the most convenient approach for the current needs of the project. However, we fully acknowledge the potential benefits of modularization and will keep your suggestions in mind for future discussions and planning.

On our end, we are indeed giving serious thought to `ServiceFactory` and its possible implementations, but we want to proceed very gradually to ensure stability and maintainability.

PR #2024, opened 2024-12-30
Display participant name when hovering over lifeline

michaelomichael:
As per [this suggestion](https://github.com/plantuml/plantuml/pull/1995#issuecomment-2525060195) by @ioplker, it uses the SVG `<title>` element to help users of large sequence diagrams to see which participant belongs to a given lifeline without having to scroll up to view the header. Making the headers "float" is a much nicer user experience (can see all participant names at once, and no need to wait several seconds for a tooltip to appear) but requires the special `!pragma svginteractive true` directive to be added, whereas this `<title>` approach can be enabled by default for all SVG sequence diagrams.

The implementation includes a new `<rect>` behind each participant's vertical line, which acts as a hover target for the mouse. (Trying to hover over a line that's a 1px wide is not practical.)

In `SvgGraphics`, the new `<rect>` cannot have a `fill="none"` attribute, as this makes it ineligible for mouse detection, so I have changed this to `fill="transparent"` for all cases. I also removed a number of explicit branches for `#00000000` fill colour value; these branches were already redundant due to the value being changed by the existing `fixColor()` method.

This initial implementation deals only with the lifeline itself. @ioplker has also suggested the possibility of adding tooltips to _messages_ between participant, e.g. "Message from alice to bob". I've held off on this as it would potentially add quite a lot of extra elements (and therefore bytes) to every SVG sequence diagram, and could be a future PR if it is desired.

### DIFF
--- a/src/net/sourceforge/plantuml/StringUtils.java
+++ b/src/net/sourceforge/plantuml/StringUtils.java
@@ -231,6 +231,10 @@ public class StringUtils {
 		return result.toString();
 	}
 
+	public static String toTooltipText(Display display) {
+		return (display == null || display.size() == 0) ? "" : display.get(0).toString();
+	}
+
 	public static String manageArrowForSequence(String s) {
 		s = s.replace('=', '-').toLowerCase();
 		return s;
--- a/src/net/sourceforge/plantuml/klimt/UGroupType.java
+++ b/src/net/sourceforge/plantuml/klimt/UGroupType.java
@@ -36,5 +36,5 @@
 package net.sourceforge.plantuml.klimt;
 
 public enum UGroupType {
-	ID, CLASS
+	ID, CLASS, TITLE
 }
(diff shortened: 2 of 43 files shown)


arnaudroques:

We sincerely apologize for the state of the code for sequence diagrams. It is quite old and not always easy to follow. The concept of "ComponentRose" is outdated and originates from a time when alternative Java implementations for these components were supported. This has since been replaced by "skinparam" and, more recently, by styles.  

As a result, a significant cleanup of this legacy code is necessary.

A big thank you for tackling this!

I made a minor adjustment related to the handling of the display, which you can see in this commit:  
[https://github.com/plantuml/plantuml/commit/498a9827b81e210e005b63635cc862f1108dd4bd](https://github.com/plantuml/plantuml/commit/498a9827b81e210e005b63635cc862f1108dd4bd)  

Thanks again!

Issue #1929, opened 2024-09-27
[Regex] Allow to expand a `[]` element with the corresponding alternative `|` element


The-Lum:
Hello PlantUML team,

**Is your feature request related to a problem? Please describe.**
- Could you allow to expand a `[]` element with the corresponding alternative `|` element?
_By a threshold, a parameter, ..._

**Describe the solution you'd like**
Here is an example, where `[z\\-]?` expanded would be equivalent to `(z|\\|-)?`:
```puml
@startregex
^[z\\-]?(z|\\|-)?$
@endregex
```
>[![](https://www.planttext.com/api/plantuml/svg/SoWkIImgAKfDJwtWYekkYedHZRNNgAg9YQdHrRHNuN98pKk1I7810000)](https://www.planttext.com/?text=SoWkIImgAKfDJwtWYekkYedHZRNNgAg9YQdHrRHNuN98pKk1I7810000) 

**Additional context**
Beware on some edge cases:
- `[+\\-]` with qualifier or Regex keyword as `+`
- `[^+\\-]` with negative parameter `^` 
- `[1-9]` with range...
- ...

_Original idea or observed on:_
- https://forum.plantuml.net/19269/error-generating-railroad-diagram-using-positive-lookbehind

Regards,
Th.

The-Lum:
Hello @arnaudroques 

As already mentioned:
> Beware on some edge cases:
`[^+\\-]` with negative parameter `^`
```puml
@startregex
[^+\\-]
@endregex
```


And also on escape char, as:
```puml
@startregex
title specialEscapes
[\t\r\n\a\e\f]
@endregex
```

And then, the first example is now a not working example, see:
```puml
@startregex
^[z\\-]?(z|\\|-)?$
@endregex
```

- Could you have a look?

Regards,
Th.

The-Lum:
Remains only issue on the `\\` management...

PR #2036, opened 2025-01-13
feat: Add method to map regex patterns to descriptive names

shunlibest:
Added a `getDrawValue` method that translates regex patterns (e.g., `\b`, `\d`, etc.) into descriptive names (e.g., `WordBoundary`, `Digit`). This improves readability and provides a clear mapping for regex special characters.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -51,7 +51,7 @@ dependencies {
     configurations.create("pdfJarDeps")
     "pdfJarDeps"(libs.fop)
     "pdfJarDeps"(libs.batik.all)
-    
+
 }
 
 repositories {
@@ -65,11 +65,12 @@ sourceSets {
 			srcDirs("src")
 		}
 		resources {
-			srcDirs("src")
+			srcDirs("src","resources")
 			include("**/graphviz.dat")
 			include("**/*.png")
 			include("**/*.svg")
 			include("**/*.txt")
+			include("resources/**/*.properties")
 		}
 	}
 	test {
@@ -77,9 +78,10 @@ sourceSets {
 			srcDirs("test")
 		}
 		resources {
-			srcDirs(".")
+			srcDirs(".", "resources")
 			include("skin/**/*.skin")
 			include("themes/**/*.puml")
+			include("resources/**/*.properties")
 		}
 	}
 }
@@ -103,16 +105,22 @@ tasks.withType<Jar>().configureEach {
     from("stdlib") { into("stdlib") }
     from("svg") { into("svg") }
     from("themes") { into("themes") }
+		from("resources") { into("resources") }
 
     // Add dependencies to the JAR
     val runtimeClasspath = configurations.runtimeClasspath.get().map { if (it.isDirectory) it else zipTree(it) }
     from(runtimeClasspath) {
         exclude("META-INF/*.SF", "META-INF/*.DSA", "META-INF/*.RSA") // Avoid conflict on signature
     }
-    
+
     duplicatesStrategy = DuplicatesStrategy.EXCLUDE
 }
 
+tasks.named<Copy>("processResources") {
+	from("resources") {
+		include("**/*")
+	}
+}
 
 publishing {
 	publications.create<MavenPublication>("maven") {
--- a/build.xml
+++ b/build.xml
@@ -111,6 +111,11 @@
 				<include name="**/*.css" />
 			</fileset>
 		</copy>
+		<copy todir = "build/resources">
+			<fileset dir = "resources">
+				<include name = "**/*.properties"/>
+			</fileset>
+		</copy>
 	</target>
 
 
--- a/plantuml-asl/build.gradle.kts
+++ b/plantuml-asl/build.gradle.kts
@@ -70,6 +70,7 @@ tasks.withType<Jar>().configureEach {
 	from("../stdlib") { into("stdlib") }
 	from("../svg") { into("svg") }
 	from("../themes") { into("themes") }
+	from("../resources") { into("resources") }
 	// source sets for java and resources are on "src", only put once into the jar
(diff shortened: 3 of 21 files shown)


The-Lum:
Thanks for this PR.

Here are some open questions...

- Perhaps some peoples preferred the old behaviour... then perhaps adding a parameter to activate/deactivate this functionality
- Then another question will be the localisation (l18n)... It will be perhaps usable to translate the description and create some minimal localisation as on the gant diagram or SVG export, in order to manage some other languages.

There are only open questions...
To exchange on this topic and to plan future PR... 

Regards,
Th.

arnaudroques:
Thank you for the effort and thought put into this PR. The proposal to translate regex patterns into descriptive names is indeed an interesting idea.  

That being said, I fully agree with the points raised by Th:  

1. **Default Behavior**  
   While the improvement adds value, it’s worth considering whether this should be the default behavior. Regex patterns such as `\b` and `\d` are standard and widely recognized among developers working with regex. Altering their display without explicit activation might not align with user expectations.  

2. **Configuration for Customization**  
   I believe it’s time to introduce parameters to offer display options for regex. For instance, we could allow users to activate or deactivate the descriptive name feature. A syntax like `(?#use descriptive names)` could also be a viable option to embed optional directive for regex displays, providing flexibility.  

3. **Localization (l18n)**  
   Supporting multiple languages is an excellent idea, especially if this feature is adopted more widely.  


I would love to hear everyone's thoughts on these suggestions and how we can collectively plan the next steps for this feature. Thanks again for this valuable contribution, and I appreciate the open discussion.

shunlibest:
@The-Lum @arnaudroques 
Thank you for your kind guidance and follow your advice. I added two options:

1. **useDescriptiveNames**: true/false. If that is empty, it defaults to false and will not display DescriptiveNames

2. **language**: i18n support was added to the project for the first time, and I'm not sure if it meets the specification.

Thanks

Here are some running screenshots:
image
image

image
image

The-Lum:
Hello @shunlibest, @arnaudroques, 

Good enhancement. 👍 
Another open questions... 😉

- For 2/
`language <xx>` is, _for example,_ conform to Gantt Diagram. ✔️ That is good.

- For 1/
But for `useDescriptiveNames [true|false]`

Beware of how to pass parameter to plantuml:
@arnaudroques:
- What is the best practice?
Between pragma or keyword?
_(cf. `!pragma compact` on EBNF diagram VS `mode compact` on Timing diagram)_

Open to debate...
Regards,
Th.

arnaudroques:
Yes, it might be time to standardize all of this.  

Regarding `!pragma`, it's meant to be an internal feature, not necessarily stable over time, so I’d lean towards saying no to using it for this purpose.  

What would you think about introducing a new directive at the preprocessor level?  

For example:  
```  
!option language cn  
!option useDescriptiveNames true  
!option mode compact  
```  

This could lead to something like:  
```  
@startregex  
!option language cn  
!option useDescriptiveNames true  

\b\B  
@endregex  
```  

From an implementation standpoint, these `!option` directives would be handled by the preprocessor. This approach would also allow us to introduce options for the preprocessor itself if needed in the future.  

What do you think?  

P.S.: A big thank you to @shunlibest for kickstarting the i18n efforts. We’ll likely apply the same principles to Gantt diagrams... The TODO list keeps growing! 😊

The-Lum:
Hi,

That seems good for a first beginning. 👍 

Regards,
Th.

arnaudroques:
@shunlibest, thank you once again for your valuable contribution!  

In the latest snapshot, we've made [a slight syntax adjustment](https://editor.plantuml.com/uml/SoWkIImgAKfDJwtWvLBCBoZ9pCzJo4dCIozDJ4zLI4b529OMfxga5YSNPGAvPQb-YRcfnGebHQL05RnSCKanJZ4fCIunvJ7XCSKnmRnS3gbvAL1J0G00) as shown below:  

```
@startregex
!option language de
!option useDescriptiveNames true

\b\B\d\D\w\W\s\S
@endregex
```

We'd love to hear your thoughts! Any feedback or suggestions are welcome.

shunlibest:
@arnaudroques 

I have reviewed your PR.I think your changes are wonderful and great. You've refactored a lot of code, which I'm not capable of doing.

This change, establish a unified standard, it is very important.

In the future, more experimental features can be added. More people can participate in plantUML, and plantUML will become more and more powerful.

Finally, for this useDescriptiveNames feature, I think it can be turned on by default after some time.

Thanks.

The-Lum:
Hello @shunlibest, @arnaudroques 

Here are some tests.
That seems good on all cases ✔️ 
```puml
@startregex
!option language de
!option useDescriptiveNames true

\d?\D+\w*\W{1,2}|\s.\S
@endregex
```
>[](https://editor.plantuml.com/uml/SoWkIImgAKfDJwtWKimlACdCprF8ISnBBqrCJrL8IOMBbHQdkgGM9nTb0hbbgNw9kQd52YL5fQbSN34fzZ4ksZ7bMZ7XrOOwHhKrCSLwCS5S3gbvAH1Z0G00)

But not on range: :x:
```puml
@startregex
!option language de
!option useDescriptiveNames true

\d\w|[\d\w0-9Z](\d|\w|[0-9]|Z)
@endregex
```
>[](https://editor.plantuml.com/uml/SoWkIImgAKfDJwtWKimlACdCprF8ISnBBqrCJrL8IOMBbHQdkgGM9nTb0hbbgNw9kQd52YL5fQbSN34fCUKrqI3IGDSoAbOZ9gK6903anDP4QN8vfEQbG8m70000)

- How to manage this case?
- We must investigate and make a proposal _(to allow descriptive names on range)_...

Link to #1929

_I can also update the doc. with this new feature (of descriptive names on REGEX diagram)..._
Regards,
Th.

PR #2058, opened 2025-01-29
Allow interactive SVG sequence diagram's header to float

michaelomichael:
Incorporates changes from #1995 and #2001, which were previously 
being worked on in a branch.

A summary of the overall changes:
* Button added to the SVG sequence diagram headers to allow the 
header to remain visible as the user scrolls vertically.
* SVG elements for messages and for each participant head or tail 
are now grouped together.
* Interactive SVG scripts now work when multiple such images are 
embedded in the same HTML page.

The above new behaviour is enabled per-diagram by adding 
`!pragma svginteractive true` to the start of the diagram 
description.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -79,6 +79,8 @@ sourceSets {
 		}
 		resources {
 			srcDirs(".", "resources")
+			include("svg/**/*.css")
+			include("svg/**/*.js")
 			include("skin/**/*.skin")
 			include("themes/**/*.puml")
 			include("resources/**/*.properties")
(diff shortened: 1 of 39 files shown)


arnaudroques:
Thanks, this is great! It's ready to merge indeed.

Maybe some minor refactoring later, but overall, very good work. Users will surely love it!

arnaudroques:
@michaelomichael We've made some minor refactoring in https://github.com/plantuml/plantuml/commit/54d534d7f318817a81d68f8edb73bb26fff203ec
Tell us if it does not work for you or if you think it was not a good idea, we are open to discussion!

michaelomichael:
It looks fine to me. 😄

Issue #2071, opened 2025-02-12
PlantUML v1.2025.0: Unable to Click Links in SVGs Due to Overlapping Transparent path Elements


RyanIkhlef:
**Describe the bug**

Hello,

I have a project that creates SVGs generated with PlantUML. Since version v1.2025.0, I'm facing a breaking change: I can't click on the links attached to my PlantUML rectangles. After a few analyses, I noticed that there are invisible PlantUML rectangles over my real PlantUML rectangles: they are `path` element. In this commit, https://github.com/plantuml/plantuml/commit/8534fd01b86165bcfaaf49180a2d71b6ea8b5811, the fill attribute was changed to `transparent` instead of `none`. This has a side effect because the mouse focuses on these elements, so we can't click on the link behind anymore.

**To Reproduce**

* Save this example an example.puml file :
    ```puml
    @startuml
    set separator none
    title =Test Diagram with not clickable areas"
    
    left to right direction
    
    skinparam {
      arrowFontSize 10
      defaultTextAlignment center
      wrapWidth 200
      maxMessageSize 100
    }
    
    skinparam rectangle<<Test 1>> {
      BackgroundColor #999999
      FontColor #ffffff
      BorderColor #6b6b6b
      roundCorner 20
      shadowing true
    }
    
    rectangle "==Test 1\nLorem ipsum dolor sit amet dolor amet eos dolor sanctus no sea stet.\nNonummy possim sed duo sit et quod erat sit sit eos amet consetetur dolor diam.\nEt dolore accusam sit sit lorem elitr ipsum sit. Labore praesent esse et erat amet facilisis aliquip delenit eos ut et dolore invidunt iusto aliquip tempor.\n Lorem ipsum et aliquam et dolores at." <<Test1>> as Test1 [[https://www.google.com]]
    rectangle "==Test 2\nLorem ipsum dolor sit amet dolor amet eos dolor sanctus no sea stet.\nNonummy possim sed duo sit et quod erat sit sit eos amet consetetur dolor diam.\nEt dolore accusam sit sit lorem elitr ipsum sit. Labore praesent esse et erat amet facilisis aliquip delenit eos ut et dolore invidunt iusto aliquip tempor.\n Lorem ipsum et aliquam et dolores at." <<Test2>> as Test2 [[https://www.google.com]]
[...]

Issue #2040, opened 2025-01-15
Since 2025.0, exporting a 'group' into svg results in a black box in Visio


gaspardpetit:
**Describe the bug**
The `group` directive generates a solid black square instead of displaying the group internal content when performing a svg export with version 2025.0

**To Reproduce**
Export the following to svg:

```puml
@startuml
group Group
A -> B : C
end
```

Open in Visio;
Observe that the group is rendered as a black box and content is not visible.

**Expected behavior**
Before,  2025.0, this resulted in a correctly formatted group when the svg was opened in Visio

**Screenshots**
![image](https://github.com/user-attachments/assets/43a44dea-1705-48e4-b45d-e67ad61c5ebe)

**Desktop (please complete the following information):**

- OS: Windows

**Additional context**

The issue seems to stem from a change in the svg `fill` directive. Before 2025.0, the `fill` would be set to `none`, since 2025..0, the `fill` is set to `transparent`:

Full svg file below:

# 2025.0 Version

```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
    contentStyleType="text/css" data-diagram-type="SEQUENCE" height="156px"
    preserveAspectRatio="none" style="width:104px;height:156px;background:#FFFFFF;" version="1.1"
    viewBox="0 0 104 156" width="104px" zoomAndPan="magnify">
    <defs />
    <g>
        <rect fill="transparent" height="48.7031" style="stroke:#000000;stroke-width:1.5;"
            width="88.9937" x="10" y="54.6094" />
        <g>
            <title>A</title>
            <rect fill="transparent" height="82.7031" width="8" x="27.6689" y="37.6094" />
            <line style="stroke:#181818;stroke-width:0.5;stroke-dasharray:5.0,5.0;" x1="31" x2="31"
                y1="37.6094" y2="120.3125" />
        </g>
        <g>
            <title>B</title>
            <rect fill="transparent" height="82.7031" width="8" x="61.0571" y="37.6094" />
            <line style="stroke:#181818;stroke-width:0.5;stroke-dasharray:5.0,5.0;" x1="64.3882"
                x2="64.3882" y1="37.6094" y2="120.3125" />
        </g>
        <rect fill="#E2E2F0" height="31.6094" rx="2.5" ry="2.5"
            style="stroke:#181818;stroke-width:0.5;" width="23.3379" x="20" y="5" />
[...]

PR #2072, opened 2025-02-14
Revert SVG fill to "none"; use UParam to indicate hit targets

michaelomichael:
Addresses issue #2071 while being being compatible with the changes made for #2024 and #2040.

I've reverted the default "transparent" SVG fill color to its original value of "none", and used the `UParam` mechanism to signal that the background color for the mouse-hover hit target rect in lifelines should use "fill-opacity".

Hopefully this is a valid use of `UParam`. @arnaudroques let me know if there is a more appropriate mechanism.

### DIFF
--- a/src/net/sourceforge/plantuml/klimt/color/HColor.java
+++ b/src/net/sourceforge/plantuml/klimt/color/HColor.java
@@ -2,7 +2,7 @@
  * PlantUML : a free UML diagram generator
  * ========================================================================
  *
- * (C) Copyright 2009-2024, Arnaud Roques
+ * (C) Copyright 2009-2025, Arnaud Roques
  *
  * Project Info:  https://plantuml.com
  * 
@@ -56,6 +56,11 @@ class Back implements UBackground {
 
 public abstract class HColor implements UChange {
 
+	public enum TransparentFillBehavior {
+		WITH_FILL_NONE,
+		WITH_FILL_OPACITY
+	}
+
 	public UBackground bg() {
 		return new Back(this);
 	}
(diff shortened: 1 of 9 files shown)


arnaudroques:
Thank you for the pull request.

While the proposed solution works, I suggest a different approach:

- **Make the following enum public:**

  ```java
  enum TransparentFillBehaviour {
      FILL_NONE,
      FILL_WITH_FILL_OPACITY
  }
  ```

- **In `HColor`, add the following methods:**

  ```java
  public TransparentFillBehaviour getTransparentFillBehaviour();

  public HColor withTransparentFillBehaviour(TransparentFillBehaviour newBehaviour);
  ```

  In my opinion, this is the core of the issue since it is indeed a property of the color.

- **In `SvgGraphic`, modify the method signature from:**

  ```java
  public final void setFillColor(String fill)
  ```

  **to:**

  ```java
  public final void setFillColor(String fill, TransparentFillBehaviour behaviour)
  ```


So, we would change the following line:

```java
ug = ug.apply(HColors.transparent());
```

to either:

```java
ug = ug.apply(HColors.transparent()).withTransparentFillBehaviour(TransparentFillBehaviour.FILL_WITH_FILL_OPACITY);
```

or:

```java
ug = ug.apply(HColors.transparent()).withTransparentFillBehaviour(TransparentFillBehaviour.FILL_NONE);
```

Or even, we could add a builder:

```java
HColors.transparent(TransparentFillBehaviour behaviour)
```


I'm not entirely sure, but I believe this approach would work. Do you understand my approach? What are your thoughts?

michaelomichael:
Yes, I think that could work. I'll give it a go.

arnaudroques:
Let's hope that all those "transparent" SVG issues are fixed now :-)

PR #2078, opened 2025-02-16
👷 fix `Linux ARM64` native image on `Native Image - Snapshot`

The-Lum:
Hi PlantUML team,

Fix #2077 issue.

_After some Copilot mistakes... He imagines a lot of wrong things..._

And the help of:
- https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/
- https://docs.github.com/fr/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories
- https://github.com/orgs/community/discussions/148648

Here is the ARM64 native image...

_FYI @ggrossetie_
Regards,
Th.

### DIFF
--- a/.github/workflows/native-image-snapshot.yml
+++ b/.github/workflows/native-image-snapshot.yml
@@ -33,7 +33,7 @@ jobs:
         include:
           - os: 'ubuntu-latest'
             platform: 'linux-amd64'
-          - os: 'ubuntu-latest'
+          - os: 'ubuntu-24.04-arm'
             platform: 'linux-arm64'
           - os: 'macos-latest'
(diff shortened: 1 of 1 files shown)


albert-github:
Note it looks like the changes for arm64 are only made for snapshots but not for the release, so there won't be arm64 releases.

The-Lum:
Hi @albert-github 

Yes, you are right.
This is only on testing period...

When it will be stable, we plan to do the same things on release delivries.

Regards,
Th.

Issue #1235, opened 2022-12-26
Interactive SVG with component ports is not working


lalmstrom:
While this works:
```puml
@startuml

!pragma svginteractive true

component c1
component c2
component c3

c1 --> c2 
c2 --> c3

@enduml
```
Hoovering component c3:
![image](https://user-images.githubusercontent.com/40788675/209489790-3b6d20e2-1460-4934-b681-b3c9ffef2b41.png)

Adding ports, does not:
```puml
@startuml

!pragma svginteractive true

component c1 {
  portout p1
}
component c2 {
  portin p2
  portout p3
}
component c3 {
  portin p4
}

p1 --> p2
p3 --> p4

@enduml
```

Hoovering component c3/port p4:
![image](https://user-images.githubusercontent.com/40788675/209489849-47fff1d2-f7f4-427a-8091-5e4ea8345c85.png)

kirchsth:
I saw the situation in other diagrams too. I think the main problem is 

a) that in the working version a component is converted into "element" with its alias,  and the entries can be selected via `class="elem"`

```xml
	<g class="elem c1 selected" id="elem_c1">
		<rect fill="#F1F1F1" height="46.2969" rx="2.5" ry="2.5" style="stroke:#181818;stroke-width:0.5;" width="57" x="7" y="8"/>
		<rect fill="#F1F1F1" height="10" style="stroke:#181818;stroke-width:0.5;" width="15" x="44" y="13"/>
		<rect fill="#F1F1F1" height="2" style="stroke:#181818;stroke-width:0.5;" width="4" x="42" y="15"/>
		<rect fill="#F1F1F1" height="2" style="stroke:#181818;stroke-width:0.5;" width="4" x="42" y="19"/>
		<text fill="#000000" font-family="sans-serif" font-size="14" lengthAdjust="spacing" textLength="17" x="22" y="40.9951">c1</text>
	</g>
```

b1) a component with a port is converted into a cluster without any `class="elem/cluster(?) c1 selected"` that it can be selected

```xml
	<!--cluster c1-->
	<g id="cluster_c1">
		<rect fill="none" height="101" rx="2.5" ry="2.5" style="stroke:#181818;stroke-width:1.0;" width="68" x="7" y="7"/>
		<rect fill="none" height="10" style="stroke:#181818;stroke-width:1.0;" width="15" x="55" y="12"/>
[...]

PR #2086, opened 2025-02-20
Make component ports interactive

michaelomichael:
As per issue #1235. The ports are now grouped and labelled in the same way as other entities in component diagrams, and will therefore participate in the SVG interactive behaviour.


The-Lum:
FYI @lalmstrom, @kirchsth.

PR #2113, opened 2025-03-04
fixes a hardcoded path in metadata generation task

asm0dey:


asm0dey:
Dear @The-Lum could you please run the build for me so we could see if they work?

arnaudroques:
> Dear @The-Lum could you please run the build for me so we could see if they work?

Done! Many thanks for your contributions.

asm0dey:
It seems to not work even on my machine, so nothing to thank for :(

The-Lum:
FYI, here is the corresponding run of the GA here:
- https://github.com/plantuml/plantuml/actions/runs/13656187888

PR #2114, opened 2025-03-04
Updating paths for test runs again, adding foreign functions support

asm0dey:


The-Lum:
FYI, here is the corresponding run of the GA here:
- https://github.com/plantuml/plantuml/actions/runs/13656781468

The-Lum:
And that seems KO ❌

asm0dey:
Aha, it seems pdf generation doesn't work not only on my machine, but on CI too. Any ideas why? It is supposed to be just a usual java run with an additional agent…

Issue #2122, opened 2025-03-12
Synchronizing qualifier behavior with its corresponding link during svg-interactive mode


takaaki-teshima:
In the latest version (1.2025.2), when svginteractive is set to true, qualifiers do not sync with their corresponding links, as shown below.

![Image](https://github.com/user-attachments/assets/d5856042-9f58-470b-b955-9ca7d78837a2)

I think qualifiers should sync with their links, as shown below. This can be done by placing qualifiers inside the same \<g\> tag as their corresponding link in the SVG data.

![Image](https://github.com/user-attachments/assets/1cc55a91-92bf-40a4-b18d-af17b7c57cf7)

Of course, any way that achieves this is fine. It would be great to see this synced behavior in future versions.

The-Lum:
Thanks for the report.
_FYI @michaelomichael_

The-Lum:
Hi all,

_Just for the record_, here is the corresponding input:

```puml
@startuml
!pragma svginteractive true
hide empty members
class A
class B
class C
class D

A [key1] *- "1" B : holds >
B [key2] o-- "1" C : holds >
C [key3] - "1" D : holds >
@enduml
@enduml
```
>[![](https://img.plantuml.biz/plantuml/svg/JSwn3eCW54RXdQVuwzZ4mNRhq2Zu5eQ1oesIWXekclZsJPN1wGpVScfE9gOrU75ReXc20M-3cn95qoUt4L9SIOpE4YWiQKUWyA78elU66Kskp5MvhH0DkY_jbSQzH54La7XXdBrbl8KyxA4nboUg2we3dnedjHUgQRB_rny0)](https://editor.plantuml.com/uml/JSwn3eCW54RXdQVuwzZ4mNRhq2Zu5eQ1oesIWXekclZsJPN1wGpVScfE9gOrU75ReXc20M-3cn95qoUt4L9SIOpE4YWiQKUWyA78elU66Kskp5MvhH0DkY_jbSQzH54La7XXdBrbl8KyxA4nboUg2we3dnedjHUgQRB_rny0)

[...]

PR #2124, opened 2025-03-13
Group link qualifiers (kals) with their edges

michaelomichael:
Fix for issue #2122

Example input:
```plantuml
@startuml
!pragma svginteractive true
hide empty members
class A
class B
class C
class D

A [key1] *- "1" B : holds >
B [key2a] o-- [key2b] C : holds >
C [key3] - "1" D : holds >
@enduml
```

Output:

https://github.com/user-attachments/assets/2b10c923-e87a-4d31-b500-b5dc43897b68

@arnaudroques and @The-Lum, my code change makes the assumption that a Kal always belongs to a link/edge - is this valid? The fact that the original code iterated over the _nodes_ in order to draw the Kals suggests that I might be missing some special case where the Kal should be displayed even if there is no corresponding link/edge.

Note that there is only a Svek version of the test, as Smetana doesn't appear to support link qualifiers yet.

### DIFF
--- a/src/main/java/net/sourceforge/plantuml/svek/SvekEdge.java
+++ b/src/main/java/net/sourceforge/plantuml/svek/SvekEdge.java
@@ -2,7 +2,7 @@
  * PlantUML : a free UML diagram generator
  * ========================================================================
  *
- * (C) Copyright 2009-2024, Arnaud Roques
+ * (C) Copyright 2009-2025, Arnaud Roques
  *
  * Project Info:  https://plantuml.com
  *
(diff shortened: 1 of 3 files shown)


arnaudroques:
> my code change makes the assumption that a Kal always belongs to a link/edge - is this valid?

Yes, this is valid

> The fact that the original code iterated over the _nodes_ in order to draw the Kals suggests that I might be missing some special case where the Kal should be displayed even if there is no corresponding link/edge.

I think it's just that we have been lazy when implementing "Kal" :-)

PR #2127, opened 2025-03-14
Fix #2125: Wrap plantuml-epl.jar in an OSGi bundle and provide an Eclipse update site

### LINKED ISSUE #2125: Publish PlantUML EPL version as OSGi bundle, publish an Eclipse update site / p2 repository
When using PlantUML in Eclipse plug-ins, the PlantUML library (jar file) has to be wrapped in an Eclipse plug-in (or an OSGi bundle) and has to be provided on an Eclipse update site (p2 repository). Otherwise, Eclipse plug-ins cannot resolve the dependency to the PlantUML jar file.

It would be great, if the PlantUML EPL jar would be released together with a Eclipse update site. This way Eclipse plug-in developers would not have to integrate each new PlantUML version as a jar in a new version of their Eclipse plug-ins, e.g. in the [PlantUML Eclipse plug-ins](https://github.com/hallvard/plantuml) (see [discussion](https://github.com/hallvard/plantuml/discussions/166)).

Another option for Eclipse plug-in developers would be to use target definition files with the new Maven dependency resolution feature that automatically wraps jar files in OSGi bundles so that they can be used as a plug-in dependency (see [details](https://xn--lubisoft-0za.gmbh/en/articles/using-maven-artifacts-in-pde-rcp-and-tycho-builds/)). This option's drawbacks are:

- such target definitions have to be specified for each Eclipse project / feature, by many developers again and again.
- such wrapped bundles have to be created or updated for each new PlantUML library version.
- there is no easy way of updating PlantUML for all Eclipse plug-ins and to the same version.

Thus, I prefer wrapping each PlantUML EPL version in an OSGi bundle and publishing it on a public Eclipse update site.

travkin79:
WIP

### DIFF
--- a/plantuml-epl/build.gradle.kts
+++ b/plantuml-epl/build.gradle.kts
@@ -10,6 +10,7 @@ plugins {
 	java
 	`maven-publish`
 	signing
+	id("org.standardout.bnd-platform") version "3.1.0"
 }
 
 group = "net.sourceforge.plantuml"
@@ -31,7 +32,7 @@ dependencies {
 	testImplementation(libs.xmlunit.core)
 
 	implementation(libs.jlatexmath)
-	
+
     implementation(libs.elk.core)
     implementation(libs.elk.alg.layered)
     implementation(libs.elk.alg.mrtree)
@@ -81,7 +82,7 @@ tasks.withType<Jar>().configureEach {
     from(runtimeClasspath) {
         exclude("META-INF/*.SF", "META-INF/*.DSA", "META-INF/*.RSA") // Avoid conflict on signature
     }
-	
+
 	// source sets for java and resources are on "src", only put once into the jar
 	duplicatesStrategy = DuplicatesStrategy.EXCLUDE
 }
@@ -136,6 +137,34 @@ tasks.compileJava{
 
 tasks.named("sourcesJar"){
 	dependsOn(preprocessLicenceAntTask)
+}
+
+configurations.all {
+	attributes {
+		attribute(TargetJvmEnvironment.TARGET_JVM_ENVIRONMENT_ATTRIBUTE, objects.named(TargetJvmEnvironment::class, TargetJvmEnvironment.STANDARD_JVM))
+	}
+}
+
+platform {
+	// config for wrapping plantuml-epl.jar in an OSGi bundle for using it in Eclipse plug-ins
+	// see https://github.com/stempler/bnd-platform
+
+	determineImportVersions = true
+
+	// configure proper names for the p2 repository (Eclipse update site)
+	categoryId = project.group as String
+	categoryName = "PlantUML"
+
+	featureId = project.group as String + "." + project.name
+	featureName = "PlantUML SDK"
+	featureVersion = project.version as String
+	featureProvider = "Arnaud Roques"
+
+	defaultQualifier = ""
+	useBndHashQualifiers = false
+	useFeatureHashQualifiers = false
+
+	bundle(project.group as String + ":" + project.name + ":" + project.version as String)
 }
 
 publishing {


plantuml-epl/build.gradle.kts
```diff
@@ -10,6 +10,7 @@ plugins {
 	java
 	`maven-publish`
 	signing
+	id("org.standardout.bnd-platform") version "3.1.0"
```
The-Lum, line 13: Hi @travkin79 


Even it is use only on this file:
 - Can you prefer to use the version and plugin name on the file [`gradle/libs.versions.toml`]( https://github.com/plantuml/plantuml/blob/79553cb7fbc32984db8bf100627ece21854e1e8e/gradle/libs.versions.toml)?

By, for example:

```toml
[versions]
standardout-bnd-platform = "3.1.0"

[plugins]
standardout-bnd-platform = { id = "org.standardout.bnd-platform", version.ref = "standardout-bnd-platform" }

```

Regards,
Th.

travkin79: Hi @The-Lum,
Thank you for the hint. Good to know, you're specifying the version in one place.

Unfortunately, I found a few issues with the gradle plug-in (generated bundle leads to compile errors when used in an Eclipse plug-in, license details are missing,...). Thus, I'm working on another solution that does not need a gradle plug-in.

travkin79:
Hi @arnaudroques and @The-Lum,
I think, this PR is ready for review.

Since the gradle plug-in did not work as expected, my solution became somewhat more complex, but seems to work well. I did not test the release scripts.

In addition, in case you agree with this approach, we need some way of publishing the Eclipse update site. This PR automatically creates an update site for PlantUML with the given version, e.g. 1.2025.3. Eclipse update sites offer usually more than only one version. For that purpose so called composite update sites are used that combine a set of update sites in one single update site, e.g. offering all available bundle versions. I think, I could extend this PR to offer such a composite update site.

travkin79:
Hi @arnaudroques,

> I know I previously mentioned that we are fine with hosting plantuml-eclipse, but we would prefer to avoid cluttering the main repository, which should remain focused solely on the core JAR library.

I understand that. Usually, I'd prefer that way, too.

> Would you be open to moving your work to this new dedicated repository instead?
https://github.com/plantuml/plantuml-eclipse

I have doubts that building the OSGi bundle and eclipse update site would work in a separate git repository. But maybe, there are ways to make it work. Let's check the following.

- My PR takes advantage of automatically triggering the build process for the Eclipse projects in plantuml-eclipse. AFAIK, there is no way of triggering a GitHub workflow in another git repository. Do you know a work-around for that? My goal was to always automatically create a bundle and update site for Eclipse as soon as there is a PlantUML jar release.
- My PR takes advantage of having direct access to the current PlantUML (snapshot) release version (since the Gradle build script is triggering the build) that I write into several plantuml-eclipse project files. I would need to read the version from another repository / release somehow.
- I hoped to simultaneously release the PlantUML jar with the corresponding bundle (and update site) for Eclipse, since, in essence, the bundle just wraps the original jar and makes it usable in Eclipse. No additions to the jar. But I can live with a separate release (for another git repository) that is created as soon as the is a PlantUML jar release.
- We need some place (URL and web space) to publish an Eclipse update site with the PlantUML bundles for Eclipse.

Do you think, we can somehow solve the challenges listed above?

arnaudroques:
Thank you for clearly outlining your concerns. Here are my thoughts on addressing each of them:

### Triggering Workflows Across Git Repositories
Directly triggering a GitHub workflow from another repository is not natively supported. However, we can effectively address this challenge using GitHub Actions and repository dispatch events:

- **Repository Dispatch Event:** Leverage GitHub's "repository dispatch" feature, where a release in the PlantUML repository can automatically trigger a build workflow in the `plantuml-eclipse` repository through a GitHub API call.
- **Webhook Integration:** Alternatively, webhooks or external CI/CD platforms integrated with GitHub Apps can also facilitate communication between repositories.

Initially, you could manually copy the latest PlantUML (snapshot) release version. Meanwhile, @The-Lum and I will collaborate on automating the synchronization between the two repositories. This approach allows you to focus primarily on the Eclipse site build.

Additionally, would you consider using Gradle instead of Maven? Since we're already utilizing Gradle, adopting it across the projects could ensure consistency and simplify maintenance.


### Regarding Hallvard's PlantUML Eclipse Plug-ins
Transferring Hallvard's Eclipse plugins to the official `plantuml` organization seems highly beneficial. I will reach out to Hallvard for his approval, though I anticipate no objections. Upon confirmation, relevant files from Hallvard's repository could be directly reused or integrated, further simplifying our workflow.

Please let me know your thoughts or if any additional clarification is needed.

arnaudroques:
Ok,regarding the build system, let's stick to Maven.

As for the repository structure, I would suggest that we start by consolidating both the PlantUML (EPL) OSGi bundle (with update site) and Hallvard's Eclipse plug-ins into a single repository, if feasible. This would simplify the overall management, automation, and visibility in the first instance. If we later identify clear reasons or practical challenges that justify separating them, we can always reconsider splitting them into two repositories.

travkin79:
Ok, then I suggest to transfer Hallvard's git repo to PlantUML in a first step and then add the new Eclipse projects for wrapping the PlantUML jar file in an OSGi bundle.

arnaudroques:
> Ok, then I suggest to transfer Hallvard's git repo to PlantUML in a first step and then add the new Eclipse projects for wrapping the PlantUML jar file in an OSGi bundle.

Great news, we've received confirmation from Hallvard, so we can proceed!

Regarding the transfer, I would suggest we **copy** the relevant files from Hallvard's repository into [plantuml-eclipse](https://github.com/plantuml/plantuml-eclipse) as a first step. I realize that by doing so, we will lose some historical data (such as commit history, issues, etc.), but this approach also offers a clean starting point, making it easier to reorganize or modernize if needed. However, this is only a suggestion: you can work as you wish.

Since you have admin rights on [plantuml-eclipse](https://github.com/plantuml/plantuml-eclipse), feel free to structure the repository however you see fit and run any experiments you have in mind. This is also one of the benefits of keeping it in a dedicated repository.

On our side, @The-Lum and I will probably set up a simple "HelloWorld" GitHub Action in the new repository. The idea is to trigger it whenever a new release or snapshot is published on [plantuml](https://github.com/plantuml/plantuml). For now, it will just output a basic "HelloWorld" message, serving as a starting point for further automation.

travkin79:
Hi @arnaudroques,

> Great news, we've received confirmation from Hallvard, so we can proceed!

Yes, that are great news! Let's proceed.

> Regarding the transfer, I would suggest we copy the relevant files from Hallvard's repository into [plantuml-eclipse](https://github.com/plantuml/plantuml-eclipse) as a first step.

I was also thinking about copying the files as a first step (for experimenting with the repo setup), but would prefer to eventually transfer the project's ownership in order not to lose any issues, history, etc. Do you think that would be a good approach?

> Since you have admin rights on [plantuml-eclipse](https://github.com/plantuml/plantuml-eclipse), feel free to structure the repository however you see fit and run any experiments you have in mind. This is also one of the benefits of keeping it in a dedicated repository.

Great. Thank you. 👍 

> On our side, @The-Lum and I will probably set up a simple "HelloWorld" GitHub Action in the new repository. The idea is to trigger it whenever a new release or snapshot is published on [plantuml](https://github.com/plantuml/plantuml). For now, it will just output a basic "HelloWorld" message, serving as a starting point for further automation.

Thank you very much for your help, @arnaudroques and @The-Lum! That's one of the most important goals to make this work. 👍

arnaudroques:
> I was also thinking about copying the files as a first step (for experimenting with the repo setup), but would prefer to eventually transfer the project's ownership in order not to lose any issues, history, etc. Do you think that would be a good approach? (An ownership transfer would require Hallvard to perform the transfer steps and he would need permission to create git repositories in the plantuml organization.)

I agree, for now, let's proceed by copying the files as a first step.
While it’s true that we will lose some history and existing issues, copying also allows us to start from a clean base, which can be beneficial at this stage, especially as we experiment with the repository setup.

The ownership transfer might be a bit more complicated to coordinate, so I suggest we revisit that option later, once we have things more stabilized.

Does that sound reasonable to you?

travkin79:
Hi @arnaudroques and @The-Lum,
I pushed all the non-obsolete git branches to the new plantuml-eclipse git repository and renamed the old master branch to main. I also added the GitHub workflow from your commit, @arnaudroques.

I guess, we have to change the project's license from MIT to EPL 1.0 as it originally was in Hallvard's repo and needs to be for the usage in Eclipse projects.

I think, I'll start adding my OSGi wrapper and update site projects in a separate git branch, soon.

Should we also activate GitHub discussions? And maybe discuss the remaining tasks there in a new discussion?

arnaudroques:
> Hi @arnaudroques and @The-Lum, I pushed all the non-obsolete git branches to the new plantuml-eclipse git repository and renamed the old master branch to main. I also added the GitHub workflow from your commit, @arnaudroques.
> 

Great!

> I guess, we have to change the project's license from MIT to EPL 1.0 as it originally was in Hallvard's repo and needs to be for the usage in Eclipse projects.
> 

Sure, you can change the license.

> I think, I'll start adding my OSGi wrapper and update site projects in a separate git branch, soon.
> 
> Should we also activate GitHub discussions? And maybe discuss the remaining tasks there in a new discussion?
>

Yes, you are right! Since you are admin you can go ahead :-)

So I close this thread, and let's continue our exchanges on https://github.com/plantuml/plantuml-eclipse

Many thanks for you work!

PR #2136, opened 2025-03-26
💄 Allow style on rnote and hnote, on sequence diagram

### LINKED ISSUE #340: skinparam on rnote and hnote
It could be nice to have the possibility to tweak the style of `rnote` and `hnote` separatedily of `note`.

My use case : 
- I use `rnote` to display the body of requests
- I use `hnote` to display the time that can spend a process
- I use `note` to ask question or give information to my reader

I would like to do this: 
```
skinparam rnote {
    BackgroundColor #101010
    FontColor #FAFAFA
}
[...]
rnote left of actor: ""field1 = valA""
note is it clear for everyone ?
```

instead of this:
```
skinparam note {
    BackgroundColor<<body>> #101010
    FontColor<<body>> #FAFAFA
}
[...]
rnote <<body>> left of actor: ""field1 = valA""
note is it clear for everyone ?
```

The-Lum:
Hello PlantUML team, @arnaudroques,

Here is a PR in order to:
 - Allow style on `rnote` and `hnote`, on sequence diagram.

Fix #340

And to manage:
```puml
@startuml
<style>
rnote {
  BackgroundColor #101010
  FontColor #FAFAFA
}
hnote {
  BackgroundColor #3333CC
  FontColor #FAFAFA
}
</style>

actor -> director

rnote left of actor: "field1 = valA"
hnote left of actor: "blue note"
note right of actor: is it clear for everyone ?
@enduml
``` 
See also:
- https://github.com/plantuml/pdiff/pull/19

Regards,
Th.

### DIFF
--- a/src/main/java/net/sourceforge/plantuml/sequencediagram/NoteStyle.java
+++ b/src/main/java/net/sourceforge/plantuml/sequencediagram/NoteStyle.java
@@ -64,7 +64,12 @@ public enum NoteStyle {
 	}
 
 	public StyleSignatureBasic getDefaultStyleDefinition() {
+		if (this == NoteStyle.HEXAGONAL)
+			return StyleSignatureBasic.of(SName.root, SName.element, SName.sequenceDiagram, SName.note, SName.hnote);
+
+		if (this == NoteStyle.BOX)
+			return StyleSignatureBasic.of(SName.root, SName.element, SName.sequenceDiagram, SName.note, SName.rnote);
+
 		return StyleSignatureBasic.of(SName.root, SName.element, SName.sequenceDiagram, SName.note);
 	}
-
 }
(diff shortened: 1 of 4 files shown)


arnaudroques:
The issue is not related to your changes.

There is a mutability problem in the `Note.java` class.

The problem is that when the `NoteStyle` is modified, it also affects the `Style`.  
A commit was made to partially address this:  
https://github.com/plantuml/plantuml/commit/18624f5137a93311a2d2462e7127161de99d0221

However, that’s not enough, because the `setStereotype()` method also modifies the style.  
So, if `setNoteStyle()` is called after `setStereotype()` (which does happen in the code), it will overwrite the changes made by `setStereotype()`.

The proper solution would be to stop storing `private Style style` and instead recompute it dynamically when needed.  
I suggest we take care of this, as it's a bit tricky.

Does that work for you?

The-Lum:
>The proper solution would be to stop storing private Style style and instead recompute it dynamically when needed.
I suggest we take care of this, as it's a bit tricky.
>
>Does that work for you?

Yes, that works for me.
_I'll let you do it and on occasion I can help..._

Regards,
Th.

PR #2147, opened 2025-04-07
👷 Add `DEV_TEST` environment variable to avoid including all gradle projects during intermediate tests

The-Lum:
Hello PlantUML team, and @arnaudroques,

Here is a PR in order to:
- add `DEV_TEST` environment variable on `intermediate-test`
- no gradle project include when `DEV_TEST` is set 

Regards,
Th.

### DIFF
--- a/.github/workflows/intermediate-test.yml
+++ b/.github/workflows/intermediate-test.yml
@@ -12,6 +12,9 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v4
 
+    - name: Set environment variable
+      run: echo "DEV_TEST=true" >> $GITHUB_ENV
+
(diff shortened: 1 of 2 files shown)


arnaudroques:
May I suggest to use `isDevTest` without "not" in the name?
And reverse the logic accordingly:

```
val isDevTest = System.getenv("DEV_TEST") != null
...
if (isCiBuild && !isDevTest) {
```

The-Lum:
@arnaudroques :
> May I suggest to use `isDevTest` without "not" in the name? And reverse the logic accordingly:
> 
> ```
> val isDevTest = System.getenv("DEV_TEST") != null
> ...
> if (isCiBuild && !isDevTest) {
> ```

Done directly on the master branch. 👍 

Regards,
Th.

Issue #2137, opened 2025-03-27
[Test] Add more unit tests


The-Lum:
## Context
To add more "Good First Issue", and to help for new contribution.

Here are some areas of ​​improvement...

## Wanted feature
- Could you add more unit tests?

## For examples

On:
- https://github.com/plantuml/plantuml/tree/master/src/test/java/net/sourceforge/plantuml
- https://github.com/plantuml/plantuml/tree/master/src/test/java/net/sourceforge/plantuml/tim

Regards,
Th.

MuhammadEzzatHBK:
Hi @The-Lum, is there any coverage report that we can use for guidance? We would like to know which areas of the project need more coverage the most :)

The-Lum:
Hi @MuhammadEzzatHBK 

> Is there any coverage report that we can use for guidance? 

Currently there is no coverage report...
_But FYI the number of unit tests is currently very limited..._

Then before to use [`Sonar`](https://www.sonarsource.com/) with [`SonarCloud`](https://github.com/marketplace/sonarcloud) or other huge tools, perhaps we can just begin to add [`Jacoco`](https://www.jacoco.org/) on the plantuml repo...

Yet another `Good first issue`:
- "Add Jococo on plantuml repo."

Regards,
Th.

The-Lum:
Hello @MuhammadEzzatHBK, and all, 

:tada: 
With last PR, here is just a Jacoco Report of PlantUML for Code Couverage on this new page:
- https://plantuml.github.io/plantuml/jacoco

Have a good look,
Enjoy, _to find a good way to make unit tests..._
Regards,
Th.

MuhammadEzzatHBK:
Oh this looks perfect, will get to work.

npiper:
This PR starts to put the Jacoco report into a 'site' build target (HTML) so some extra reporting may be visible once merged.

There may be a way to push the HTML to a part of the official site or for maintainers , contributors to view.

https://github.com/plantuml/plantuml/pull/2497

The-Lum:
@npiper 

Thank a lot.

>There may be a way to push the HTML to a part of the official site or for maintainers, contributors to view.

I can take care of that once it will be in the `main` branch.

This could make a nice homepage for:
- https://plantuml.github.io/plantuml/jacoco
- https://plantuml.github.io/plantuml/javadoc

Regards,
Th.

npiper:
[...]

PR #2192, opened 2025-05-14
#2137 Adding UT coverage to EvalMath.java

MuhammadEzzatHBK:
This class verifies the correctness of mathematical expression evaluation, including basic operations, operator precedence, unary operators, and error handling.

Hi @The-Lum 
This is my first PR. I would like to work on adding UTs to this repo to build some rapport. I found preproc/EvalMath.java simple enough to use as a starting practice. So please let me know if there's anything that should've been done another way.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -55,6 +55,7 @@ dependencies {
     "pdfJarDeps"(libs.fop)
     "pdfJarDeps"(libs.batik.all)
 
+    testImplementation("junit:junit:4.13.2")
 }
 
 repositories {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -55,7 +55,6 @@ dependencies {
     "pdfJarDeps"(libs.fop)
     "pdfJarDeps"(libs.batik.all)
 
-    testImplementation("junit:junit:4.13.2")
 }
 
 repositories {
--- a/src/test/java/net/sourceforge/plantuml/preproc/EvalMathTest.java
+++ b/src/test/java/net/sourceforge/plantuml/preproc/EvalMathTest.java
@@ -34,6 +34,8 @@
 package net.sourceforge.plantuml.preproc;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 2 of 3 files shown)


build.gradle.kts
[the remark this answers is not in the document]
MuhammadEzzatHBK: Not sure indeed, I tried using assertEquals but I just couldn't till I added this dependency. I will take a second look and see what could be done. Thanks for pointing this out!

src/test/java/net/sourceforge/plantuml/preproc/EvalMathTest.java
```diff
@@ -0,0 +1,90 @@
...
+	@Test
+	public void testBasicOperations() {
+		assertEquals(5.0, new EvalMath("2+3").eval(), DELTA);
+		assertEquals(-1.0, new EvalMath("2-3").eval(), DELTA);
+		assertEquals(6.0, new EvalMath("2*3").eval(), DELTA);
+		assertEquals(2.0, new EvalMath("6/3").eval(), DELTA);
+	}
+
+	@Test
+	public void testPEMDAS() {
```
The-Lum, line 59: @MuhammadEzzatHBK:

Minor remark:
- What is the meaning of `PEMDAS`?

src/test/java/net/sourceforge/plantuml/preproc/EvalMathTest.java
```diff
@@ -0,0 +1,90 @@
...
+	public void testUnaryOperators() {
+		assertEquals(-5.0, new EvalMath("-5").eval(), DELTA);
+		assertEquals(5.0, new EvalMath("+5").eval(), DELTA);
+		assertEquals(8.0, new EvalMath("-(2-10)").eval(), DELTA);
+	}
+
+	@Test
+	public void testDecimalNumbers() {
+		assertEquals(3.14, new EvalMath("3.14").eval(), DELTA);
+		assertEquals(4.24, new EvalMath("2.12+2.12").eval(), DELTA);
```
The-Lum, line 76: Perhaps:
- Could you use `@ParameterizedTest` and [`@CsvSource`](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-CsvSource) in order to factorise some of tests?

_It depends of the taste of each people..._
Regards,

build.gradle.kts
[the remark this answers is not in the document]
The-Lum: See doc. here:
- https://docs.gradle.org/current/userguide/java_testing.html#using_junit5

src/test/java/net/sourceforge/plantuml/preproc/EvalMathTest.java
MuhammadEzzatHBK: It's the acronym from the order of mathematical operations in an expression. (P)arentheses, (E)xponents, (M)ultiplication and (D)ivision and (A)ddition and (S)ubtraction.

src/test/java/net/sourceforge/plantuml/preproc/EvalMathTest.java
MuhammadEzzatHBK: Will see what I can do!

build.gradle.kts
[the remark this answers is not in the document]
MuhammadEzzatHBK: Will take a look, thanks!

MuhammadEzzatHBK:
@arnaudroques @The-Lum Comments addressed. Thanks for the support! Please let me know if there's anything else we would like to modify!

src/test/java/net/sourceforge/plantuml/preproc/EvalMathTest.java
```diff
@@ -0,0 +1,87 @@
...
+	@CsvSource({"'-5', -5.0", "'+5', 5.0", "'-(2-10)', 8.0"})
+	void testUnaryOperators(String expression, double expected) {
+		assertEquals(expected, new EvalMath(expression).eval(), DELTA);
+	}
+
+	@ParameterizedTest
+	@CsvSource({"'3.14', 3.14", "'2.12+2.12', 4.24"})
+	void testDecimalNumbers(String expression, double expected) {
+		assertEquals(expected, new EvalMath(expression).eval(), DELTA);
+	}
```
The-Lum, line 75: _A last remark_

If we want to be minimalist and for factorization...

Perhaps could you factorise all `EvalMath` call with a `testFunction` like:
```java
double cut(String str) { // Class Under Test
	 return EvalMath(str).eval();
} 
```
Or
```java
double cut(String str) { // Class Under Test
	final EvalMath evalMath = new EvalMath(str);
	return evalMath.eval();
 }
```
Regards,
Th.

MuhammadEzzatHBK: @The-Lum, I'm assuming the suggestion is replacing the direct instantiation of EvalMath with a function that does that in every test. But I think we're just replacing a one-liner with another one liner, which doesn't feel very minimalistic. Please feel free to correct me if I'm wrong.

The-Lum:
That is now merged.

@MuhammadEzzatHBK 
Thanks for your contribution. 👍 

Regards,
Th.

PR #2212, opened 2025-06-04
[Salt Wireframes] Allow checkbox without text

baflo:
Hi there,

I want to suggest this change in the behavior of checkboxes and hope you find it useful.

Currently, when creating a checkbox without text, there's always the struggle that Salt interprets it as empty button. There's a workaround with non-visible characters, but since these are not on the keyboard and usually must be copy-pasted, it didn't seem appropriate to me.

This is the current behavior:

image

As you can see, a dot `.` immediately following the closing bracken `]` is already forcing it to be a checkbox instead of a button in the current behaviour, but it (the dot) prints in the wireframe diagram.

I suggest (and implemented) that this dot `.` immediately following the closing bracket is not printed in the wireframe diagram anymore and hence allowing to have a checkbox without any text.

This is what it would look like:

salt-wireframe

I decided for the dot `.`, because it is already used to create empty lines.

### DIFF
--- a/src/main/java/net/sourceforge/plantuml/salt/factory/ElementFactoryCheckboxOff.java
+++ b/src/main/java/net/sourceforge/plantuml/salt/factory/ElementFactoryCheckboxOff.java
@@ -70,7 +70,8 @@ public class ElementFactoryCheckboxOff implements ElementFactory {
 
 	private List<String> extracted(final String text) {
 		final int x = text.indexOf(']');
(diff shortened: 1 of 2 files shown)


arnaudroques:
Many thanks for your contribution.
It sounds more like a bug to us, so we propose a fix with https://github.com/plantuml/plantuml/commit/17a956879545820b7b2c1904a2432970d1e04ddc

Tell us if it's not working for you!

Thanks again,

baflo:
Even better, thank you :)
salt-wireframe

Issue #2170, opened 2025-05-01
Add Domain Driven Design "types" to class diagram


xenoterracide:
**Is your feature request related to a problem? Please describe.**

In order to more cleanly show a Domain Driven Design model I think the following should be added

* `aggregate` green (ar) isa entity which isa class
* `repository` these technically could be an interface or a class.
  * `repository class` green (r)
  * `repository interface` purple (r)
* `event`
* `identfier` green (id)

This additionally requires that circles allow "oval" shapes as I would like to have 2-3 letters be rendered in the spot of a circle now. Not taller, just allowing wider.

Issue #1438, opened 2023-05-29
Feature Request: abstract entity


christianhujer:
**Is your feature request related to a problem? Please describe.**
There can be situations where entities are abstract.

For design modelling, abstract entities can occur when working with Spring or Jakarta JPA.
For analysis modelling, abstract entities can occur when entities form a class hierarchy.

**Describe the solution you'd like**
I'd like to be able to model abstract entities in PlantUML class diagrams.

**Describe alternatives you've considered**
There are three alternatives that work, they're just all not as nice as having abstract entities:
* Use an `abstract class` and add the `<<entity>>` stereotype to it.
* Use an `entity` and add an `<<abstract>>` stereotype to it (which is less correct than the previous alternative).
* Ignore the whole `abstract` thing for entities.

**Additional context**
I wanted to model payments for a POS. A payment is an entity. I wanted to do this:

```plantuml
@startuml
abstract entity Payment {
    id: UUID<Payment>
    amount: MonetaryAmount
}
entity CardPayment extends Payment {
    lastFourDigitsOfCard: String
}
entity CashPayment extends Payment {
    drawerId: UUID<CashDrawer>
}
@enduml
```

PR #2232, opened 2025-06-12
Add Record and Data classes

Intybyte:
I used #1028 and #1056 as a base, Data classes are a mutable set of data, whereas record are immutable set of data, many programming language implemented the latter (Java, C#, etc.), the former is less used but still present (POJOs and kotlin's data classes).

Let me know if there are any needed change, or I missed something

### DIFF
--- a/src/main/java/net/sourceforge/plantuml/abel/LeafType.java
+++ b/src/main/java/net/sourceforge/plantuml/abel/LeafType.java
@@ -44,7 +44,7 @@ public enum LeafType {
 	EMPTY_PACKAGE,
 
 	ABSTRACT_CLASS, CLASS, INTERFACE, ANNOTATION, PROTOCOL, STRUCT, EXCEPTION, METACLASS, STEREOTYPE, LOLLIPOP_FULL,
-	LOLLIPOP_HALF, NOTE, TIPS, OBJECT, MAP, JSON, ASSOCIATION, ENUM, CIRCLE,
+	LOLLIPOP_HALF, NOTE, TIPS, OBJECT, MAP, JSON, ASSOCIATION, ENUM, CIRCLE, DATA_CLASS, RECORD,
 
 	USECASE, USECASE_BUSINESS,
 
@@ -68,6 +68,9 @@ public enum LeafType {
 
 	public static LeafType getLeafType(String type) {
 		type = StringUtils.goUpperCase(type);
+		if (type.startsWith("DATA"))
+			return LeafType.DATA_CLASS;
+
 		if (type.startsWith("ABSTRACT"))
 			return LeafType.ABSTRACT_CLASS;
 
@@ -84,7 +87,7 @@ public enum LeafType {
 		return this == LeafType.ANNOTATION || this == LeafType.ABSTRACT_CLASS || this == LeafType.CLASS
 				|| this == LeafType.INTERFACE || this == LeafType.ENUM || this == LeafType.ENTITY
 				|| this == LeafType.PROTOCOL || this == LeafType.STRUCT || this == LeafType.EXCEPTION
-				|| this == LeafType.METACLASS || this == LeafType.STEREOTYPE;
+				|| this == LeafType.METACLASS || this == LeafType.STEREOTYPE || this == LeafType.DATA_CLASS || this == RECORD;
 	}
 
 	public String toHtml() {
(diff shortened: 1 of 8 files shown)


arnaudroques:
Thanks, this sounds like a good idea.

However, I would have used "dataclass" instead of "data class" (for consistency with other type)

Which means change the regex `data[%s]+class` to `dataclass`

Does it make sense ?

src/main/java/net/sourceforge/plantuml/abel/LeafType.java
```diff
@@ -84,7 +87,7 @@ public boolean isLikeClass() {
 		return this == LeafType.ANNOTATION || this == LeafType.ABSTRACT_CLASS || this == LeafType.CLASS
 				|| this == LeafType.INTERFACE || this == LeafType.ENUM || this == LeafType.ENTITY
 				|| this == LeafType.PROTOCOL || this == LeafType.STRUCT || this == LeafType.EXCEPTION
-				|| this == LeafType.METACLASS || this == LeafType.STEREOTYPE;
+				|| this == LeafType.METACLASS || this == LeafType.STEREOTYPE || this == LeafType.DATA_CLASS || this == RECORD;
```
The-Lum, line 90: Hi @Intybyte 

Isn't there an `LeafType` missing to make `LeafType.RECORD`?

Regards,
Th.

src/main/java/net/sourceforge/plantuml/classdiagram/command/CommandHideShowByGender.java
arnaudroques, line 1: I may be wrong, but shouldn't we have `dataclass` here:
```
"GENDER", "(?:(class|object|interface|enum|annotation|data|record|abstract|[%pLN_.]+|[%g][^%g]+[%g]|\\<\\<.*\\>\\>)[%s]+)*?"), //
```

arnaudroques:
I may be wrong, but shouldn't we have `dataclass` here:
```
"GENDER", "(?:(class|object|interface|enum|annotation|data|record|abstract|[%pLN_.]+|[%g][^%g]+[%g]|\\<\\<.*\\>\\>)[%s]+)*?"), //
```

arnaudroques:

If possible, could you also add some basic examples to https://github.com/plantuml/pdiff ? This tool is used to manage non-regression tests.

If you're unsure how to do this, feel free to share your examples here in the thread, we'll be happy to add them to `pdiff` ourselves.



PS:
I think that this is not needed anymore in `LeafType.java`

```
		if (type.startsWith("DATA"))
			return LeafType.DATACLASS;
```

Intybyte:
Yeah without a reference I have no idea how to add my references there, however I can show some examples here for `record` and `dataclass` in java

These are the java classes:

The java code for the record
```java
public record Key(String namespace, String key) {
    public static final Pattern NAMESPACED_KEY_PATTERN = Pattern.compile("^[a-z0-9._-]+:[a-z0-9._-]+$");

    public Key(String namespace, String key) {
        this.namespace = namespace.toLowerCase(Locale.ROOT);
        this.key = key.toLowerCase(Locale.ROOT);

        if (!NAMESPACED_KEY_PATTERN.matcher(this.full()).matches()) {
            throw new IllegalArgumentException("Invalid NS key: " + this.full());
        }
    }

    public String full() {
        return namespace + ":" + key;
    }

    @Override
    public String toString() {
        return "Key[" +
                "namespace=" + namespace + ", " +
                "key=" + key + ']';
    }
}
```

To PlantUML
```puml
@startuml
record Key {
  - namespace : String
  - key : String
  + {static} NAMESPACED_KEY_PATTERN : Pattern
  + Key(namespace : String, key : String)
  + full() : String
  + toString() : String
}

@enduml
```

A java POJO (dataclass example)
```java
// lombok data annotation generates setters and getters
@Data public class AngleData {
    private double horizontalAngle;
    private double verticalAngle;
    private double additionalHorizontalAngle;
    private double additionalVerticalAngle;
}
```

To PlantUML

```puml
dataclass AngleData <<lombok>> {
  - aimingPitch : double
  - aimingYaw : double
  - aimingFinished : boolean
  - lastAimed : long
  - timestampAimingMode : long
}
```

arnaudroques:
For the record, done here https://github.com/plantuml/pdiff/commit/f2d71d23027573e4551a3ccf252dd7017e393e4b

The-Lum:
Hi @Intybyte, @arnaudroques 

_To anticipate new request..._

- Do you have any other keywords evolutions in mind for class diagrams?

Hints:
- #1438
- module _(and [Module System](https://openjdk.org/jeps/261))_
- ...

_Ref._:
- https://en.wikipedia.org/wiki/List_of_Java_keywords


Regards,
Th.

Intybyte:
> Hi @Intybyte, @arnaudroques
> 
> _To anticipate new request..._
> 
>     * Do you have any other keywords evolutions in mind for class diagrams?
> 
> 
> Hints:
> 
>     * [Feature Request: abstract entity #1438](https://github.com/plantuml/plantuml/issues/1438)
> 
>     * module _(and [Module System](https://openjdk.org/jeps/261))_
> 
>     * ...
> 
> 
> _Ref._:
> 
>     * https://en.wikipedia.org/wiki/List_of_Java_keywords
> 
> 
> Regards, Th.

I think the module keyword might be useful in the future for adding another layer of abstraction over packages

Also worth considering are sealed classes, though they’d require special handling to define permitted subclasses

The-Lum:
_[Just for the record]_

I found (also) this request:
- #2170

Regards,
Th.

PR #2234, opened 2025-06-13
Pico Web Server support for the /language endpoint

skuro:
In order to support the `pico` web server from Emacs `plantuml-mode`, I added the `/language` endpoint to the `PicoWebServer` class, together with a related test.

### DIFF
--- a/src/test/java/net/sourceforge/plantuml/picoweb/PicoWebServerTest.java
+++ b/src/test/java/net/sourceforge/plantuml/picoweb/PicoWebServerTest.java
@@ -34,6 +34,7 @@ public class PicoWebServerTest {
 		test_GET_svg();
 		test_POST_render();
 		test_unknown_path();
+		test_language();
 	}
 
 	//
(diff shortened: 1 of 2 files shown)


src/test/java/net/sourceforge/plantuml/picoweb/PicoWebServerTest.java
```diff
@@ -227,6 +228,13 @@ private static void test_unknown_path() throws Exception {
 		assert response.getHeaderField("Location").equals("/plantuml/png/oqbDJyrBuGh8ISmh2VNrKGZ8JCuFJqqAJYqgIotY0aefG5G00000");
 	}
 
+	private static void test_language() throws Exception {
+		HttpURLConnection response = httpGet("/language");
+		assert response.getResponseCode() == 200;
+		assert readStreamAsString(response.getInputStream()).equals("foo\n"
```
The-Lum, line 234: Hi @skuro 

- Can you confirm what `foo\n` corresponds to?

Regards,
Th.

Issue #1542, opened 2023-09-19
Building plantuml in 2023


blinxen:
### Discussed in https://github.com/plantuml/plantuml/discussions/1540

<div type='discussions-op-text'>

<sup>Originally posted by **blinxen** September 18, 2023</sup>
Am I correct to assume that `gradle` is the only supported build tool for plantuml?

The reason I am asking this is because I recently took over the maintenance of plantuml in Fedora. The packaged version for fedora is `1.2023.7`. Currently the sources for plantuml are taken from https://sourceforge.net/projects/plantuml/. The latest version there (`1.2023.7`) is not in sync with GitHub. Also, the sources on sourceforge always included a `build.xml` in the `tar` archive. Is that gone now?</div>

Creating this issue  because I think creating documentation on how to build plantuml would be a good idea.

arnaudroques:
Yes, right now `gradle` is the only supported build tool for plantuml. Is this an issue for you?

And plantuml is now fully hosted on github, so you really should switch from sourceforge.

And yes, creating documentation on how to build plantuml is a good idea, we are going to work on this :-)

Thanks!

arnaudroques:
A first draft here https://github.com/plantuml/plantuml/blob/master/BUILDING.md
Hope this helps!

blinxen:
> Yes, right now gradle is the only supported build tool for plantuml. Is this an issue for you?

Unfortunately yes :( `gradle` is not packaged for Fedora (long story) and therefore cannot be used to build Fedora packages. Previously `ant` was used to build `plantuml` in Fedora. Is it possible to bring that back? `maven` would be fine too.

HusseinKabbout:
>  I Hope this is fine for you.

That would be perfect!

> However, ant will probably only build the GPL version

The initial packager of `plantuml` packaged the `LGPL` version anyway.


Thanks for the fast responses :D!

blinxen:
> I Hope this is fine for you.

That would be perfect!

> However, ant will probably only build the GPL version

The initial packager of plantuml packaged the LGPL version anyway.

Thanks for the fast responses :D!

arnaudroques:
The build.xml file is back. Please test it now :-D

blinxen:
Thanks!! My local build works again :D. The documentation looks good too.

PR #2239, opened 2025-06-15
Fix `ant` build

blinxen:
I am the package maintainer of `plantuml` in Fedora and the latest version (1.2025.3) fails to build because of invalid paths in `build.xml`. With the changes in this PR the build succeeds. I also tried to improve the build file. Let me know if this is OK.

One thing that I should mention is that even with these changes, compiling fails because `src/test` is now also being compiled and the required depenencies are not in the classpath. For Fedora I think I will just disable them but if you want me to fix that too then I can look into it.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -28,115 +28,115 @@
     ========================================================================
 -->
 <project name="PlantUML" default="dist" basedir=".">
-	<description>
+    <description>
         PlantUML Build File
     </description>
 
-	<!-- Compile source code and copy necessary files -->
-	<target name="compile">
-		<!-- Prepare the build directory -->
-		<delete dir="build" />
-		<mkdir dir="build" />
+    <!-- Compile source code and copy necessary files -->
+    <target name="compile">
+        <!-- Prepare the build directory -->
+        <delete dir="build" />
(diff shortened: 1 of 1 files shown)


arnaudroques:
> One thing that I should mention is that even with these changes, compiling fails because `src/test` is now also being compiled and the required depenencies are not in the classpath. For Fedora I think I will just disable them but if you want me to fix that too then I can look into it.

We can safely remove the tests from the Ant compilation, since these tests are already handled by Gradle. Let’s keep things simple. Would you be able to propose a PR for this?

If you want to improve the Ant script further, feel free to do so. While Gradle is now the official and recommended way to build the project, we do not want to force anyone, having several options is always valuable.

Thanks again for your help!

blinxen:
> If you want to improve the Ant script further, feel free to do so. While Gradle is now the official and recommended way to build the project, we do not want to force anyone, having several options is always valuable.

Thanks for being so open about this :D. Really appreciate it!

Using Gradle is sadly not possible with Fedora. See the initial [ant support request](https://github.com/plantuml/plantuml/issues/1542).

PR #2243, opened 2025-06-16
✅ Added some UT for EvalBoolean

EternalGeniuz:
While studying the module, I found that an unopened parenthesis throws an exception, but an unclosed parenthesis does not and the expression is evaluated correctly. I am not quite sure if this is the correct behavior of the function, but I feel it is necessary to inform about it.

[Issue#2137](https://github.com/plantuml/plantuml/issues/2137)


EternalGeniuz:
Oops, one of the checks failed. Do I need to do something about it? Like, rewrite the code to be compatible with Java 8

PR #2269, opened 2025-07-08
Add workflow for triggering another workflow in plantuml-eclipse repo

travkin79:
This change should trigger https://github.com/plantuml/plantuml-eclipse/blob/main/.github/workflows/hello-world.yml as soon as a new PlantUML release (pre-release or release) is published.

### DIFF
--- a/.github/workflows/trigger-plantuml-eclipse-release.yml
+++ b/.github/workflows/trigger-plantuml-eclipse-release.yml
@@ -1,0 +1,34 @@
+name: Trigger release workflow in plantuml-eclipse project
+
+on:
+  # run this manually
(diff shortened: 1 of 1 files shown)


travkin79:
Hi @arnaudroques,
It seems, my new workflow was not triggered, since (as I now found out) the release event is only created when a release is created via one of the following. So I think, we have to explicitly call this workflow, similar to calling the native-image-release workflow. Do you want me to create another PR or do you prefer to do it yourself?

- The GitHub web UI
- The REST API
- The GraphQL API

arnaudroques:
It seems to be working now.
Many thanks for pointing us in the right direction!

PR #2270, opened 2025-07-08
Explicitly call the new workflow

travkin79:
Follow up to PR #2269. Explicitly calls the new workflow instead of relying on the release event that is not triggered through the ci workflow, as it seems.


arnaudroques:
Thanks for you work!

We've made a dummy commit: maybe the token was misnamed in our github configuration...
We will see :-)

arnaudroques:
It seems to be working now.
Many thanks for pointing us in the right direction!

travkin79:
Thank you @arnaudroques for fixing the secrets settings and for merging. I created a follow-up PR #2271 in order to pass the release version and some other details to the called workflow. Could you please take a look at it?

PR #2271, opened 2025-07-09
Add workflow parameters with release details

travkin79:
This is a follow-up to PR #2270. It passes some release details (version, ref, and snapshot flag) to the called workflow. In addition, it removes the obsolete, unused release event trigger.


arnaudroques:
It's not a big issue, but we're currently seeing the following:

```
github/workflows/ci.yml (Line: 365, Col: 17): Unexpected value 'true'
```

I'll let you take care of this to avoid any overlap in our work.

Please feel free to submit other PRs in the meantime, no need to worry.
At the end, we'll most likely do a squash commit.

Thanks again!

travkin79:
Ok, thank you very much @arnaudroques. I've created a PR #2272 to fix that. I think, there is a type cast missing.

PR #2273, opened 2025-07-09
Fix typo in json argument

travkin79:
This is a follow-up to PR #2272 and fixes json syntax in trigger-plantuml-eclipse-release.yml

### DIFF
--- a/.github/workflows/trigger-plantuml-eclipse-release.yml
+++ b/.github/workflows/trigger-plantuml-eclipse-release.yml
@@ -43,5 +43,5 @@ jobs:
             {
               "release": "${{ inputs.pom-version }}",
               "snapshot": "${{ inputs.snapshot }}",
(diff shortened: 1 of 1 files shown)


travkin79:
Hi @arnaudroques,
Thanks again for merging. Now, we have no syntax errors anymore, but I'd still like to fix some things. The values passed to the called workflow look a little odd:
image

- Do you have an idea why the version is only "..0-SNAPSHOT" and not the complete version number?
- Is there some variable that contains the git ref to the git tag for a release that I could use? Currently we only have the master branch in the ref variable.

arnaudroques:
I think odd values have been fixed in last commit :-)
You can check and propose other PR if needed!

Thanks again

travkin79:
Hi @arnaudroques,
That's great! Thank you very much for fixing that.

PR #2313, opened 2025-09-04
Add missing double dash un-escaping in SVG metadata

### LINKED ISSUE #2306: SVG exported Image cant be imported again
**Describe the bug**
In some rare cases we cant import the source code of an exported svg. This only happens with svgs. With pngs we dont see such behaviour. The error we get is the note that this url uses huffman encoding but we exported the picture with the plantuml server version 1.2025.3.
We have like 100 diagrams and this happens only with two diagrams. If we change only on character the export and import works flawlessly. 

**To Reproduce**
Really difficult because the diagram creation works without any issues and only one character change fixes this phenomena

**Expected behavior**
Working import and export of diagrams and source code

mrtzcspr:
As doubles dashes/hyphens are not allowed in xml comments, they are getting escaped to `- -`, but the reverse operation was missing when extracting the metadata again from an existing SVG file.

This patch is untested and I have near zero Java experience, but let's give it a try..

Fixes #2306


arnaudroques:
Thanks a lot for jumping in and proposing this fix! 🎉
Even if it's untested and you don’t have much Java experience, it's really appreciated, sometimes that’s exactly how progress starts. Let's give it a try and see how it goes!

We've replied in the issue with another solution to avoid inserting spaces in the first place, but we'll keep your patch since it can be useful for existing files.
Later, we might even go for a more brute-force approach where we simply remove all spaces before decoding.

And if you feel like it, you can even try to code the modification yourself, it's exactly in the same place in the code.

The-Lum:
Hi PlantUML team, @arnaudroques, @mrtzcspr 

Maybe instead of putting it directly in a comment we could encapsulate it in a [CDATA](https://developer.mozilla.org/en-US/docs/Web/API/CDATASection), and avoiding to escape `--`!

_Open to debate..._
Regards,
Th.

arnaudroques:
> Maybe instead of putting it directly in a comment we could encapsulate it in a [CDATA](https://developer.mozilla.org/en-US/docs/Web/API/CDATASection)

Funny, that's what proposed now in https://github.com/plantuml/plantuml/issues/2306
But I'm not sure it really works, and there are other alternatives!

mrtzcspr:
Thanks for the quick merge! I think it's important to fix this for existing files anyway.

Regarding other solutions: Could there be other edge cases that cause issues in svg files? Not having looked at how the encoding works, but could there e.g. be 3 dashes in a row? Or any other illegal string combinations?

PR #2368, opened 2025-10-04
Add node-based transition labels for state diagrams

davmf:
Implemented support for displaying transition labels as separate nodes in state diagrams using the [node] style keyword or skinparam stateDiagramEdgeLabelStyle node setting.

Changes:
- Added STATE_TRANSITION_LABEL leaf type to represent transition label nodes
- Created EntityImageTransitionLabel class to render labels as rounded rectangles
- Modified CommandLinkStateCommon to split labeled transitions into two links with an intermediate node when node style is enabled
- Updated WithLinkType and CommandLinkElement to recognize "node" as a valid arrow style keyword
- Font size matches standard arrow labels (FontParam.ARROW)

Usage:
  S1 -[node]-> S2 : event / action or globally: skinparam stateDiagramEdgeLabelStyle node


The-Lum:
Hi @davmf, @arnaudroques 

> [!TIP]
> Just for your information:
>
> Instead of having an _[old]_ `skinparam`:
> ```plantuml
> skinparam stateDiagramEdgeLabelStyle node
> ```
> Perhaps add a _[new]_ `option`, like:
> ```plantuml
> !option stateDiagramEdgeLabelStyle node
> ```
> Similar to:
> ```plantuml
> !option handwritten true
> ```
>
> See:
> - [plantuml/preproc/OptionKey.java](https://github.com/plantuml/plantuml/blob/master/src/main/java/net/sourceforge/plantuml/preproc/OptionKey.java)


Regards,
Th.

The-Lum:
Hello all,

_[Just for the record]_
See also this list of PR:
- #2369
- #2379

Regards,
Th.

PR #2389, opened 2025-10-17
Add support for rounded corners on ortho edges

davmf:
This commit implements a solution for adding rounded corners to orthogonal edges in PlantUML diagrams. The implementation uses SVG post- processing to detect corners and add smooth curves, working with many diagram types including state diagrams and class diagrams.

Radius configured globally via 'skinparam edgeCornerRadius <value>'

Usage example:
```plantuml
  @startuml
  skinparam linetype ortho
  skinparam edgeCornerRadius 15
  [*] --> A
  A --> B
  A --> C
  A --> D
  C <-- B
  @enduml
```


The-Lum:
Hi @davmf, @arnaudroques 

That a good improvement.

But for future improvements, don't forget to manage `edgeCornerRadius` as a style also...

```puml
<style>
  edgeCornerRadius 15
</style>
```

Regards,
Th.

arnaudroques:
Thank you for your PR, it's a very good idea.

We are not going to accept this PR as it is, and I'll explain why.
However, we will guide you on how to achieve the same goal differently.

## Why are we not accepting this PR?

The main drawback is that it only works for SVG exports. Other formats (PNG, EPS, etc.) would not benefit from this feature.

That said, it's great that you were able to make this modification on the SVG side, considering the code's limited documentation, that was not easy.
So now, let's look at how we can implement this feature across all formats.
But before that, I need to give you some historical context to help you better understand the codebase.

## A bit of history

When we launched PlantUML back in 2009 (!), we only supported sequence and class diagrams at first.
And PNG generation was the only available output format. At that time:

* Sequence diagrams used the standard `java.awt.Graphics2D` API to generate PNGs.
* Class diagrams were converted to `.dot` files in Java, and then rendered into PNG directly by GraphViz.

This worked, but it had a major drawback: since PNGs were generated either by GraphViz or Java depending on the case, it was hard to achieve a consistent look & feel across all diagram types.

Moreover, when we wanted to add SVG export, it was straightforward for class diagrams (GraphViz supports SVG output) but not for sequence diagrams (since Java does not natively support SVG export).

We were also unhappy with several aspects of `java.awt.Graphics2D`, this API is very old (Java 1.1?).
So in 2011 (based on the [September 8, 2011 news](https://plantuml.com/old-news)), we completely redesigned the architecture.

We did two major things:

1. **Created the UGraphic abstraction**
   This allows drawing shapes (`UShape`) without worrying about the actual output format.
   We have several implementations (`UGraphicG2d`, `UGraphicSvg`, `UGraphicEps`, …) depending on the format.

   * `UGraphicG2d` is based on `java.awt.Graphics2D`.
   * `UGraphicSvg` uses our own implementation `SvgGraphics`, equivalent to `Graphics2D` but for SVG.

   These layers are well separated: at the bottom, we have `Graphics2D` and `SvgGraphics` (independent from UGraphic).
   On top, UGraphic abstracts away the differences between formats.

   We also defined multiple shape types (`ULine`, `UEllipse`, `URectangle`, `UPolygon`, …).
   Each shape has dedicated drivers to render it in the proper format (for example, `DriverLineG2d`).

2. **Changed how we use GraphViz**
   Until 2011, GraphViz handled both layout calculation and rendering. Starting with the new architecture (called **Svek**), GraphViz was only supposed to handle **layout calculation**, while PlantUML handled rendering using UGraphic.

   This allowed many improvements and provided a uniform rendering across all diagram types.

   But how do we use GraphViz only for layout? The CLI options for exporting node/edge positions are incomplete.
   We found the simplest solution was to let GraphViz generate an **SVG**, which we then **parse** to extract positions.

   It's a hack, but effective: PlantUML generates a simplified `.dot` file, assigns unique color codes to elements, and during parsing we map elements by color and retrieve their positions.

So, SVG is used in two very different contexts inside PlantUML:

* **Communication with GraphViz** (in Svek classes).
* **Exporting as SVG** (via `UGraphicSvg` and `SvgGraphics`).

## Back to your modification

I hope the context is clearer now.

* On the **Svek side**, the `.dot` file is generated in `DotStringFactory.createDotString()`. This part should not be changed by your PR.
* The SVG (from GraphViz) is retrieved by `getSvg()` and then **parsed** by `solve()` in the same class.
  (We're not particularly proud of this old code, to be honest…)

Instead of modifying the SVG **generated by GraphViz** (as you did), the correct place is to adjust the **parsed result** (“resolution”).

For edges, this parsing happens in the loop:

```java
for (SvekEdge line : getBibliotekon().allLines())
    line.solveLine(svgResult);
```

This leads us to the `SvekEdge.solveLine()` method.
A `SvekEdge` represents an edge in Dot/GraphViz: a Bézier curve with optional labels, decorations, and so on.

The curve itself is represented by `DotPath` (via `toDotPath()`).

This class is key:

* It directly corresponds to a “path” returned by Dot.
* It even implements `UShape`, so it can be drawn by a UGraphic.
* It's mutable (with methods like `moveXXX()` and `simulateCompound()`).

The actual drawing happens in `SvekEdge.drawU()`.

## What does this mean in practice?

You can refactor your current `SvgPathRounder` logic into a new method in `DotPath`:

```java
DotPath roundOrthogonalPaths(double radius);
```

or

```java
void muteToRoundOrthogonalPaths(double radius);
```

(whichever is simpler for you).

This way, you no longer need to parse SVG at that stage.

When should this method be called?
In `SvekEdge.drawU()`, you'll see:

```java
DotPath todraw = dotPath.copy();
```

Since the path is slightly modified before drawing (endpoints adjustments, etc.), that's the right moment to call `roundOrthogonalPaths` or `muteToRoundOrthogonalPaths`.

And how do we know if the user requested it? Through:

```java
private final Pragma pragma;
```

We are gradually phasing out `skinparam`, so instead of:

```
skinparam edgeCornerRadius 15
```

it should become:

```
!pragma edgeCornerRadius 15
```

This way, the value is directly accessible in `SvekEdge`.

(By the way, the same should eventually apply to `skinparam linetype ortho`, but this message is already too long.)

## Conclusion

So, instead of patching SVG output, the right approach is to extend `DotPath` and integrate your rounding logic at the drawing stage, controlled by `!pragma`.

What do you think, does this make sense?

arnaudroques:
> Hi @davmf, @arnaudroques
> 
> That a good improvement.
> 
> But for future improvements, don't forget to manage `edgeCornerRadius` as a style also...
> 
> ```wsd
> <style>
>   edgeCornerRadius 15
> </style>
> ```
> 
> Regards, Th.

Yes, that's true, you're right, but we'll do that later on.

davmf:
Thanks @arnaudroques for the detailed explanation.  I'll take a look at modifying this over the next few days per your advice.  I also tried an alternative approach of adding rounded corners to graphviz itself (which _does_ work, though I am yet to submit a MR to graphviz), then having PlantUML process the svg from graphviz with the corners already rounded.  That became difficult, however.  I also assumed this solution would be less suitable as it would require a graphviz update.

davmf:
I have pushed an update hopefully per your request.  Works for PNG as well now.

arnaudroques:
> I have pushed an update hopefully per your request. Works for PNG as well now.

Many thanks for your contribution.
This is a great new feature!

PR #2398, opened 2025-10-22
✨ allow unit (s, m, h, d, y, Y) on timing scale as pixel

The-Lum:
Hello PlantUML Team,

Here is a PR in order to:
- allow unit (`s`, `m`, `h`, `d`, `y`, `Y`) on timing scale as pixel

_To see if we should not outsource all of this 'unit' management..._

Regards,
Th.

### DIFF
--- a/src/main/java/net/sourceforge/plantuml/timingdiagram/command/CommandScalePixel.java
+++ b/src/main/java/net/sourceforge/plantuml/timingdiagram/command/CommandScalePixel.java
@@ -72,7 +72,7 @@ public class CommandScalePixel extends SingleLineCommand2<TimingDiagram> {
 
 	@Override
 	final protected CommandExecutionResult executeArg(TimingDiagram diagram, LineLocation location, RegexResult arg, ParserPass currentPass) {
-		long tick = Long.parseLong(arg.get("TICK", 0)) * getUnitFactor(arg.get("UNIT", 0));
+		final long tick = Long.parseLong(arg.get("TICK", 0)) * getUnitFactor(arg.get("UNIT", 0));
 		final long pixel = Long.parseLong(arg.get("PIXEL", 0));
 		if (tick <= 0 || pixel <= 0)
 			return CommandExecutionResult.error("Bad value");


src/main/java/net/sourceforge/plantuml/timingdiagram/command/CommandScalePixel.java
```diff
@@ -67,7 +72,24 @@ private static IRegex getRegexConcat() {
 
 	@Override
 	final protected CommandExecutionResult executeArg(TimingDiagram diagram, LineLocation location, RegexResult arg, ParserPass currentPass) {
-		final long tick = Long.parseLong(arg.get("TICK", 0));
+		long tick = Long.parseLong(arg.get("TICK", 0));
+		final char unit = arg.get("UNIT", 0).toLowerCase().charAt(0);
```
arnaudroques: Did you test it? Because I think that if the user doesn’t specify a unit, `arg.get("UNIT", 0)` will return `null`.
And we’ll get a NPE.
Am I wrong?

The-Lum: You have right.

Perhaps I can test it with:
```
if (unit == null)
    unit = 's';
```

Are you ok ?

Regards,
Th.

arnaudroques: > Are you ok ?

No :-)
`unit` is a char and cannot be `null` :-)


```
final String unit = arg.get("UNIT", 0);
if (unit!=null) {
switch (unit.toLowerCase().charAt(0)) {
....
```

arnaudroques: I think it's better to build a dedicated method here:

```java
	private long getUnitFactor(String unit) {
		if (unit==null)
			return 1;
		switch (unit) {
		case "s":
			return 1;
		case "m":
			return 60;
		case "Y":
		case "y":
			return 3600L * 8766L; // 24 * 365.25 = 8766;
		}
		....
	}
```

and use it:

```java
final long tick = Long.parseLong(arg.get("TICK", 0)) * getUnitFactor(arg.get("UNIT", 0));
```

Does it sound better?

arnaudroques: > The subtleties of Java are starting to overwhelm me tonight...

_Sleep, you must. Debug tomorrow, you will._ ☕💫

src/main/java/net/sourceforge/plantuml/timingdiagram/command/CommandScalePixel.java
```diff
@@ -67,7 +72,7 @@ private static IRegex getRegexConcat() {
 
 	@Override
 	final protected CommandExecutionResult executeArg(TimingDiagram diagram, LineLocation location, RegexResult arg, ParserPass currentPass) {
-		final long tick = Long.parseLong(arg.get("TICK", 0));
+		long tick = Long.parseLong(arg.get("TICK", 0)) * getUnitFactor(arg.get("UNIT", 0));
```
arnaudroques: `tick` can be `final` now :-)

The-Lum: Done...

Issue #2380, opened 2025-10-14
Text on top of branch not appear if ConditionStyle InsideDiamond


Tan-AcamarVN:
**Describe the bug**
When I use the skinparam ConditionStyle InsideDiamond, the text on the branch line not appear

**To Reproduce**
Minimal code to reproduce the behavior:
```puml
@startuml
skinparam ConditionStyle InsideDiamond

repeat :SET INITIAL VALUES
repeat while (Start?) is (N) not (Y)
:Done;
@enduml
```
Comment out the skinparam let we see the text on the branch, but I want the condition block to be a diamond shape, not a hexagon.

**Expected behavior**
Should have the N and Y text on the branch.

**Screenshots**
In case of comment out the skinparam: 

<img width="295" height="312" alt="Image" src="https://github.com/user-attachments/assets/26242ead-cffa-4907-af41-cd30fb372c3c" />

In case of not comment out:

<img width="303" height="318" alt="Image" src="https://github.com/user-attachments/assets/61a11734-e5a6-4947-91e6-9fec1b58f133" />

**Desktop (please complete the following information):**

- OS: Window 10
- Browser: Google Chrome
- Version: tested on https://www.planttext.com/

AlexJauregui02:
Hi, I can try to implement this, do I need to be assigned?

arnaudroques:
> Hi, I can try to implement this, do I need to be assigned?

You can give a try without being assigned.
We won't work on this right now.
Thanks for trying :-)

The-Lum:
Hi all,

_[This is an Issue Review]_ 👀 
This is now fixed _(on V1.2026.0)_.

- Thanks @Tan-AcamarVN for the report.
- Thanks @AlexJauregui02 for the your PR.
- Thanks PlantUML team for your work.

Regards.

PR #2422, opened 2025-11-07
Added repeat while labels to diamond Inside figure

AlexJauregui02:
This pull request fixes the issue: #2380 
Added the labels to the diamond inside figure when it's repeat condition


arnaudroques:
Thanks for your contribution!

However, using [pdiff](https://github.com/plantuml/pdiff), I think there is a slight regression with the following example [o8eb0bca9dyghfsjbhgcr12isju0wd4
](https://github.com/plantuml/pdiff/blob/main/db/o8/o8eb0bca9dyghfsjbhgcr12isju0wd4.puml)


```
@startuml
skinparam style strictuml
skinparam conditionStyle InsideDiamond
start
:Activity 1;
repeat
    :Activity 2;
    :Activity 3;
repeat while (condition) is (yes) -[#black]-> yes
-> no;
:Activity 4;
stop
@enduml
```

Nothing urgent, though.
Could you have a look on it?
Thanks!

Issue #2343, opened 2025-09-17
connector missing at end of switch in swimlane


smfgh:
**Describe the bug**
Sometimes the last connector in a swimlane is missing for "case".

**To Reproduce**
There is a line missing:
```puml
@startuml
|S1|
|S2|

|S2|
  :preAct;
  switch (Test)
  case (S1!)
    :ActPre;
    |S1|
    :Act;
  case (S2!)
    |S2|
    :Act;
  endswitch
  |S2|
  :postAct;
@enduml
```

This line is correct drawn using "if" (but "if" is not an option here as there are other cases in the real process):
```puml
  :preAct;
  if (Test) then (S1!)
    :ActPre;
    |S1|
    :Act;
  else (S2!)
    |S2|
    :Act;
  end if
  |S2|
  :postAct;
```

<img width="221" height="623" alt="Image" src="https://github.com/user-attachments/assets/443182bc-5542-4653-9f8a-456df5b81fab" />

**Expected behavior**
Switch-case-connector should appear even used in swimlane.

**Screenshots**

**Desktop (please complete the following information):**
- PlantUML: 1-2025-7

AlexJauregui02:
Hi, I can try to tackle this problem

AlexJauregui02:
Hi saw that this problem repeat itself when trying to make a connection directly from a switch tile to another tile in a different swimlane. I found how to fix it, I'm still working on it

smfgh:
Thank you so much for your effort

AlexJauregui02:
The last pull request should fix this issue

The-Lum:
Hi all,

_[This is an Issue Review]_ 👀 
This is now fixed _(on V1.2026.0)_.
- Thanks @smfgh  for the report.
- Thanks @AlexJauregui02 for the your PR.
- Thanks PlantUML team for your work.

Regards.

PR #2423, opened 2025-11-11
Connectors from switch tile creates arrow to a different tile in a different swimlane

AlexJauregui02:
Connections that came from the switch didn't render the arrow if the destination is in another swimlane.

```puml
@startuml
|S1|
|S2|

|S2|
  :preAct;
  switch (Test)
  case (!S2)
  :ActprS3;
  case (Aba1)
  	|S1|
  	:ActPrS1;
  endswitch
  |S2|
  :postAct;
@enduml
```

Before:
Screenshot 2025-11-10 at 8 36 55 p m

After: 
Screenshot 2025-11-10 at 8 39 35 p m

This pull request is part of the issue: #2343


arnaudroques:
Many thanks for your fix!

You can also add some non-regression tests in the **pdiff** project https://github.com/plantuml/pdiff

That's exactly what we've just done https://github.com/plantuml/pdiff/commit/e614cf1f8a19f68cf301190267a82e25288dadac

This project is useful for detecting regressions when we make code changes, feel free to use it as well. The documentation should be clear enough to get you started, but don't hesitate to contact us if you need any help.

PR #2429, opened 2025-11-13
Create light distributions with light classifier

nbauma109:
I created a light distribution along side each of the existing ones with classifier light :

$ ls -lrt **/*-light.jar
-rw-r--r-- 1 nicol 197612  7298415 Nov 13 20:53 plantuml-asl/build/libs/plantuml-asl-1.2025.11beta3-light.jar
-rw-r--r-- 1 nicol 197612  7301911 Nov 13 20:54 plantuml-bsd/build/libs/plantuml-bsd-1.2025.11beta3-light.jar
-rw-r--r-- 1 nicol 197612 16700052 Nov 13 20:54 plantuml-epl/build/libs/plantuml-epl-1.2025.11beta3-light.jar
-rw-r--r-- 1 nicol 197612  7297973 Nov 13 20:55 plantuml-lgpl/build/libs/plantuml-lgpl-1.2025.11beta3-light.jar
-rw-r--r-- 1 nicol 197612  7300641 Nov 13 20:56 plantuml-mit/build/libs/plantuml-mit-1.2025.11beta3-light.jar

I built with:

```
set CI=true
gradlew clean build -x test -x javadoc
```

I tested `plantuml-mit-1.2025.11beta3-light.jar` with my application and it worked without breaking under my application.


arnaudroques:

We would prefer a different approach: adding a `plantuml-mit-light` directory that would provide the light distribution **only** for the MIT version.

**Why?**

* In the current proposal, too many additional JAR files are generated, and we already have more than enough.
* We would like this light variant to remain exceptional, so we don’t want to generalize it to all distributions.

Would you be able to update your PR accordingly?

Thank you and sorry for the inconvenience!

nbauma109:
ok, noted, I will create another branch with the new changes as master was force-pushed since then.

PR #2438, opened 2025-11-16
Add chart diagram support with comprehensive features

davmf:
This is a draft PR for a feature I have been wanting PlantUML to support for a while.  Initially inspired by Mermaid's support for basic line and bar charts (https://mermaid.js.org/syntax/xyChart.html), but extended to incorporate PlantUML's more comprehensive capabilities.  Could also be a basis for future work.

For documentation and examples, see https://github.com/davmf/plantuml/blob/feature/dmf/charts/docs/chart-diagram-documentation/chart-diagram-documentation.md

If you think this would be good to merge, let me know of any errors/omissions.

I was also thinking of adding some tests.

### DIFF
--- a/src/main/java/net/sourceforge/plantuml/PSystemBuilder.java
+++ b/src/main/java/net/sourceforge/plantuml/PSystemBuilder.java
@@ -109,6 +109,7 @@ import net.sourceforge.plantuml.timingdiagram.TimingDiagramFactory;
 import net.sourceforge.plantuml.utils.Log;
 import net.sourceforge.plantuml.version.PSystemLicenseFactory;
 import net.sourceforge.plantuml.version.PSystemVersionFactory;
+import net.sourceforge.plantuml.chart.ChartDiagramFactory;
 import net.sourceforge.plantuml.wbs.WBSDiagramFactory;
 import net.sourceforge.plantuml.wire.WireDiagramFactory;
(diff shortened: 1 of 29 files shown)


arnaudroques:
Honestly, this is excellent work and a very good idea!

Reviewing the code directly through GitHub is not very convenient, so I suggest working in a dedicated branch https://github.com/plantuml/plantuml/tree/chart 

Could you update the PR so that it targets this branch?

For now, my only comment is that in the headers of the new files you created, you should include your name or a pseudonym as the “Original Author”. I don't want to take credit for code I didn't write :-)

davmf:
Thanks, this has now been updated.  Yes it occurred to me that PlantUML is well suited to charts.

- There are limited options for drawing charts outside of more heavyweight tools such as matplotlib, tikz, gnuplot or even excel.
- Being text, PlantUML is well suited to chart generation by AI with limited token use and reproduceable results.
- Charts only need relatively simple specifications to get a good result.  The code can easily determine how to represent the data and there is minimal layout complexity.
- It would be simple to use the pre-processor to import data (eg. from csv) to add capability.

davmf:
I left the docs out of the PR to make it simpler.  The link to my own repo contains examples all generated using PlantUML and forms the basis for the documentation that would need to be added to the website and pdf.

PR #2511, opened 2026-02-02
group individual states in SVG

fcnjd:
Continuing the first PR from last week, this one adds groups for individual states in a state diagram as well, not just starting and ending ones. This is especially important so that entity IDs which are right now already present in links can correctly be resolved to their corresponding element.
As a side note, even with the newly introduced SVG scheme, IDs are in cases still assigned twice.


fcnjd:
Sorry for the confusion, the ID seemed to be a mistake on my side. Tested and ready for merge.

fcnjd:
@arnaudroques  Thank you very much for always merging those changes so quickly. That really helps a lot.

Issue #1523, opened 2023-08-29
Please add Protocol bit-structure support for Network diagram in PlantUML


liuhangbin:
**Is your feature request related to a problem? Please describe.**
The Protocol bit-structure is always used to describe what a packet structure looks like.  But I can't find a similar example in plantuml nwdiag. https://plantuml.com/nwdiag . Would you please help add this support?

**Describe the solution you'd like**
Here are what other tools look like

http://blockdiag.com/en/nwdiag/packetdiag-examples.html.
https://kroki.io/examples.html#packetdiag

**Describe alternatives you've considered**

**Additional context**

It also would be good if we could set the values for each field.  see the picture as an example.

![1](https://github.com/plantuml/plantuml/assets/1594572/62df8ba1-5db5-42ad-ad99-ab34746350cc)

liuhangbin:
Here is another example of the [bytefield](https://kroki.io/examples.html#bytefield)

The-Lum:
_[Just for traceability] See also similar request here:_
- https://forum.plantuml.net/8787/nwdiag-package-includes-scripts-called-rackdiag-packetdiag
- https://forum.plantuml.net/9841/protocol-bit-structure-tool
- https://forum.plantuml.net/16484/can-use-plantuml-for-describing-the-octet-layout-of-message

Regards,
Th.

liuhangbin:
> _[Just for traceability] See also similar request here:_
> 
> * https://forum.plantuml.net/8787/nwdiag-package-includes-scripts-called-rackdiag-packetdiag
> * https://forum.plantuml.net/9841/protocol-bit-structure-tool
> * https://forum.plantuml.net/16484/can-use-plantuml-for-describing-the-octet-layout-of-message

Yes, I also saw the requests in the forum. But I don't see progress on the forum. So I opened this request.

kolulu23:
I just started to work on this, adding both [rackdiag](http://blockdiag.com/en/nwdiag/rackdiag-examples.html) and [packetdiag](http://blockdiag.com/en/nwdiag/packetdiag-examples.html) from `nwdiag` into PlantUML.

However, I do think these two diagrams provide little features compared to other newly added diagram formats. For protocol protraiting diagrams, something like [bytefield](https://texdoc.org/serve/bytefield.pdf/0) seems to be more appealing.
[...]

PR #2520, opened 2026-02-11
Add new diagram type packetdiag

kolulu23:
Related issue: https://github.com/plantuml/plantuml/issues/1523


The-Lum:
Hello @kolulu23 

Just to make a todo list...
- What are the remains missing fonctionalities from the original project (in order to be backward compatible)?

Hints: 
- ~rotate~ [it seems that is already implemented :+1:]
- _tbc_

Then we will have a look on your PR.

Regards,
Th.

arnaudroques:
Thanks you for your work!

Could you target https://github.com/plantuml/plantuml/tree/packetdiag branch instead of "master" for this PR : it would help us to review your code.

But in short, it's a great idea :-)

kolulu23:
@The-Lum 

> * ~rotate~ [it seems that is already implemented 👍]

I added rotation into the command parser but I haven't figure out the best way to implement this, you see in the original implementation, rotation is not contraint by packet dimension, having a roation like 60 or 1270 could easily give you diagram like this(which I am not very fond of).

image

Another thing I haven't done yet is to draw the boundaries of the overflowing packet with dashed lines. Currently I use rectangle shape for each packet, but URectangle doesn't provide a way to override the stroke for individual border. 

image

Also it seems that packets with custom height attribute in the original `packetdiag` do not play well with each other. This is different from my implementation where I _equalize_ jagged packets while keeping height proportion.

image

image

---


@arnaudroques I have changed target branch for this PR.

Thank you all again for your time.

The-Lum:
@kolulu23 

FYI, this page was created for the documentation:
- http://alphadoc.plantuml.com/doc/markdown/en/packetdiag

_If you want to contribute and complete the documentation with examples and all syntax._

Regards,
Th.

The-Lum:
@plantuml, @arnaudroques 

Here is an open question to PlantUML team:
_It time to uniformise or standardise all diagram definition_...

## `nwdiag`
For nwdiag, here is the current definition:

```puml
@startuml
nwdiag {
  network nw {
      address = "address"
  }
}
@enduml
```

```puml
@startuml
start
:
{{
nwdiag {
  network nw {
      address = "address"
  }
}
}}
;
end
@enduml
```

Perhaps we can standardize by (Here is a proposal):
```puml
@startnwdiag
  network nw {
      address = "address"
  }
@endnwdiag
```

```puml
@startuml
start
:
{{nwdiag
  network nw {
      address = "address"
  }
}}
;
end
@enduml
```

## `packetdiag`

I see the `packetdiag` keyword is optional:
https://github.com/plantuml/plantuml/blob/2d95eb874323e24842bd2a8c9867ba0af50f1837/src/main/java/net/sourceforge/plantuml/packetdiag/CommandPacketDiagStart.java#L21

We could even remove the curly braces...

```puml
@startpacketdiag
colwidth = 32
node_height = 72

0-15: Source Port
16-31: Destination Port
32-63: Sequence Number
@endpacketdiag
```

Open to debate...
Regards,
Th.

The-Lum:
Hi @kolulu23, @arnaudroques 

To complete:
>What are the remains missing functionalities from the original project (in order to be backward compatible)?

_After research (on the [nwdiag repo.](https://github.com/[blockdiag/nwdiag](https://github.com/blockdiag/nwdiag/tree/master/src)/tree/master/srchttps://github.com/[blockdiag/nwdiag](https://github.com/blockdiag/nwdiag/tree/master/src)/tree/master/src))_, there are:
- management of `desctable` with 'number', 'label', 'description'
Ref.:
https://github.com/blockdiag/nwdiag/blob/04bbb4581e90ea891fe826be6c82f4aaee522f26/src/packetdiag/elements.py#L25
- management of 'sparse packet'
```puml
{
1: A
5: B
}
```
- handle error or error management _(e.g.: test of the range)_
Here are some not working example:
 ```puml
{
4-2: A
5: B
}
```
↪ Fatal crash error: java.lang.ArithmeticException: / by zero ❌ 
  or 
```puml
{
5-2: A
6: B
}
```
↪ java.lang.IllegalArgumentException ❌ 

---

I can add this on:
- #2530


Regards,
Th.

Issue #2536, opened 2026-02-17
Add role label to associations in class diagrams


jpilgrim:
**Is your feature request related to a problem? Please describe.**
In UML, associations between classifiers are quite important. Two important concepts are end name (aka role which is used in the following) and multiplicity (aka cardinality). While an association name (aka label) cannot be expressed in most programming languages, the role usually becomes the name of the field of the implementing structure (e.g., class) and the multiplicity is often expressed by means of a simple reference (0..1) or a collection type (e.g. List). That is, role and multiplicity are crucial information for the implementation.

In PlantUML, there is only one slot per link end, which is used for the multiplicity (aka cardinality) in most examples. 

**Describe the solution you'd like**

In the UML specification and most UML tools, these two attributes are usually drawn on opposite sides of the association as follows

```
    User
owner | 1
      | 
 0..n | items 
     Item
```

I want a syntax and mechanism to specify multiplicity and role separately. I would suggest the following syntax:

`User "owner"/"1" -- "0..n"/"items" Item`

That is, the role can optionally be added after a slash after the end label. It is optional and the slots are not restricted to any content.

**Describe alternatives you've considered**

Several people have asked how to model the role, and a work-around at the moment simply is to write role and multiplicity in a single string, e.g. 

`User "owner 1" -- "0..n items" Item`

which looks like that:

```
          User
           | owner 1 
           |
items 0..n |
          Item
```

However this differs from the notation guidelines in the UML specifications and how most UML diagramming tools draw these attributes, making it harder to read.

**Additional context**

I will submit a pull request with an implementation of this feature.

arnaudroques:
@jpilgrim Thanks again for your contribution!

The following example:

```
@startuml
class User
class Item
User "owner which is very long"/1 -- "0..n"/items Item
@enduml
```

is even working there https://plantuml.github.io/plantuml/js-plantuml/  :-)

I must say that the syntax may sound odd: any ideas is welcome !

jpilgrim:
@arnaudroques Thanks for merging it that quickly. About the syntax: I thought that "/" would be a good divider and it hopefully does not break any existing diagrams.

Can I help adding tests and documentation somehow? I haven't found the user documentation ("Drawin UML with PlantUML") in the plantuml repo.

And what about tests? Should I add some tests in src/test/java/nonreg/simple ?

arnaudroques:
The syntax I'm proposing is really too weird: `User "owner which is very long"/1 -- "0..n"/items Item`

[...]

PR #2537, opened 2026-02-17
2536 Add role label to associations in class diagrams

jpilgrim:
This is the solution for issue #2536 

That is, the syntax for links in class diagrams is slightly extended to allow for specifying a role name, e.g.

`User "owner"/"1" -- "0..n"/"items" Item`

That is, the role name can be optionally added after a slash. There is no restriction on the content of the strings, so users can swap the attributes or use them for different purposes. It is optional, that is, the following (old) versions still work:

`User "owner" -- "items" Item`
or
`User "1" -- "items" Item`

I tried to keep changes at a minimum. The new fields (or local variables) are always called "role", although the content is not really restricted.

`gradle test`

works, but I haven't added any new tests (I assume tests are in a different repository?) If you would accept this PR and if you would give me a hint where to add tests, I would be happy to contribute some.

<details>
<summary>Disclaimer: The solution was generated by Claude Opus 4.6.</summary>
The solution was generated by Claude Opus 4.6. It was thoroughly reviewed by myself (the final version and intermediate solutions), and changes were kept at a minimum. This was the initial prompt:

```
Commandlinkclass several labels. at the moment, it is possible to add a label to the whole link (or relation) and labels to each line end. E.g.,

User 1 -- "0..*" Item: association

adds a label "1" on the line end of attachted to class "User", "0..*" on the line end attached to the class "Item" and a label "assocation" in the center of the line. Note that all labels can be placed in quotes if they contain spaces. Additionally quantifiers can be added (using square brackets).

Add the possibility to add a second label to a line end. This second label is to placed at the same end, but on the other side. E.g.,

User 1/owner -- "0..*"/items Item: association
```

Several solutions have been discarded, and eventually the following prompt had been used:

```
this made it worse. labels of vertical lines are now placed both on the left hand side of the line. Can you revert the rendering code and use xlabels (in svek, many similar thing for the other renderes) for the role labels?
```

Probably the most important information is the summary of the last fix, explaining the idea behind the positioning of the role labels:

```
Summary of the fix: Replaced the perpendicular reflection algorithm in drawRoleLabel() across all three renderers (SvekEdge, SmetanaEdge, MyElkEdge) with a simpler edge-aligned approach:

Vertical lines: Determine which side of the line the quantifier is on (left/right by comparing X coordinates), then place the role on the opposite side with its edge at the line + a small gap
Horizontal lines: Same logic but mirroring across the Y axis (above/below)
The quantifier's Y (or X for horizontal) position is preserved for the role, keeping them at the same height along the line
```

</details>


arnaudroques:
@jpilgrim Great job!

I only have one remark before merging this into master.

Current code:

```java
                new RegexOptional(new RegexLeaf(1, "FIRST_LABEL", "[%g]([^%g]+)[%g]")), //
                new RegexOptional(new RegexLeaf(1, "FIRST_ROLE", "/([%pLN_]+|[%g][^%g]+[%g])")), //
...
                new RegexOptional(new RegexLeaf(1, "SECOND_LABEL", "[%g]([^%g]+)[%g]")), //
                new RegexOptional(new RegexLeaf(1, "SECOND_ROLE", "/([%pLN_]+|[%g][^%g]+[%g])")), //
```

What about switching to:

```java
                new RegexOptional(new RegexLeaf(1, "FIRST_LABEL", "[%g]([^%g]+)[%g]")), //
                new RegexOptional(new RegexLeaf(1, "FIRST_ROLE", "/([^%s]+|[%g][^%g]+[%g])")), //
...
                new RegexOptional(new RegexLeaf(1, "SECOND_LABEL", "[%g]([^%g]+)[%g]")), //
                new RegexOptional(new RegexLeaf(1, "SECOND_ROLE", "/([^%s]+|[%g][^%g]+[%g])")), //
```

Rationale: `[%pLN_]+` restricts the “unquoted role” to letters/digits/underscore only (`%pLN_`), which is a bit too strict in practice. Using `[^%s]+` (i.e., “anything but whitespace”, since `%s` means space) makes the unquoted form more permissive while still keeping the parsing unambiguous: the role token remains a single contiguous chunk up to the next space, and the quoted alternative (`[%g]...[%g]`) continues to cover roles that may contain spaces.

In short: this keeps the same behavior for quoted roles, but makes the unquoted role accept common characters like `-`, `.`, `:`, etc., without forcing users to add quotes.

Any thought?

jpilgrim:
@arnaudroques Yes, `%pLN_` is a bit conservative... I tested `^%s` and all tests are still working. I was a little bit worried about consuming `-`, but it seems to work (only `A "label"/1-- B` seems to produce a weird result, as it is interpreted similar to `A "label"/"1-" - B`, but this is probably quite ok).

I created a PR #2544 with your improvement, but you maybe can simply delete that PR and add the change directly.

PR #2544, opened 2026-02-18
2536a Minor improvement to PR 2537

jpilgrim:
Added simplification of regex for second label aka role as suggested by @arnaudroques in #2537 as tests are working when applying that change.


jpilgrim:
@arnaudroques I'm a bit confused about the changes of this PR (11 files). I only added your suggested changes. It is probably easier if you just ignore this PR and delete it, and add your improvement directly to master-2536.

Issue #2506, opened 2026-01-31
Zero-Dependency SAX-based SVG Sprite Parser


npiper:
**Is your feature request related to a problem? Please describe.**

The SVG Dom parser introduced for SVG Sprite parsing does potentially add maintenance, dependencies and Jar size impacts to the plantuml codebase to be considered.

(NOTE: The following context below has been generated with AI CoPilot assistance and some assertions made may need further testing / verification)

**Describe the solution you'd like**

At request of reviewers, maintainers - consider alternative approaches that might not need Batik - SAX based parsing may be an option as it is still in the JDK

#
Create `SvgSaxParser` as an alternative SVG parsing implementation using only Java SDK components (no external dependencies). This would complement the existing `SvgDomParser` which uses Apache Batik.

### Motivation

While the current `SvgDomParser` works well with Batik dependencies managed via Maven Central auto-download, a pure Java SDK implementation offers several advantages:

- **Zero external dependencies** - No Batik JARs required (eliminates 4.2MB download)
- **Memory efficiency** - Event-driven SAX streaming vs. in-memory DOM tree
- **Build simplicity** - Works on any Java 8+ system without network access
- **Faster parsing** - Lower overhead for large SVG files
- **Philosophy alignment** - Matches PlantUML's "lightweight, minimal dependencies" design principle

#### Core Technology

Use Java SDK's built-in SAX parser (available since Java 1.4):
```java
import org.xml.sax.XMLReader;
import org.xml.sax.helpers.DefaultHandler;
import javax.xml.parsers.SAXParser;
import javax.xml.parsers.SAXParserFactory;
```

#### Implementation Strategy

**Two-pass parsing architecture:**

1. **Pass 1 - Collect Definitions:** Parse `<defs>`, `<symbol>`, `<linearGradient>`, `<radialGradient>` and store by ID
2. **Pass 2 - Render Elements:** Process shape elements, resolve `<use>` references, apply transforms and styles

**State management:**
- Manual stack for nested `<g>` group states (transforms, styles)
- Accumulator for `<text>` element content
- Map for forward reference resolution (`<use href="#later">`)

#### Code Reuse Opportunities

**~60% of existing SvgDomParser logic can be extracted and reused:**

✅ **Transform parsing** (~200 lines):
- `getTranslate()`, `getScale()`, `applyRotate()`, `applyMatrix()`
[...]

PR #2548, opened 2026-02-21
Feature/sax svg sprite parser 2506

npiper:
PR to support issue #2506 

# PR Description

## Summary
- Introduces a zero-dependency SAX SVG parser for sprite/emoji rendering.
- Adds parser selection via `plantuml.svg.parser` and wires sprite entry points to the factory.
- Expands sprite fixtures/tests for transforms, gradients, text, and defs/use.

## What Changed
- **Parser**: two-pass SAX parse (defs collection + render pass).
- **Factory**: `ISvgSpriteParser` + `SvgSpriteParserFactory` for SAX/Nano selection.
- **Sprites/tests**: additional fixtures for gradients, text anchors, transforms, defs/use.
- **Docs**: emoji package docs updated with constraints and defaults.

## Key Behaviors (SAX vs Nano)
- **Parsing model**: SAX is structured/two-pass; Nano is regex-based.
- **Text**: `text-anchor` honored in SAX; positioning can differ vs Nano.
- **Gradients**: SAX resolves linear gradients; radial falls back to first stop (renderer limitation). Nano has no defs/gradient resolution.
- **Defs/use**: SAX collects defs/symbols and resolves `<use>` (simplified). Nano has no defs/use.
- **Styling**: inline attributes only (no `<style>` blocks) in both.

## Constraints
- SVG 1.1 core subset only; no SVG 2.0 extensions.
- No `<style>` blocks, class selectors, clipPath, mask, filter, pattern.
- No advanced text (`tspan`, overline).
- Embedded raster images: PNG/JPEG data URIs only; no external URLs or embedded SVG.
- Inline `style` parsing limited to common text/paint attributes.
- SVG output wraps data-URI rasters in `<image>` to avoid clip suppression.
- TeaVM build lacks SAX APIs: Nano is forced under TeaVM.

## Testing
- `./gradlew clean test`
- Sprite image generation: `SvgSpriteImageGenerationTest`, `SvgNanoParserTest`
- Site build (no TeaVM): `./gradlew siteAssemble` or `./gradlew site -x generateJavaScript`
- TeaVM JS: `./gradlew generateJavaScript -Dorg.gradle.console=plain` (outputs under `build/teavm/js`)

## Risks / Rollback
- **Compatibility**: default is SAX; switch back with `-Dplantuml.svg.parser=nano`.
- **Performance/memory**: SAX is two-pass and buffers defs; may cost more on large sprites.
- **Security**: relies on JDK SAX defaults; can harden by disabling external entities/DOCTYPE if needed.

### DIFF
--- a/src/test/java/net/sourceforge/plantuml/emoji/SvgSpriteParserFactoryTest.java
+++ b/src/test/java/net/sourceforge/plantuml/emoji/SvgSpriteParserFactoryTest.java
@@ -1,0 +1,54 @@
+package net.sourceforge.plantuml.emoji;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+
+public class SvgSpriteParserFactoryTest {
+
+	private static final String SAMPLE_SVG = "<svg xmlns=\"http://www.w3.org/2000/svg\"></svg>";
+
+	@Test
+	public void testDefaultParserIsSax() {
+		String original = System.getProperty("plantuml.svg.parser");
+		try {
+			System.clearProperty("plantuml.svg.parser");
+			ISvgSpriteParser parser = SvgSpriteParserFactory.create(SAMPLE_SVG);
+			assertTrue(parser instanceof SvgSaxParser, "Default parser should be SAX");
+		} finally {
+			restoreProperty(original);
+		}
+	}
+
+	@Test
+	public void testExplicitSaxParser() {
+		String original = System.getProperty("plantuml.svg.parser");
+		try {
+			System.setProperty("plantuml.svg.parser", "sax");
+			ISvgSpriteParser parser = SvgSpriteParserFactory.create(SAMPLE_SVG);
+			assertTrue(parser instanceof SvgSaxParser, "Property 'sax' should select SAX parser");
+		} finally {
+			restoreProperty(original);
+		}
+	}
+
+	@Test
+	public void testExplicitNanoParser() {
+		String original = System.getProperty("plantuml.svg.parser");
+		try {
+			System.setProperty("plantuml.svg.parser", "nano");
+			ISvgSpriteParser parser = SvgSpriteParserFactory.create(SAMPLE_SVG);
+			assertTrue(parser instanceof SvgNanoParser, "Property 'nano' should select Nano parser");
+		} finally {
+			restoreProperty(original);
+		}
+	}
+
+	private void restoreProperty(String original) {
+		if (original == null) {
+			System.clearProperty("plantuml.svg.parser");
+		} else {
+			System.setProperty("plantuml.svg.parser", original);
+		}
+	}
+}


npiper:
Generated samples:

svg2Groups.svg
svg2Groups

svgDiagonalGradientBLTR

svgDiagonalGradientBLTRMonochrome

svgUseDefs

svgDiagonalGradientTLBR

svgVerticalGradient
ssets/c6ca0ad3-2c19-4287-b291-f1d4134722d7)


svgTransformTranslate
ents/assets/57b26a84-e3b0-4b0b-8bac-dc7e6487300e)

svgTransformGroup

svgTextStyleAttr

svgTextBold

svgTextItalic
5c68d6dc49)


svgTextDecoration

svgRedText

svgTextAnchor
87f3d3fa0)

svgRedRectLinearGradient

svgRedPolyLine
com/user-attachments/assets/b53a1609-c8dc-4acc-9138-9cb8cd074811)

npiper:
Fixes  #2506

arnaudroques:
Thank you very much for your work!

We have created a dedicated branch to continue working on this:
[https://github.com/plantuml/plantuml/tree/master-sax-svg-parser](https://github.com/plantuml/plantuml/tree/master-sax-svg-parser)
You can base your work on this branch. I have only moved a few classes into a new package named `svg.parser`.

I would like to share a general remark first. In PlantUML, we are aiming for *universality* of diagrams: the same diagram should render identically whether it is produced as PNG, SVG, EPS, and regardless of the execution environment (Java, and now JavaScript as well).

From that perspective, having multiple SVG parsers for sprites is not ideal. It means that the visual result of a diagram could vary depending on which parser is selected, which is something we would prefer to avoid.

That said, I am absolutely *not* saying that we do not want the new SAX parser — quite the opposite. Our goal is for it to eventually replace the existing Nano parser. However, this constraint means we will need to be a bit creative, especially because of the JavaScript/TeaVM target 🙂

### Short-term plan (before merging into `master`)

Before merging your PR into `master`, we would like to slightly adjust the approach.

For now, the Nano parser will remain the default. This allows us to preserve backward compatibility and avoid disrupting existing users.

The SAX parser can be enabled explicitly via a pragma:

```
!pragma svgparser sax
```

or

```
!pragma svgparser nano
```

You might wonder why this is preferable to a JVM option such as `-Dplantuml.svg.parser=nano`. If you usually generate diagrams locally, the difference may not be obvious. However, in a web deployment context, this allows the *end user* (who is not necessarily the person running the server) to choose the parser on a per-diagram basis.

I understand that this is slightly inconvenient for you, since you probably do not want to modify all your existing diagrams. To address this, we propose that the JVM option:

```
-Dplantuml.svg.parser=sax
```

changes the *default* parser (i.e. when no pragma is present in the diagram). I believe this should be an acceptable compromise.

Would you be willing to implement this change? Given the work you have already done on the SAX side, it should be fairly straightforward.

With this in place, we would be able to merge your branch, with the SAX parser initially considered as *beta*.

### Longer-term perspective

Looking ahead, the main challenge will be to make the SAX parser work with TeaVM 🙂

At first glance, this may sound impossible — but for XML parsing on the JavaScript side, we do have viable options:

* Use the native JavaScript XML APIs (`DOMParser`, `XMLSerializer`)
* Expose the result to Java via `@JSBody` or `JSObject` interfaces

Conceptual example:

```java
@JSBody(params = "xml", script = """
  var doc = new DOMParser().parseFromString(xml, "application/xml");
  return doc;
""")
public static native JSObject parseXml(String xml);
```

This suggests a possible architectural evolution: instead of the SAX parser directly producing SVG output, it would populate an intermediate model — for example, a `VectorielIntermediate` class that represents the SVG structure.

On the JavaScript side, the SVG would be parsed using `DOMParser`, and this would populate the *same* `VectorielIntermediate` model.

Finally, the code that generates the actual SVG from `VectorielIntermediate` would be shared between both approaches.

I hope this makes the direction clear. Let me know if you see where I am heading with this, and whether this architecture makes sense to you.

npiper:
No problem the `!pragma svgparser` approach for diagram editors makes sense vs. doing it from the server commands so people can choose an updated approach.

Q1) Just a question on the maintanance approach after this merge to make sure I'm following the process.

- assuming I can work directly and create a new PR from [plantuml:master-sax-svg-parser](https://github.com/plantuml/plantuml/tree/master-sax-svg-parser) or should I do it from my forked repo and pull across again?


**Tea VM**

Q2) Assuming this is a 'later feature' consideration that wouldn't have to address right now in the next PR to just address the `pragma` argument?

The TeaVM changes I bumped into at the end when doing the 'site' generation so might need to do some more reading to understand the direction / purpose of TeaVM to consider it.

At the moment this logic for SVG based Sprites (nano or SAX) is doing effectively an intermediate translation to the Plantuml Graphic (UGraphic*) object before converting to the target format ( PNG, SVG, PDF,..etc;)

```
Sprite + (SVG Def) 
   --> ISvgSpriteParser.drawU ( UGraphic*)  
     --->  UGraphic --> SVG, PNG Output
 ```
     
If I am understanding correctly it might be a similar approach for TeaVM with an intermediate modelling / representation.

An alternative approach may be to find a way to treat the Sprite SVG represenation as an 'embedded' part of a final SVG however this may be tricky in terms of positioning, viewboxes etc; when that logic for sprite handling may already be 'solved'.

**Comment by arnaudroques** [comment]
> Q1) Just a question on the maintanance approach after this merge to make sure I'm following the process.
> 
> * assuming I can work directly and create a new PR from [plantuml:master-sax-svg-parser](https://github.com/plantuml/plantuml/tree/master-sax-svg-parser) or should I do it from my forked repo and pull across again?
> 

Yes, you can work directly and create a new PR from [plantuml:master-sax-svg-parser](https://github.com/plantuml/plantuml/tree/master-sax-svg-parser)
This would be easier for everyone 

> **Tea VM**
> 
> Q2) Assuming this is a 'later feature' consideration that wouldn't have to address right now in the next PR to just address the `pragma` argument?
> 

Yes, exactly — you understood correctly.
Everything related to the TeaVM part is explicitly postponed for now.
The next PR will focus solely on the `pragma` argument. I'll explain the rationale for postponing TeaVM separately.

The introduction of TeaVM into PlantUML is very recent. It works, but it is not yet optimal.

**1)** To begin with, we relied on the SJPP preprocessor for the integration, but it now appears that this may not be necessary.
We already have an initial branch where we are working on removing this preprocessor for TeaVM and adopting a more “native” TeaVM approach.

See:
[https://github.com/plantuml/plantuml/tree/master-teavm](https://github.com/plantuml/plantuml/tree/master-teavm)

**2)** More importantly, a more coherent use of TeaVM leads to a fundamental architectural change.

As a result, we are actively working on another branch:
[https://github.com/plantuml/plantuml/tree/master-handwritten](https://github.com/plantuml/plantuml/tree/master-handwritten)

The core difference lies in the `Diagram.java` interface.

Until now, we had:

```java
public interface Diagram {

    public ImageData exportDiagram(OutputStream os, int num, FileFormatOption fileFormat) throws IOException;

    public void exportDiagramGraphic(UGraphic ug, FileFormatOption fileFormat);

    ...
}
```

The first method, `exportDiagram`, has historical roots: at the time, PlantUML only generated “files” (which is no longer true with TeaVM).
The second method was later added to allow diagrams to contain other diagrams. Conceptually, however, this is already somewhat inconsistent, as both methods largely serve the same purpose.

We now want to move toward something closer to:

```java
public interface Diagram {

    public TextBlock getTextBlock(int num, FileFormatOption fileFormatOption);

    ...
}
```

This single method returns a `TextBlock` representing the diagram itself, without taking headers, footers, legends, captions, etc. into account.
Those elements are shared by (almost) all diagrams and were a source of code duplication.

With this approach, the responsibility of the `Diagram` interface is refocused on its core purpose.
All decorations and file-oriented export logic will be delegated to another class.

This will simplify the codebase and make the TeaVM integration more rational, since TeaVM has no notion of files.
It will also enable additional improvements, but there is not enough space to go into them here 🙂

In short, we are currently in the middle of a major refactoring, and for now I would prefer not to introduce an additional work stream around SVG sprites and TeaVM.

I therefore suggest that we first proceed with the initial PR for `!pragma svgparser sax`.
I will be happy to explain the next steps in more detail afterwards.

npiper:
@arnaudroques I have a local committed branch with the Pragma change ready to push but have a permission block and might need write permission to the plantuml:plantuml repo to push directly to the SAX parser branch.

`remote: Permission to plantuml/plantuml.git denied to npiper.`

I could push to my fork instead perhaps and open a PR from there.

Here's my draft PR for the change I'll be pushing:
-----
# PR: SVG sprite parser selection via pragma

## Title
Enable SVG sprite parser selection via pragma (Nano default)

## Description
This PR implements pragma-based selection for SVG sprite pqarsing while keeping the Nano parser as the default to preserve backward compatibility. The SAX parser can be enabled explicitly with a pragma and remains opt-in. TeaVM builds always force Nano to avoid missing `javax.xml.parsers` during JavaScript generation.

## Motivation / Context
Maintainers requested that Nano remain the default parser and SAX be explicitly enabled. This change provides a simple, diagram-local switch that avoids breaking existing users while allowing testing and adoption of the SAX parser.

Related PR: https://github.com/plantuml/plantuml/pull/2548 (closed, #2548)

## Current behavior
- SVG sprite parsing uses Nano by default.
- There was no pragma-based selector for SAX vs Nano.

## New behavior
- Nano remains the default.
- The SAX parser is enabled per diagram using pragmas:

```
!pragma svgparser sax
```

or

```
!pragma svgparser nano
```

## TeaVM behavior
TeaVM does not include `javax.xml.parsers`. The parser factory forces Nano when `TeaVM.isTeaVM()` is true. TeaVM resolves this branch at compile time, so the JVM-only SAX path is removed from generated JavaScript output and class stubs.

## Tests
- Updated SVG sprite tests to use the SAX pragma where appropriate.
- Parser selection tests now validate pragma behavior and Nano default.

## Documentation
- Added SVG parser constraints to package Javadoc.
- Added a package README for SAX and gradient constraints.

## Notes / Limitations
- SAX parser supports inline attributes only (no CSS `<style>` blocks or class selectors).
- Gradient rendering is simplified (first/last stops only; direction approximated).
- Embedded raster images are supported only via data URIs (PNG/JPEG).

## Files changed (high level)
- Pragma key and selection logic for SVG parser choice.
- TeaVM fallback to Nano.
- Tests and SVG sprite fixtures updated to use `!pragma svgparser sax`.
- SVG parser documentation updates.

arnaudroques:
> @arnaudroques I have a local committed branch with the Pragma change ready to push but have a permission block and might need write permission to the plantuml:plantuml repo to push directly to the SAX parser branch.
> 
> `remote: Permission to plantuml/plantuml.git denied to npiper.`

You should have received an invitation now.
Thanks for your contribution!

PR #2612, opened 2026-03-23
Namespace Plantuml in the browser, and add a way to render to an svg …

acenturyandabit:
…string instead of a div

This makes plantuml.js nicer to use for projects which integrate plantuml as a component.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -82,6 +82,7 @@ repositories {
 teavm {
 	js {
 		mainClass.set("net.sourceforge.plantuml.teavm.browser.PlantUMLBrowser")
+		entryPointName.set("plantumlLoad")
 		obfuscated.set(true)
 		optimization.set(org.teavm.gradle.api.OptimizationLevel.BALANCED)
 		// obfuscated.set(false)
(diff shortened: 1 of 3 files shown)


arnaudroques:
Thanks for the patch! The namespace approach and `renderToString` are great additions. This makes the JS API much cleaner for projects integrating PlantUML as a component.

We have merged it with two minor tweaks:
- Moved the `removeAllChildren()` call after `buildSvg()` in `doRender()`, so the old SVG stays visible during rendering and avoids a visual flicker.
- Restored the `BrowserLog.reset()` call that was dropped during the refactoring, so the render duration in the status bar is measured correctly.

So many thanks for your contribution!

By the way, I'm curious about what project are you planning to use the JS version of PlantUML in?

acenturyandabit:
Thanks for tweaking and merging! I'm using plantuml as part of an app that extends plantuml with a nested bullet list and filtering, which allows dynamic expansion, rehighlighting and collapsing of subcomponents of the diagram.

image

I've currently been using plantuml on server-side, because as a TS/JS programmer, plantuml's Java code has been a dense closed box, but now with claude, I'm able to navigate the code and suggest some improvements. I'll look out a bit more closely for accidental removals in the future 😅 

Love this project, thank you for sharing!

PR #2619, opened 2026-03-24
Restore Group functionality in SvgGraphicsTeaVM

acenturyandabit:
I was using groups previously for highlighting with css selectors - this patch creates similar functionality (not quite drop in, but attaches 'data-entity' to the svg-groups, allowing mapping between svg elements and plantuml source strings.


acenturyandabit:
For some reason consecutive renders stop working after this. I'm still investigating.

acenturyandabit:
My error was in my own external code - this PR is fine.

Issue #878, opened 2022-01-24
It's possible to specified link position between packages and between component and packages?


hjianhao:
when I link packages

```
package p1 {
    component a1
    component b1
}

package p2 {
    component a2
    component b2
}

p1 --> p2
```
I got the result

![image](https://user-images.githubusercontent.com/6102007/150774292-58cdde7d-aada-4dfa-9425-d58178056081.png)

Can I specify the endpoint position of the link between packages, for example, link endpoint at the center of the package side?

![image](https://user-images.githubusercontent.com/6102007/150774778-5511e4ac-bde3-4a0f-b151-4593b45093f1.png)

So as the link between component and package

 
![image](https://user-images.githubusercontent.com/6102007/150774963-c5806047-c023-4b3b-90b5-bcadb1fec9ee.png)

The-Lum:
Hello @hjianhao,

_If the internal components are without any connection with the external components..._
A possible workaround is to use **sub-diagram**, as:
```puml
@startuml
package p1 [
{{
    component a1
    component b1
}}
]

package p2 [
{{
    component a2
    component b2
}}
]

p1 --> p2
@enduml
```
>[![](http://www.plantuml.com/plantuml/svg/SoWkIImgAStDuIhEpimhI2nAp5KeI4pEJanFTG9IwKNvfNafoVav-KKAvHcP9QbSK4c50aE5QAxgQYu584ZEpot8pqlDAr58D4GJI3BaggtbYkL2w3F2giy8NPyHJ9-XWgwk7L0RbqDgNWhGeG00)](http://www.plantuml.com/plantuml/uml/SoWkIImgAStDuIhEpimhI2nAp5KeI4pEJanFTG9IwKNvfNafoVav-KKAvHcP9QbSK4c50aE5QAxgQYu584ZEpot8pqlDAr58D4GJI3BaggtbYkL2w3F2giy8NPyHJ9-XWgwk7L0RbqDgNWhGeG00)

See also:
- https://forum.plantuml.net/13167/there-any-way-make-the-arrow-between-frames-being-align-center

Regards.

hjianhao:
This workaround has serious side effects,  the components in the sub-diagram can not link to other elements outside the package. That's almost unacceptable.  T_T

The-Lum:
I agree, and this is the mentioned constraint:
>If the internal components are without any connection with the external components...

The-Lum:
Hello @hjianhao,

[...]

PR #2654, opened 2026-04-09
Fix: center cluster edge arrows via za positioning and sibling-edge

dnguyenv:
[Svek] fix arrow endpoints centering on cluster borders
  - defer za anchor declaration between children for centered Graphviz positioning
  - enable `compound=true` with `lhead/ltail` for native cluster edge clipping
  - add `hasNoInternalEdges` guard to protect state machine layouts
  - add `hasNoEdgeToSibling` guard to prevent arc regressions on nested clusters
  - skip `compound=true` for LTR + ortho combinations
  - etc 

  Ref: #878, #2644 

@arnaudroques Here is my approach for the cluster edge centering fix. Appreciate your review and feedback!

### DIFF
--- a/src/main/java/net/sourceforge/plantuml/svek/Cluster.java
+++ b/src/main/java/net/sourceforge/plantuml/svek/Cluster.java
@@ -537,11 +537,23 @@ public class Cluster implements Moveable {
 
 	public SvekNode printCluster2(StringBuilder sb, Collection<SvekEdge> lines, StringBounder stringBounder,
 			DotMode dotMode, GraphvizVersion graphvizVersion, DiagramType type) {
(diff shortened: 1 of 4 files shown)


The-Lum:
Hi @arnaudroques, @dnguyenv 

- What is your conclusion?
- Are the modifications on `plantuml:master-2645` are acceptable?
- Do you plan to add on the main branch?

[EDIT]
See also:
- #2645

Regards,
Th.

dnguyenv:
Thanks @The-Lum for following up. The PR is as far as I can take it for now..would love to get feedback from @arnaudroques (and others) on whether it’s good enough to merge into main. happy to maintain the enhancements+  address any issues that come up.

PR #2665, opened 2026-04-22
🐛 fix label overlapping nodes in ortho layout

hundredGrand:
## Summary
Reduces cases where edge labels overlap nodes when using `skinparam linetype ortho`.

## Changes
- Extend existing collision handling to include edge labels (`labelXY`)
- Add label-label collision handling to reduce stacking in dense diagrams

## Tests
- Added Java tests for ortho rendering scenarios

## Notes
This change operates in the post-layout phase and does not modify Graphviz layout constraints.
In very dense diagrams, placement may still be suboptimal due to layout limitations.

## Reproduction
See `ortho_label_basic.puml` for a minimal example.

After: 
test

Before: 
Screenshot 2026-04-22 114940


The-Lum:
Hi @hundredGrand 

- Why not place the labels directly far above the arrows instead of far below?

Regards,
Th.

PR #2680, opened 2026-05-03
Complete implementation of Record and Dataclasses

Intybyte:
After some testing I realized I missed some implementation when making the PR #2232, this should add the missing stuff, and allow muting between the new added class types


src/main/java/net/sourceforge/plantuml/classdiagram/command/CommandHideShowByGender.java
```diff
@@ -143,6 +143,10 @@ else if (arg1.equalsIgnoreCase("metaclass"))
 			gender = EntityGenderUtils.byEntityType(LeafType.METACLASS);
 		else if (arg1.equalsIgnoreCase("stereotype"))
 			gender = EntityGenderUtils.byEntityType(LeafType.STEREOTYPE);
+		else if (arg1.equalsIgnoreCase("record"))
+			gender = EntityGenderUtils.byEntityType(LeafType.RECORD);
+		else if (arg1.equalsIgnoreCase("data"))
```
Intybyte, line 148: uhmmm whoops? should be dataclass, realized now

arnaudroques: Yes, probably :-)
I've just forced push a fix... :-)

PR #2714, opened 2026-05-21
🐛 accept decimal period, pulse and offset for timing clock

### LINKED ISSUE #2711: Decimals in clock not working in timing diagram
**Describe the bug**
Giving a decimal as period or offset of clock in timing diagram gives an error.

**To Reproduce**

```puml
@startuml
clock "MyClock" as Clk with period 12.5 offset 0.1
binary "State" as B

@0
B is low

@10.5
B is high

@enduml
```
gives the error in line 2

```puml
@startuml
clock "MyClock" as Clk with period 12 offset 1
binary "State" as B

@0
B is low

@10.5
B is high

@enduml
```
works as expected.

**Expected behavior**
As I depict the timing in ns, I also need decimals, not only integers

**Additional context**
I am using PlantUML version 1.2026.3 / 208af3a [2026-05-08 16:11:48 UTC] (GPL source distribution) and Java(TM) SE Runtime Environment (build 26.0.1+8-34)

IgniparousTempest:
https://github.com/plantuml/plantuml/issues/2711

The CommandClock regex only matched integer values for the clock's period, pulse and offset, so

    clock "MyClock" as Clk with period 12.5 offset 0.1

failed to parse. The downstream rendering code (PanelsClock, Panels.xOfTime, drawHorizontalBetweenTimes) already operates on doubles, so the values were being constrained needlessly at the parser level.

Allow decimals in the regex and propagate the values as double through PlayerClock, PanelsClock and TimingDiagram.createPlayerClock / getClockValue. BigDecimal.valueOf(double) is used in place of new BigDecimal(double) to avoid binary-precision noise.

Adds a non-regression test reproducing the original bug report.

Closes #2711


arnaudroques:
Thanks for your PR!

I have two points:

1) instead of changing code from `int` to `double`, could you rather move directly from `int` to  `BigDecimal`?
```java
	private final BigDecimal period;
	private final BigDecimal pulse;
	private final BigDecimal offset;
```

2) About test, it's great that you have provide some of them.

However, we are moving to a better & simplier way.
It's not documented yet, but you'll find an example here:

https://github.com/plantuml/plantuml/blob/master/src/test/resources/vega/mvp/hello-both.puml

You just have to create a new .puml file somewhere in the "vega" tree structure (you can even create folder).
In the YAML header, you write a few lines, with the output format and the expected-description.
Something like:

```puml
---
output: svg, debug
expected-description: (Timing Diagram)
---
@startuml
clock "MyClock" as Clk with period 12.5 offset 0.1
binary "State" as B

@0
B is low

@10.5
B is high

@enduml

```

When you run test for the first time, a .txt or/and .svg file will be created for you.
This will be the expected result for future tests.


Does it sound good to you?

IgniparousTempest:
Done

Issue #2740, opened 2026-06-06
Publish the TeaVM JS build from the MIT license flavor?


nanodeath:
**Is your feature request related to a problem? Please describe.**

I'm developing an MIT-licensed Logseq plugin that renders diagram code blocks, and I'd
like to vendor the JS engine the way I vendor mermaid. `@plantuml/core` being GPL
makes that awkward — but since MIT jars exist, I tried building the JS engine from
the `plantuml-mit` flavor.

**Describe the solution you'd like**

It works. The TeaVM browser code is excluded from all non-GPL flavors by blanket
sjpp pragmas; narrowing those to `JAVA8`-only (and giving sudoku the same
`::comment when` guard that `PSystemBuilder` already uses) lets
`CI=1 ./gradlew :plantuml-mit:generateJavaScript` produce a working `plantuml-mit.js`.
Validated against the GPL build in headless Chromium across seven fixtures (incl. C4
and AWS stdlib includes): byte-identical output, with sudoku verifiably absent.

My guess is the blanket exclusion was packaging (the JS build originally only
targeting the GPL edition) rather than a licensing decision — the earlier browser
builds (`plantuml.js`, `plantuml-core`) were MIT, and TeaVM itself is Apache-2.0.
But #1336 describes the pragmas as marking code that *cannot* be MIT-licensed, so I'd
rather ask than assume:

1. Is the TeaVM browser code intended to be available under the permissive
   flavors?
2. If yes, would you accept a PR (6 files, +24/−11, no change to the GPL build)? I
   have it ready, and I'm happy to also extend the npm packaging (e.g.
   `@plantuml/core-mit`) if welcome.

If the exclusion is deliberate, that's entirely your call — a one-line statement to
that effect somewhere would still be valuable, since the MIT jars make the JS
situation ambiguous from the outside.

**Describe alternatives you've considered**

- Loading the GPL `@plantuml/core` from a CDN at runtime instead of vendoring —
  workable, but murkier than I'd like for a plugin I distribute.
- Server-side rendering (PlantUML server / Kroki) — planned as a separate path
  anyway, but it loses the zero-setup, fully client-side experience the TeaVM
  build enables.

**Additional context**

Thanks for the TeaVM browser build and the npm publishing (#2715) — the engine
running fully client-side is fantastic, and this request only exists because I want
to build on it.

arnaudroques:
[...]

PR #2741, opened 2026-06-07
✨ enable TeaVM JS build for the MIT license flavor

nanodeath:
Follows up on #2740 — thanks for the quick green light!

This enables `CI=1 ./gradlew :plantuml-mit:generateJavaScript`, producing a
`plantuml-mit.js` from the MIT licence flavor (7.1 MB vs the GPL build's 7.2 MB).

**What changed:** the sjpp pragmas that removed the TeaVM browser code from every
non-GPL flavor are narrowed to `JAVA8`-only. Sudoku — the one GPL-bound component
reachable from the browser entrypoint — now uses the same `::comment when` guard in
`PSystemBuilder2` that `PSystemBuilder` already uses, so it remains GPL-exclusive.
No behavior change for the existing GPL JS build.

**Validation (headless Chromium, MIT build vs published GPL `@plantuml/core@1.2026.5`
behavior at the same base commit):**

| Fixture | Result |
| --- | --- |
| sequence, activity, 20-node component, class | byte-identical SVG output |
| `!include <C4/C4_Container>` | byte-identical (stdlib bridge intact) |
| `!include <awslib/...>` sprites | byte-identical |
| invalid input | identical error SVG |
| `grep -c sudoku plantuml-mit.js` | 0 (vs 1 in the GPL build) |

(The single byte difference across all fixtures was sub-pixel coordinate jitter in
one hand-drawn `cloud` outline — the pseudo-random wobble path.)

**Testing done** (on this branch, Java 25, master @ c09f4b2):

- `CI=1 ./gradlew :plantuml-mit:generateJavaScript` — BUILD SUCCESSFUL, 7.1 MB `plantuml-mit.js`
- `CI=1 ./gradlew :plantuml-{epl,bsd,asl,lgpl,gplv2}:compileJava` — all compile
  (the narrowed pragmas include the TeaVM sources in every flavor, so this was the
  regression risk worth checking)
- Rendered 7 fixtures in headless Chromium against the built artifact: sequence,
  activity, component, class, `!include <C4/...>`, `!include <awslib/...>`, and an
  invalid-input case (returns the usual error SVG)
- `grep -ci sudoku plantuml-mit.js` → 0
- Checked the sjpp-processed sources: the newly included teavm files get the MIT
  header stamped as expected
- `./gradlew test` (root project, GPL flavor) — BUILD SUCCESSFUL, 0 failures
  (expected, since the root-tree diff is comment-only, but verified rather than assumed)

npm packaging/publishing changes aren't included — this just makes the MIT
flavor produce the artifact, so transitioning `@plantuml/core` is unblocked
on your side.

### DIFF
--- a/plantuml-mit/build.gradle.kts
+++ b/plantuml-mit/build.gradle.kts
@@ -10,6 +10,7 @@ plugins {
 	java
 	`maven-publish`
 	signing
+	alias(libs.plugins.teavm)
 }
 
 group = "net.sourceforge.plantuml"
@@ -29,6 +30,16 @@ dependencies {
 	testImplementation(libs.junit.jupiter)
 	testImplementation(libs.jlatexmath)
 	testImplementation(libs.xmlunit.core)
+	teavm(teavm.libs.jsoApis)
+}
+
+teavm {
+	js {
+		mainClass.set("net.sourceforge.plantuml.teavm.browser.PlantUMLBrowser")
+		moduleType.set(org.teavm.gradle.api.JSModuleType.ES2015)
+		obfuscated.set(true)
+		optimization.set(org.teavm.gradle.api.OptimizationLevel.BALANCED)
+	}
 }
 
 repositories {
--- a/src/main/java/net/atmp/CucaDiagram.java
+++ b/src/main/java/net/atmp/CucaDiagram.java
@@ -466,7 +466,7 @@ public abstract class CucaDiagram extends TitledDiagram implements GroupHierarch
 		final CucaDiagramFileMaker maker;
 
 		if (TeaVM.isTeaVM())
-			// ::revert when __MIT__ __EPL__ __BSD__ __ASL__ __LGPL__ __GPLV2__ JAVA8
+			// ::revert when JAVA8
 			maker = new CucaDiagramFileMakerTeaVM(this);
 		// maker = new CucaDiagramFileMakerSmetana(this);
 		// ::done
--- a/src/main/java/net/sourceforge/plantuml/nio/PathSystem.java
+++ b/src/main/java/net/sourceforge/plantuml/nio/PathSystem.java
@@ -64,7 +64,7 @@ import net.sourceforge.plantuml.teavm.browser.TeaVmScriptLoader;
 public class PathSystem {
 
 	public static PathSystem fetch() {
-		// ::comment when __MIT__ __EPL__ __BSD__ __ASL__ __LGPL__ __GPLV2__ JAVA8
+		// ::comment when JAVA8
 		if (TeaVM.isTeaVM())
 			return new PathSystem(null);
 		// ::done
@@ -72,7 +72,7 @@ public class PathSystem {
 	}
 
 	public JsonValue getTeaVMStdlibJson(String path) {
-		// ::revert when __MIT__ __EPL__ __BSD__ __ASL__ __LGPL__ __GPLV2__ JAVA8
+		// ::revert when JAVA8
 		// return null;
 		path = path.replaceAll("\\.json$", "");
 		final String full = path.toLowerCase();
@@ -89,7 +89,7 @@ public class PathSystem {
 	}
 
 	public InputStream getTeaVMStdlibInputStream(String path) {
-		// ::revert when __MIT__ __EPL__ __BSD__ __ASL__ __LGPL__ __GPLV2__ JAVA8
+		// ::revert when JAVA8
 		// return null;
 		final String full = path.substring(1, path.length() - 1).toLowerCase();
 		String libname = full.substring(0, full.indexOf('/'));
@@ -113,7 +113,7 @@ public class PathSystem {
 		// ::done
 	}
 
-	// ::comment when __MIT__ __EPL__ __BSD__ __ASL__ __LGPL__ __GPLV2__ JAVA8
+	// ::comment when JAVA8
 	private Map<String, String> getInfo(final String libname) {
 		final JSObject info = TeaVmScriptLoader.getRaw_PLANTUML_STDLIB_INFO(libname);
 		final Map<String, String> map = new HashMap<>();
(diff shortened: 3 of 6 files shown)


arnaudroques:
@nanodeath Following up on your MIT work, here's where it landed on the npm side:

https://github.com/plantuml/plantuml/commit/0043b6b7da2b74208296ce732be85b2bb2f8c0ca

The short version: `@plantuml/core` will now be built from the MIT flavor instead of GPL. The `plantuml-mit` subproject gets its own `npmPackage` task (the root GPL build stays as-is for the site/demo). Since TeaVM names the output `plantuml-mit.js`, we copy it into the package renamed to `plantuml.js` so existing CDN URLs and the demo imports keep working.

You're more fluent in npm than we are, so we'd love your eyes on a few things specifically:

- the `package.json` (exports map, `files` list, `"license": "MIT"`, `sideEffects: false`)
- whether shipping `plantuml.js` from the MIT build under the same package name is the right call, or whether the licence change warrants anything special on the npm side (a major bump, a note, etc.)
- anything you'd do differently in the publish flow (now documented in `PUBLISHING_NPM.md`)

We're planning a release in the next few days, so any feedback before then would be much appreciated. And of course, don't hesitate to push back. Thanks again!

nanodeath:
Hey @arnaudroques, thanks for reviewing so quickly, appreciate it. Given the CLAUDE.md file in the repo, I wonder if we're just using Claude to chat back and forth on these or...if it's just me :sweat_smile: so I'm not sure how much insight I really have to add here; I haven't actually published anything on npm. I do review what I'm commenting, course.

I'm pasting what my Claude Code suggested below after the divider, feel free to take it or leave it, but re: license, if you're going from GPL to MIT (e.g. more permissive), I agree with it that you don't need to worry too much about that; it's more like if you were going from MIT to GPL, you'd want a clean break there so it's easier for the community to stay on the last MIT version as a kind of unofficial "LTS" release, or in case they want to fork from that point. Regardless, a note is definitely called for in the release notes.

---

Ran the task locally and put the assembled package through the same validation
harness I used for the PR (7 fixtures incl. stdlib includes, rendered against the
renamed `plantuml.js`) — everything works. The rename-at-copy approach holds up.

A few findings, in decreasing order of importance:

**1. The tarball has no LICENSE file.** `npm pack --dry-run` shows 17 files and
no license text — npm only auto-includes a `LICENSE` if it exists in the package
dir, and the task doesn't copy one. `plantuml-mit/mit-license.txt` is already in
the repo, so it's one more copy block:

```kotlin
copy {
    from(layout.projectDirectory) {
        include("mit-license.txt")
        rename { "LICENSE" }
    }
    into(pkgDir)
}
```

(npm picks up `LICENSE` automatically — no `files` change needed.) Given the whole
point of this release is the license, I'd consider this the one must-fix.

**2. `"sideEffects": false` will break `viz-global.js` for bundler users.**
`viz-global.js` is a classic script whose entire job is installing a global — it
is nothing *but* side effects. With `sideEffects: false`, webpack production
builds that do `import "@plantuml/core/viz-global.js"` will tree-shake the import
away entirely and Graphviz layout silently disappears. Fix:

```json
"sideEffects": ["./viz-global.js"]
```

(path-prefixed form pins it to the exact file; a bare `"viz-global.js"` is
treated as `**/viz-global.js`, which also works but is looser than intended.)

**3. The `exports` map locks out `emoji.js` and `openiconic.js`.** Once
`exports` exists, every subpath not listed throws
`ERR_PACKAGE_PATH_NOT_EXPORTED` for Node/bundler consumers — the files are in
the tarball but unreachable by bare specifier (direct CDN URLs are unaffected).
Suggest:

```json
"exports": {
  ".": "./plantuml.js",
  "./plantuml.js": "./plantuml.js",
  "./viz-global.js": "./viz-global.js",
  "./emoji.js": "./emoji.js",
  "./openiconic.js": "./openiconic.js",
  "./package.json": "./package.json"
}
```

**On same-name vs. major bump:** keep the name and the versioning scheme — no
major bump. GPL→MIT only *loosens* obligations: no consumer's build breaks, the
API is unchanged, and npm records the license per version, so already-published
versions stay accurately labeled GPL. A 2.x would desync your calendar-locked
versioning for no real protection. Two things I would do instead:

- a line in the README and release notes: "**versions ≥ 1.2026.6 are MIT;
  earlier versions are GPL-3.0-or-later**" — the npm UI only displays the
  *latest* version's license, so users pinned to older versions should be told
  what they have;
- one release-note line that sudoku diagrams no longer render (the single
  intentional MIT delta), so the one person using that isn't surprised.

**On the publish flow:** the procedure is correct (the version-derivation regex
checks out — `1.2026.6beta1` → `1.2026.6-beta.1` — and the `--tag beta` guidance
is right). Two suggestions:

- The "bypass 2FA" granular token is the weakest link — a standing secret that
  can publish. npm's **Trusted Publishing** (OIDC, GA since July 2025; GitHub
  Actions, GitLab CI, and now CircleCI) removes it entirely: no token exists,
  CI authenticates via short-lived workflow-scoped credentials, and **provenance
  attestations are generated automatically** (no `--provenance` flag needed in
  that flow) — a verifiable link from the published tarball back to this repo's
  CI build. For a package whose pitch is "you can trust what's in this," that's
  a strong fit. Requires npm CLI ≥ 11.5.1 / Node ≥ 22.14.0 and a public repo for
  the provenance half. This is also somewhat forced going forward: npm
  permanently deprecated classic tokens in December 2025, so the long-lived-token
  path is on its way out regardless. If you'd rather stay manual for now, an
  interactive `npm publish --otp=...` is still safer than a bypass-2FA token.
- Consider moving the sudoku grep *into* the gradle task — have `npmPackage`
  fail if the count isn't 0, rather than relying on a manual checklist step:

```kotlin
val engine = file("$pkgDir/plantuml.js").readText()
check(!engine.contains("sudoku", ignoreCase = true)) {
    "GPL build leaked into the npm package (sudoku found in plantuml.js)"
}
```

Happy to sanity-check the published tarball once it's on npm.

PR #2771, opened 2026-07-02
Fix #2099: allow a class to reuse a name claimed by an implicit namespace

### LINKED ISSUE #2099: No longer have class and namespace with same name
**Describe the bug**

I use the VSCode extension for plantuml which ships with `1.2024.3-gplv2` as far as I can tell. In this version I was capable of creating a namespace with a certain name, and then a class with the same name. I use this to organise sub-classes in my diagrams better.

Since upgrading Plantuml with homebrew to `1.2025.1` this is not working and throws an error. I am unsure which version introduced this issue as it had been a while since I upgraded Homebrew's Plantuml, but I think the issue was also present back in `1.2024.8`.

**To Reproduce**

Use diagram with `1.2025.1` (or `1.2024.8`):

```
namespace Organisation {
  class SubClass
}
class Organisation
```

to output:

Image

**Expected behavior**

Use same uml code should output:
Image

**Version info:**

- OS: macos 15.3.1
- Browser: n/a
- Plantuml Version: 1.2025.1, 1.2024.8

herley-shaori:
## Background

`class a.b` implicitly creates a namespace/group `a` containing `b`. Since `1.2023.2` (commit `58cc4e096b4`, Feb 2023), that implicit group is promoted eagerly during parsing instead of lazily at final render time, so a later `class a` finds quark `a` already occupied by a `GROUP` entity and fails `Entity.muteToType`'s whitelist check, producing `(Error)`.

In the issue thread, `@The-Lum` posted the minimal repro:
```plantuml
@startuml
set separator .
class a.b
class a
@enduml
```
`@arnaudroques` confirmed this is a real problem ("Having the same ID for a class and a package is indeed an issue") and agreed to rework the design to support it. `@designermonkey` reported real-world impact - this is a regression (functionality that previously worked), not a new feature request, and it forced editing "many, many UML diagrams" to work around.

One correction worth noting: I root-caused this with `git bisect` against actual historical builds rather than just reading the code, and the break is older than the thread assumed - it's been present in every release since `v1.2023.2` (Feb 2023), not something that changed between `1.2024.3` and `1.2025.1`.

## What this PR does

**Commit 1** - the core fix: adds a `phantomGroup` flag to `Entity`, set when a group is auto-created purely from a dotted child name (`eventuallyBuildPhantomGroups`), and cleared when a group is explicitly declared via `namespace`/`package`. `Entity.muteToType` now allows converting a phantom group into a leaf/class, while explicit namespaces are untouched and still correctly reject a same-named class (relation/arrow resolution depends on that staying unambiguous for real namespaces - verified with a manual check, see below).

**Commit 2** - a real bug the first commit's own tests didn't catch: checking only `Diagram.getDescription()` isn't enough. A promoted phantom-group `Entity` kept the `Bodier` built by `CucaDiagram.createGroup` (a plain `BodierSimple`, never bound to a leaf via `Bodier.setLeaf`, and unable to render fields/methods - it throws `UnsupportedOperationException` on `getMethodsToDisplay`/`getFieldsToDisplay`). The description succeeded while actually rendering the image crashed with a `NullPointerException` in `BodyEnhanced1` ("Cannot invoke `Entity.getColors()` because `entity` is null"), reproducibly, even for the basic repro above with no relations or body content. Fixed by building a fresh `Bodier` via `BodyFactory.createLeaf` (the same call used for a brand-new leaf) and binding it with `setLeaf`, right at the point of promotion.

## Testing

All of this was verified by actually rendering SVG images, not just checking diagram descriptions/entity state - the second commit exists specifically because the first round of testing looked clean at the description level but crashed on real render:

- Basic repro (`class a.b` / `class a`): previously `(Error)`, now succeeds; previously crashed on image render, now renders cleanly.
- A relation pointing at the phantom group before it gets promoted (`class a.b` / `X --> a` / `class a`): previously crashed on render, now renders cleanly.
- Multiple children under the phantom group before promotion (`class a.b`, `class a.c`, `class a`): previously crashed on render, now renders cleanly.
- The promoted class with real body content (fields/methods) - the scenario that specifically needed a proper leaf `Bodier`: previously crashed on render, now renders cleanly.
- Explicit `namespace a { }` + `class a` still correctly rejected - confirmed this fix does not weaken that protection.
- Full suite: 2985 passed, 0 failed, 11 skipped (all pre-existing, unrelated to this change).

The four rendering scenarios above are now permanent regression tests in `NamespaceSameNameAsClassTest`, asserting on the actual rendered SVG (not just the description), so this specific class of bug - fix looks correct at the description level but crashes on real render - can't silently reappear.

Refs #2099

### DIFF
--- a/src/main/java/net/atmp/CucaDiagram.java
+++ b/src/main/java/net/atmp/CucaDiagram.java
@@ -329,6 +329,7 @@ public abstract class CucaDiagram extends TitledDiagram implements GroupHierarch
 				// final Display display = Display.getWithNewlines(quark.getQualifiedName());
 				final Display display = Display.getWithNewlines(getPragma(), quark.getName());
 				final Entity result = this.createGroup(location, quark, GroupType.PACKAGE);
+				result.setPhantomGroup(true);
 				result.setDisplay(display);
 			}
 		}
@@ -353,6 +354,9 @@ public abstract class CucaDiagram extends TitledDiagram implements GroupHierarch
 		}
 		final Entity ent = quark.getData();
 		ent.muteToGroupType(type);
+		// This group is now explicitly declared by the user (namespace/package
+		// command), so it is no longer a mere implicit/phantom group.
+		ent.setPhantomGroup(false);
 		if (usymbol != null)
 			ent.setUSymbol(usymbol);
 
--- a/src/main/java/net/sourceforge/plantuml/abel/Entity.java
+++ b/src/main/java/net/sourceforge/plantuml/abel/Entity.java
@@ -106,6 +106,7 @@ final public class Entity implements SpecificBackcolorable, Hideable, Removeable
 	private String generic;
 
 	private GroupType groupType;
+	private boolean phantomGroup;
 
 	// Other
 	private Margins margins = Margins.NONE;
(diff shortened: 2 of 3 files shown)


arnaudroques:
Thanks for your contribution!

I may be wrong, but this does not completely solve the issue.
(see https://github.com/plantuml/plantuml/commit/06848d28d6ad03bdcfb26e3363b009a92ce50457 )

The following example:

```plantuml
@startuml
set separator .
class a.b
class a
@enduml
```

render this: 
image

Without any "b" class.

Is this the expected result?

herley-shaori:
You're right, that's not the expected result - confirmed it as a real bug and opened a follow-up fix: #2772.

**Quick summary:**
- `b` wasn't actually lost from the data model (it still exists as a genuine leaf `Entity`), it was a rendering gap: `GraphvizImageBuilder` only finds entities via (1) recursing into direct children that are still `GROUP` type, or (2) leafs whose parent container is exactly the diagram root. Since `a` is promoted from a phantom namespace to a plain class (no longer a `GROUP`), neither path reached `b` anymore - it fell between the two.
- Fix: `getUnpackagedEntities()` now walks up the container chain past any plain-leaf container until it hits the root (treated as top-level, same as before) or a real `GROUP` (left to the normal nesting path, untouched). Explicit `namespace`/`package` nesting is unaffected.
- Verified with failing-first tests reproducing your exact example (asserting the rendered SVG contains both `a` and `b`), visual confirmation that `a`/`b`/`c` all render correctly now, confirmed real namespaces still nest normally, and a full suite run (2988 passed, 0 failed).

Details and evidence are in the #2772 description.

PR #2825, opened 2026-08-21
feat(js-plantuml): Share diagrams via URL fragments

toransahu:
Why:
- Make diagrams shareable as self-contained URLs without server-side state.
- Let recipients open a link and immediately recover the same diagram source.

What:
- Synchronize editor content to a PlantUML-compatible URL fragment.
- Restore content on initial load and hash navigation, and clear the hash for empty content.
- Add a share button with clipboard fallback and a brief top-centered “Copied” notification.

How:
- Encode UTF-8 source with raw DEFLATE and PlantUML’s URL-safe alphabet.
- Vendor fflate 0.8.2 locally so encoding and decoding remain fully browser-local.
- Debounce hash updates, use history.replaceState, and announce copy results through an accessible status region.

Testing:
- Verified Unicode round trips and compatibility with plantuml-encoder 1.4.0 fragments.
- Exercised render, hash update/restore/navigation, clipboard fallback, empty/invalid fragments, and toast behavior in headless Chrome.
- Checked JavaScript syntax, whitespace, and the vendored bundle integrity.


toransahu:
Sample change is available at: https://www.toran.xyz/plantuml/js-plantuml/index.html#3Son3S9030NG_gRu0y02512e41LIXAWTR0JYt_da-FOd2xnfJOaSbUNCxylsnuilEE1gfEFfGTrX_lYWuavf-QZ4RSPYwE7LqnIoGkqjWmdf_LJ2SaI3_yfaJJVw3m00

arnaudroques:
Great, many thanks for your contribution!

toransahu:
@arnaudroques Thanks for accepting the PR. Here goes the next: https://github.com/plantuml/plantuml/pull/2826

Issue #2847, opened 2026-08-31
`!theme` is silently ignored in the browser (TeaVM) build


lemonlion:
**Describe the bug**
In the browser (TeaVM) engine, `!theme <name>` is accepted and then silently ignored. No styling is applied and no error is reported. An unknown theme name (`!theme zzz-does-not-exist`) behaves exactly like a real one, so a typo is invisible too.

The engine also advertises the themes it cannot apply: `%get_all_theme()` returns all 44 bundled names in the browser (#2725 made it TeaVM aware), while `%get_current_theme()` returns `{}` after a `!theme` line naming one of them.

**To Reproduce**

1. Render this diagram with the browser engine (`@plantuml/core` on npm, or a `plantuml.js` built from master), for example in the playground `index.html` shipped with the package:

```
@startuml
!theme amiga
Alice -> Bob: hello
Bob --> Alice: hi
@enduml
```

2. Compare with the same diagram without the `!theme` line.
3. The two rendered SVGs are identical apart from the embedded `<?plantuml-src ?>` metadata: default `#E2E2F0` participants instead of the amiga theme's `#0B58A8` on `#FFFFFF`.

**Expected behavior**
Same behaviour as the desktop build: the theme is applied, and an unknown theme name reports an error.

**Screenshots**
Four different `!theme` directives, all rendering identically unthemed:

![four themed diagrams all rendering with the default styling](https://raw.githubusercontent.com/lemonlion/plantuml/783f7ebdc9244b5e835a5138f1dceb5e52cdee45/themes-before.png)

**Desktop (please complete the following information):**

- OS: any (verified on Windows 11; the cause is in the compiled engine, not the host)
- Browser: any (verified on Chromium headless)
- Version: `@plantuml/core` up to and including 1.2026.7, and current master (8c574c5)

**Additional context**
Two causes, both browser only:

1. `TContext.executeTheme` has its whole body inside `if (!TeaVM.isTeaVM())`, so the directive is a no-op there. Since `isTeaVM()` is a `@PlatformMarker` the body is also dead-code eliminated, which is why no theme content appears anywhere in the compiled `plantuml.js`.
2. `ThemeUtils.loadBundledOrLocalTheme` reads `/themes/puml-theme-<name>.puml` through `getResourceAsStream`, and the browser build has no classpath, so the lookup could not succeed even if the code path were live.

[...]

PR #2849, opened 2026-08-31
🍵 Paint the document background in the browser (TeaVM) build (stacked on #2848)

lemonlion:
Follow-up to #2848, now rebased onto master after its merge, so the diff is just the four commits of this fix (RED, GREEN, REFACTOR, plus edge-case pins from a later review pass).

- **#2848** (merged): `!theme` was silently ignored by the browser engine (#2847). It made the themes apply.
- **This PR:** even with themes applied, the browser engine never paints the document background. That bug also exists without themes (`skinparam backgroundColor` is ignored too), but themes made it visible: white-on-blue amiga renders white text on the host page. This is the fix promised in the Scope section of #2848.

## What happens today

The browser (TeaVM) engine never paints the diagram background. The Java build honours the merged `root.document` BackGroundColor, which is where both `skinparam backgroundColor` and a theme's background land, by writing `style="...background:#RRGGBB;"` on the `<svg>` plus a rectangle covering the whole drawing. The browser build emits neither.

With #2848 this became visible: 21 of the 44 bundled themes set a document background in the Java build, 12 of them non-white, and those themes apply their element colours but lose their backdrop. `!theme amiga` is white on blue, so on a white page its arrows and text vanish.

## Why

Everything downstream already existed. `SvgGraphicsTeaVM` takes a background color in its constructor and paints the style plus a full viewBox rectangle, with the same skip rules as `SvgGraphics.paintBackcolor`. It just never received the color: `PlantUMLBrowser.buildSvg` only asked `tb.getBackcolor()` on the TextBlock, where the document background never surfaces, so it always took the no-background branch.

## The change

`buildSvg` now falls back to the merged `root.document` style via `TitledDiagram.calculateBackColor()`, exactly where `ImageBuilder.styled()` gets it for the Java build, extracted as `getPaintableDocumentBackground`.

The diversion only happens for a background that should actually be painted, judged on the plain (light mapped) color so the decision is mode independent: transparent and the skin default of white keep painting nothing, in dark mode too, where the skin pairs white with a dark value but the page has always supplied its own backdrop. Falling through also keeps the historic WHITE and BLACK fallback as the contrast reference for automatic colors.

## Red, green, refactor

The commits after the #2848 base are the actual TDD sequence (plus a later commit pinning edge cases):

1. RED: `browser-test/check-background.js`, asserting the Java contract (background style plus full viewBox rectangle, verified against the viewBox dimensions) for `skinparam backgroundColor`, the `<style>` document form and two themes, and asserting the skips (default white, transparent, dark mode control) keep painting nothing. Against the engine as master builds it today (with #2848 merged), the four painting checks fail:

```
PASS  control renders
PASS  control paints no background (white is skipped)
FAIL  skinparam backgroundColor paints the background
        expected background #0B58A8, got style=null rect=null
FAIL  <style> document BackGroundColor paints the background
FAIL  !theme amiga paints its blue background
FAIL  !theme blueprint paints its background
PASS  skinparam backgroundColor transparent paints nothing
PASS  dark mode control renders
PASS  dark mode control paints no background
```

2. GREEN: the fix. The RED checks earned their keep twice during this step: an unconditional diversion and a diversion judged on the mode mapped color both made the dark mode control start painting a background it never painted before, and the dark check caught both.

3. REFACTOR: the resolution extracted into a named method with the reasoning as javadoc, and the new check wired into the browser-test workflow. All checks re-run unchanged.

## Verification

- `check-background.js`: 16 of 16 on the fixed engine (the 4 painting checks fail before it). A later review pass probed the edges and found no defect; the probes are committed as pins: dark mode keeps an explicit background while suppressing only the default, scale keeps the rectangle covering the viewBox, activity and class diagrams paint too, a skinparam after `!theme` overrides the theme background like the Java build, and a gradient degrades to its first color instead of erroring.
- `check-themes.js` from #2848 (now on master): still 13 of 13.
- The 22 corpus diagrams (default background) are byte-identical to the build without this fix, so nothing changed for diagrams that set no background.
- `gradlew test -Pci` green.

Before (engine from current master: themes apply but backgrounds are missing):

before: themed diagrams with their page backgrounds missing

After:

after: the same themes with their backgrounds painted

## Scope

Gradient backgrounds degrade to the gradient's first color (`HColorGradient.toColor` under TeaVM); the Java build renders a real gradient. Left as is here.

### DIFF
--- a/.github/workflows/browser-test.yml
+++ b/.github/workflows/browser-test.yml
@@ -66,3 +66,6 @@ jobs:
 
       - name: Check themes
         run: node browser-test/check-themes.js target=plantuml-mit/build/npm-plantuml
+
+      - name: Check background
+        run: node browser-test/check-background.js target=plantuml-mit/build/npm-plantuml
--- a/browser-test/README.md
+++ b/browser-test/README.md
@@ -31,6 +31,14 @@ There are no golden files. The expected rendering for a theme is derived from th
 text, read out of the `themes.js` being tested, so adding or editing a theme needs no fixture
 update.
 
+## check-background.js
+
+Checks that the document background is painted the way the Java build paints it: a
+background style on the svg element plus a rectangle covering the whole viewBox, for
+`skinparam backgroundColor`, the `<style>` document form, and themes. The deliberate
+skips are pinned too: transparent and the default white paint nothing, in dark mode as
+well. Run the same way with `node check-background.js target=...`.
+
 ## Running it
 
 ```
--- a/browser-test/check-background.js
+++ b/browser-test/check-background.js
@@ -1,0 +1,157 @@
+'use strict';
+// Functional check that the diagram background is painted in the PlantUML browser (TeaVM)
+// engine, the way the Java build paints it.
+//
+// usage: node check-background.js target=<dir-or-js>
+//   target   a directory containing plantuml.js (viz-global.js and themes.js are served from
+//            there too when present), or a path to the engine .js file itself
+//
+// The Java build honours the merged root.document BackGroundColor (which is where both
+// skinparam backgroundColor and a theme's background land) by writing a background style on
+// the svg element plus a rectangle covering the whole drawing. These checks assert the same
+// contract on the browser engine, including the deliberate skips: null, transparent, pure
+// black and pure white paint nothing (see SvgGraphics.paintBackcolor and the
+// SvgGraphicsTeaVM constructor). Every check runs; the exit code is non-zero if any failed.
+const path = require('path'), http = require('http'), fs = require('fs');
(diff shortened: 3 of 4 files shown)


arnaudroques:
@lemonlion Once again, many thanks for your contributions!
There are some minor conflicts, is it possible for you to fix them?

lemonlion:
Done. The conflicts were the five #2848 commits this branch carried while stacked; after your squash merge I rebased onto master, so the diff is now just the four commits of this fix. Re-ran both check suites on the rebased head: check-background 16 of 16, check-themes 13 of 13. PR description updated to match. Thanks for merging #2848!

PR #2850, opened 2026-08-31
Document missing Javadoc @param tags #2629

AzazelSensei:
#2629 listed a few methods whose Javadoc skipped a parameter.

I filled those in. Huffman is the vendored brotli copy so I left it. The NGM types now live under `gantt`.


arnaudroques:
@copilot resolve the merge conflicts in this pull request

AzazelSensei:
Rebased onto current master. The Notion commit sitting under this branch had already landed, which is what made it conflict. The javadoc @param tags are the only change now.

PR #2862, opened 2026-09-02
👷 perf-bench: recenter expected bands on the measured ratios

lemonlion:
Small follow-up on perf-bench.

The bands in `expected-bands.json` are all still 1.0, but the pinned reference `4ce99a47` is from before P2 to P5 landed (#2836, #2837, #2838, #2839), so master renders the corpus in roughly half the pinned engine's time. Every row of every run comes back as "faster than band by 25-50pp".

Nice as that is to look at, a flag that fires on all 13 rows in all 11 runs so far isn't telling us much. The bigger problem is the other direction: if something gave back most of the P2-P5 win and pushed gen-500 from 0.42 up to 0.55, the summary would still print "faster than band by 35pp" and nobody would spot it.

So this just moves each center to the median of the 11 master runs to date (25077cc9 through b2392e62), all of which ran against the current pin. Tolerance stays at 0.1 and the A/A calibration behind it is untouched.

| Diagram | Was | Now |
|---|---|---|
| gen-50 | 1.00 | 0.47 |
| gen-200 | 1.00 | 0.46 |
| gen-500 | 1.00 | 0.42 |
| puml-0 | 1.00 | 0.45 |
| puml-1 | 1.00 | 0.45 |
| puml-19 | 1.00 | 0.44 |
| shape-groups | 1.00 | 0.62 |
| shape-parallel | 1.00 | 0.44 |
| shape-wide | 1.00 | 0.53 |
| shape-activation | 1.00 | 0.50 |
| shape-self | 1.00 | 0.52 |
| shape-createdestroy | 1.00 | 0.62 |
| small-30 | 1.00 | 0.56 |

Worst deviation from these medians across the 11 runs is 0.07 on gen-50, and everything else is 0.04 or under, so the whole existing history still sits comfortably inside the band. No run that has already happened would start complaining.

A couple of things worth flagging:

* The numbers are tied to the current pin. Once `reference.json` moves to `npm:1.2026.8` these all need recomputing, so I put the median rule in a `comment-3` next to the existing notes rather than leaving it in a commit message.
* Nothing about the non-blocking behaviour changes. Out of band rows are still only marked in the summary and the job still exits 0.

One thing I noticed while pulling the numbers together, not addressed here: the target/reference ratio still correlates with runner speed at about -0.66 across these 11 runs, so the ratio cancels most of the host difference but not all of it. Over the 1.4x host range we have seen that is worth roughly 0.03 of ratio. Well inside the 0.1 tolerance, so it does not affect this change, but it might be worth a look if the tolerance ever gets tightened.

Numbers came from the step logs rather than the artifacts, since those are on 1 day retention and everything before run 11 had already expired. Happy to share the extraction script if it is useful.

Draft for now in case the medians look off to you, or you would rather bump the pin than recenter.

---

One follow-up commit, prompted by review: `--only-shell` on the Playwright install in both workflows, answering the question about trimming the browser downloads. Everything here launches with `headless: true`, which Playwright 1.62.1 serves with the Chrome Headless Shell, so the full Chrome for Testing download (about 170MB per run) was never executed. FFmpeg has no skip flag in the installer, but it is only a few MB. The executed browser is unchanged, so the recentered bands above are unaffected (verified by installing only the shell into an isolated browsers path and running the bench against it).

The move of `perf-bench/` and `browser-test/` under `tools/` is split out as #2868 so this PR stays reviewable.


The-Lum:
Hello @lemonlion, @arnaudroques 

_[Just an open question]_
To avoid to had a lot of folders on the project root repository:
- Could you create a new root folder `js`, `npm` or `plantuml-js` in order to put the two `perf-bench` and `browser-test` folders ?

Is it a good idea?
What is the better name?
_Open to debate..._

Regards,
Th.

lemonlion:
Hey @The-Lum  - good idea.  Personally I'd be a little more inclined to put them into the `/tools` folder - not sure the other options quite hit the mark?

arnaudroques:
Nice suggestion, ok for:

```
/tools/perf-bench
/tools/browser-test
```

@lemonlion I let you do the change on your side. No pressure, we have time :-)

The-Lum:
Hi @lemonlion,

_Another dummy question..._

Based on the log:
- https://github.com/plantuml/plantuml/actions/runs/33653439287/job/100325950747#step:9:207

_(For testing and perf-bench)_, do we need both:
- Chrome for Testing 151.0.7922.34 (`playwright chromium v1234`)
- Chrome Headless Shell 151.0.7922.34 (`playwright chromium-headless-shell v1234`)

As well as:
- FFmpeg (`playwright ffmpeg v1011`)

Then here is my question:
- Could we limit it to the strict minimum?

Regards,
Th.

lemonlion:
@The-Lum yup - we can trim them, good spot!

The-Lum:
Hello @lemonlion 

As mention:
>The move of `perf-bench/` and `browser-test/` under `tools/` is split out as #2868 so this PR stays reviewable.

- Could we move this PR from `draft` to `ready to review` ?

Regards,
Th.

arnaudroques:
@lemonlion Sorry, I have merged in the wrong order :-)
Could you fix the conflicts so we can merge this PR?
Sorry about that and many thanks for your contributions

Issue #2872, opened 2026-09-05
Make stdlib includes work for every host of the JS engine (base URL, JSON bundles, loader hook)


lemonlion:
## The problem

Write a C4 architecture diagram, or anything else that uses the standard library, and render it with the JS engine anywhere other than the project site, and it dies on its first line:

```
@startuml
!include <C4/C4_Context>
System(a, "A", "...")
@enduml
```

gives `Fatal parsing error` on the `!include`. For the people this hits, C4 is the whole point: teams keep architecture docs as Markdown with the diagrams inline, and the standard library is how those diagrams are written. Today the stdlib works on plantuml.github.io and effectively nowhere else. Three classes of hosts are locked out:

1. **Every page consuming the engine from npm.** The engine resolves an include by lazy-loading `<lib>.min.js` through a script tag whose URL is relative, and a relative script src resolves against the consumer's document, not against the engine's location. So a page at `example.com/docs/arch.html` importing `@plantuml/core` from unpkg requests `example.com/docs/c4.min.js` and gets a 404. The npm package deliberately excludes the bundles (they total ~95 MB), and the only published copies live in `js-plantuml/` on this repo's gh-pages branch, where no npm consumer's loader will ever look.

2. **Browser extensions.** The small bundles can simply be shipped next to the extension's `renderer.html` (a PR doing that for 16 libraries is up at the extension repo), but the sprite collections cannot: ibm.min.js is 23.6 MB, tupadr3.min.js 20.2 MB, material7.4.47.min.js 17.2 MB, the full set roughly 90 MB. The natural answer is fetching them on demand, and `plantuml.github.io` already sends `Access-Control-Allow-Origin: *`, so the network side is solved. What is not solved: an extension may fetch remote **data** but may not execute remote **code**. Chrome's MV3 remotely hosted code policy and the AMO "add-ons must be self-contained" rule both forbid running a fetched `.min.js`, and the manifest CSP blocks it anyway. Today the bundles only exist as executable JS, so a store-compliant extension has nothing it is allowed to fetch.

3. **Web Workers and non-browser runtimes.** There is no `document` to append a script tag to, so stdlib includes cannot work at all in a worker, even though the engine itself runs fine there. `getTheme` already dodged the same trap by reading `PLANTUML_THEMES` from `globalThis` instead of loading anything.

[...]

Issue #2870, opened 2026-09-03
🍵 [TeaVM] Folder to put all `stdlib`


The-Lum:
## Context
To follow:
- #2868 
- 2ddb64d on #2869

## Wanted feature
- Could you put all `<stdlib>.min.js` from [`resources/teavm`](https://github.com/plantuml/plantuml/tree/master/src/main/resources/teavm)  to a folder `resources/teavm/stdlib`?

Regards,
Th.

The-Lum:
Hi @arnaudroques 

Here is the first commit:
- 3b66131

I will leave the others _(when TeaVM/JS read the stdlib file,...)_ to someone else...

Regards,
Th.

PR #2873, opened 2026-09-05
Stdlib includes for every host of the browser engine: PLANTUML_STDLIB_BASE and PLANTUML_STDLIB_LOADER

lemonlion:
Part of #2872; this is the engine side. The JSON data bundles are the companion plantuml-stdlib PR, linked below.

## Problem

A diagram that uses the standard library, `!include <C4/C4_Context>` above all, renders fine on plantuml.github.io and dies with `Fatal parsing error` on the include line everywhere else the browser engine runs:

1. **Every page importing the engine from npm or a CDN.** The engine lazy-loads `<lib>.min.js` through a script tag whose URL is relative, and a relative script src resolves against the consumer's document, not the engine's location. The include requests the bundle from the consumer's own origin and gets a 404. In practice stdlib works on the project site and nowhere else.
2. **Browser extensions.** plantuml/plantuml-for-github#10 is the field report: teams keeping C4 architecture docs as Markdown on GitHub cannot use the short stdlib syntax at all. The small bundles can be shipped in the package (plantuml/plantuml-for-github#13 does that for 16 libraries), but the sprite collections total roughly 90 MB and need on-demand loading. Store rules allow fetching remote data and forbid executing remote code, and today the bundles only exist as executable JS.
3. **Web Workers and non-browser runtimes.** No document, so a script tag has no path at all, even though the engine itself renders fine in a worker.

## Change

All in `TeaVmScriptLoader`, JSBody only, no Java-side behaviour change. One scoping fact shaped the design: every lazily loaded support script goes through this one loader, not just stdlib bundles but also `themes.js`, `emoji.js` and `openiconic.js`, and they all suffer the same relative-URL problem.

- `loadOnce` consults two host-provided globals before falling back to the script tag, in this order:
  - `PLANTUML_STDLIB_LOADER(url, onOk, onErr)`: a callback that delivers the script's content itself, however the host wants. For a stdlib bundle it populates `PLANTUML_STDLIB` / `PLANTUML_STDLIB_JSON` / `PLANTUML_STDLIB_INFO` and calls `onOk`, or calls `onErr` with a message. The coalescing map and its states behave exactly as for the script path. Because the hook receives every support script, it can decline a URL by strictly returning `false`, and loading falls back to the script tag, so a host that only handles stdlib bundles does not break themes or emoji.
  - `PLANTUML_STDLIB_BASE`: a string prefixed onto the relative URL, so one line points the loading of all of these files at the project site, a CDN or the page's own assets.
- `loadOnce`, `isLoaded` and the three `getRaw_PLANTUML_STDLIB*` readers address the globals through `globalThis` (with the same fallback chain `getTheme` already uses) instead of `window`, so a worker host can supply bundles through the hook. On a page `globalThis === window`, so nothing changes there.
- `GITHUB_INTEGRATION.md` gains a section documenting both globals with a working hook example, the npm README templates (both flavors) point their sprite-libraries paragraph at `PLANTUML_STDLIB_BASE` instead of the previously impossible advice to load them from the project site, and the browser-test workflow runs the new check alongside the existing five. All five existing checks still pass against the engine built from this branch, which matters because themes.js loading goes through the same loader this PR touches.

With neither global set, behaviour is exactly what it always was, and the warm-path benchmark below shows the cost of the new branches is not measurable.

The companion plantuml-stdlib PR (plantuml/plantuml-stdlib#174) publishes each library as a `<lib>.json` data bundle, which is what a loader hook fetches when it may not execute code.

## Tests

`tools/browser-test/check-stdlib-loader.js` (README section included), same shape as the other checks: synthetic one-participant libraries, each served only at the location its scenario is supposed to use, so a wrong loading path cannot render by accident. 18 assertions:

- legacy behaviour pinned: relative bundle loads, missing bundle fails visibly without hanging, stdlib-free diagrams untouched, no page errors;
- base URL: bundle fetched from the prefix and never from the page origin;
- loader hook: renders from fetched JSON with no `.min.js` request anywhere, concurrent includes coalesce into one loader call per library, a library whose info carries `link` resolves through two loader calls, and a loader failure surfaces as a visible include error;
- decline contract: a hook that returns `false` for everything is consulted and loading still succeeds through the script tag.

Against the engine built from this branch (`:plantuml-mit:npmPackage`): all 18 pass. Against the current engine: the legacy assertions pass and the new-behaviour assertions fail, so the check discriminates rather than passing vacuously.

## Evidence

Before, current engine on a localhost page, the exact C4 sample from the extension issue:

before: 404 and fatal parsing error

After, engine built from this branch, one line of configuration (`PLANTUML_STDLIB_BASE` pointed at the project site):

after: C4 renders via base URL

Same diagram through the loader hook, bundle delivered as fetched JSON, no script execution:

after: C4 renders via JSON hook

The worker case, which no configuration of the current engine can reach, renders in about 900 ms with the hook (mock-DOM worker host, OffscreenCanvas text metrics; the current engine in the identical host never completes the render). A stricter variant pins the globalThis change specifically: with `window` replaced by a measurement-only stub (`{ document }`, no PlantUML globals reachable through it) and the hook writing stdlib data to `globalThis` only, the engine from this branch still renders, while a control engine carrying the hook but window-based readers fails the include even though the data is present. So the stdlib path demonstrably reads through `globalThis`, not through a host's window alias.

## Performance

render-time cost chart

- Warm render, no stdlib include, median of 29: 11 ms on the current engine, 11 ms with this change (p95 13 both). The new branches sit behind the already-loaded fast path.
- C4 first render, cold, engine from this branch, median of 5: 826 ms via the JSON hook, 896 ms in a Web Worker, 996 ms via base URL to plantuml.github.io with the bundle fetch included.
- Wire cost of the data format: c4.json is 23.4 KB gzipped against 24.8 KB for c4.min.js, so JSON is slightly smaller than the JS form of the same content.

Prior art: this is the shape Mermaid settled on for the same two problems (lazy chunks resolved against the module, and `registerIconPacks` with a host loader returning Iconify JSON).

### DIFF
--- a/.github/workflows/browser-test.yml
+++ b/.github/workflows/browser-test.yml
@@ -79,5 +79,8 @@ jobs:
       - name: Check viz missing
         run: node tools/browser-test/check-viz-missing.js target=plantuml-mit/build/npm-plantuml
 
+      - name: Check stdlib loader
+        run: node tools/browser-test/check-stdlib-loader.js target=plantuml-mit/build/npm-plantuml
+
       - name: Add Gradle title to Summary
         run: echo "## Gradle" >> $GITHUB_STEP_SUMMARY
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1012,8 +1012,22 @@ tasks.register("npmPackage") {
 			  integration proofs of concept
 
 			Heavy optional sprite libraries (IBM, tupadr3, material, AWS...) are **not**
-			bundled here to keep the package small; load them from the project site if
-			you need them.
+			bundled here to keep the package small. To use them (and the rest of the
+			standard library: `!include <C4/C4_Context>`, azure, kubernetes, ...), point
+			the engine at wherever the bundles are hosted before rendering; they are
+			fetched lazily, one bundle per library, only when a diagram includes it:
+
+			```html
+			<script>
+			  window.PLANTUML_STDLIB_BASE = "https://plantuml.github.io/plantuml/js-plantuml/";
+			</script>
+			```
+
+			(Note the trailing slash: the value is a plain URL prefix.) For production,
+			self-host the bundles your diagrams use and point the base at your own
+			assets. Hosts that cannot load scripts at all (Web Workers, browser
+			extensions) set a `PLANTUML_STDLIB_LOADER` callback instead -- see
+			[GITHUB_INTEGRATION.md](./GITHUB_INTEGRATION.md).
 
 			## License
 
--- a/plantuml-mit/build.gradle.kts
+++ b/plantuml-mit/build.gradle.kts
@@ -432,8 +432,22 @@ tasks.register("npmPackage") {
 			  integration proofs of concept
 
 			Heavy optional sprite libraries (IBM, tupadr3, material, AWS...) are **not**
-			bundled here to keep the package small; load them from the project site if
-			you need them.
+			bundled here to keep the package small. To use them (and the rest of the
+			standard library: `!include <C4/C4_Context>`, azure, kubernetes, ...), point
+			the engine at wherever the bundles are hosted before rendering; they are
+			fetched lazily, one bundle per library, only when a diagram includes it:
+
+			```html
+			<script>
+			  window.PLANTUML_STDLIB_BASE = "https://plantuml.github.io/plantuml/js-plantuml/";
+			</script>
+			```
+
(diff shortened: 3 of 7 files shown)


The-Lum:
Thanks a lot for this enhancement.
Link to this another enchancment:
- #2870

PR #2875, opened 2026-09-06
2870 teavm folder to put all stdlib

### LINKED ISSUE #2870: 🍵 [TeaVM] Folder to put all `stdlib`

arnaudroques:
## Summary

Commit 1580b706dd68 moved every `<lib>.min.js` from `resources/teavm/` to
`resources/teavm/stdlib/` with no other change (per #2870), but two things
depended on the old flat layout and broke silently:

- `TeaVmScriptLoader.loadOnce` built the script-tag / `PLANTUML_STDLIB_BASE`
  fetch URL from the bare `<lib>.min.js`, so every `!include <lib/...>` in
  the JS engine now 404s. Fixed by adding `stdlib/` to the actual fetch path
  only — keyed off the `.min.js` suffix that already tells a stdlib bundle
  apart from `themes.js`/`emoji.js`/`openiconic.js`. The `url` seen by a
  `PLANTUML_STDLIB_LOADER` hook, and used as the internal cache key, is left
  untouched, so a host already using that hook API (shipped in 1.2026.8)
  keeps working unmodified.
- The `teavmZip` Gradle task's `include("*.js", ...)` pattern only matches
  files directly under the output dir — Gradle/Ant-style patterns are not
  implicitly recursive — so the published `js-plantuml` zip (GitHub Pages
  demo, GH Actions artifact) silently dropped every stdlib bundle. Added
  `stdlib/**`.

## Other changes

- The `teavm` task's "TeaVM JS files" size report used a non-recursive
  `File.listFiles`, so it silently stopped listing/summing the stdlib
  bundles. Switched to a recursive walk with relative-path display.
- Updated `GITHUB_INTEGRATION.md` to document the `stdlib/` subfolder and
  clarify that the `PLANTUML_STDLIB_LOADER` hook contract is unaffected.
- Updated `tools/browser-test/check-stdlib-loader.js` so the synthetic
  libraries and assertions exercise the real `stdlib/` (and
  `cdn/stdlib/`) paths for the relative/base/decline scenarios.

## Test plan

- [ ] `gradlew teavmZip` and confirm the zip contains `stdlib/*.min.js`
- [ ] `node tools/browser-test/check-stdlib-loader.js target=...` passes
- [ ] Manually render a diagram with `!include <c4/C4_Context>` in a
      built js-plantuml page and confirm it no longer 404s

cc @lemonlion — could you take a look? In particular I'd like a second
opinion on the `TeaVmScriptLoader` approach (prefixing the fetch path only,
not the hook-facing `url`) vs. simply prefixing `stdlib/` in `PathSystem.java`.


lemonlion:
Thanks for the ping. Good catch on the zip pattern too, that one would have been painful to track down after a release.

I'd keep it the way you have it, fetch path only. The hook contract from 1.2026.8 is the thing worth protecting: hosts key everything on the bare `<lib>.min.js`. The javadoc even documents pre-seeding `__pl_script_state['c4.min.js'] = { state: 'loaded' }`, so a PathSystem prefix would silently break any host doing that. The follow-up I have in mind for the heavy libraries also maps `url.replace(/\.min\.js$/, '.json')`, which keeps working untouched with your version. And since loadOnceSync just delegates to loadOnce, there's only one fetch site and the fix covers it. Where the file sits on disk feels like a loader detail, so the loader is the right place to know about it.

Two things worth keeping in mind for later rather than for this PR.

The extension vendors its bundles flat at the extension root, and fetch_stdlib.py downloads from the flat site URLs. Both are fine on the 1.2026.8beta1 engine it ships today, but the moment the vendored engine gets bumped past this PR they start to 404. So if the extension release after 1.2027.8 includes an engine bump, the bundle move to stdlib/ (plus the web_accessible_resources entries and the two build scripts) needs to land in the same commit. Happy to prepare that on my side when the time comes.

Anyone who followed the "self-host the files next to your page" guidance also gets silent 404s after upgrading, since BASE now composes with stdlib/. Your doc update covers it, but it's probably worth a line in the release notes. If you want to make the transition unbreakable, the script tag's onerror could retry the bare path once, at the cost of a wasted request when a bundle is genuinely missing. Not sure it's worth the noise, just flagging the option.

Two small ones. The .min.js suffix now does double duty (hook opt-out convention and the path rewrite), so maybe a one-line note in the resources folder saying nothing except stdlib bundles may use that suffix, otherwise a future themes.min.js would quietly land under stdlib/. And when the .json bundles from plantuml-stdlib#174 eventually get published, they should mirror the stdlib/ layout so the BASE story stays uniform.

lemonlion:
Also, thanks for adding me to the repo - I hope to continue adding value 🙂

arnaudroques:
Actually, on second thought, I feel this has too much of an impact.

I agree that having all the `.min.js` files at the root of `resources/teavm` isn't very clean in the source, but it seems to be the standard way to keep all these `.min.js` files in the same directory as `plantuml.js`, which is better in term of deployment (I think).

Therefore, perhaps temporarily, I propose we hold off on this change (sorry @The-Lum !).
That said, with all of @lemonlion suggestions, we might end up with a better solution.

Still regarding TeaVM, I'm also wondering if we shouldn't just remove the `viz.js` part altogether. Since we now have a working setup thanks to Smetana, it would keep things simpler.
(And from my perspective, it's a first step towards dropping `GraphViz` itself in the pure Java version.)

What do you all think?

The-Lum:
Hi @arnaudroques, @lemonlion,

If there are real technologies contraints: that not a problem for me.

_It was just to do a little tidying up._ 🧹

Regards,
Th.

lemonlion:
> Still regarding TeaVM, I'm also wondering if we shouldn't just remove the viz.js part altogether. Since we now have a working setup thanks to Smetana, it would keep things simpler.(And from my perspective, it's a first step towards dropping GraphViz itself in the pure Java version.)

Makes sense to me, I was going to suggest that we remove it from the front end to simplify when i put in the optimisations, but thought it better to have parity with the backend for the moment. That said, if we go down that route might be worth doing a large scale render and comparison between the two engines and find out the diffs so we can try and reduce layout changes to future consumers when switching from one engine to the other (eg from backend to frontend)

